PR #149, opened 2017-12-04
Pagination for Cromwell, add jm_utils library

dshiga:
Make use of server-side pagination in Cromwell.

Also, move code from servers/dsub/jobs/controllers/utils/page_tokens.py into a new common library for backends in servers/jm_utils, so that both dsub and cromwell backends can use it.

I also removed the `sort` step from the pip freeze instructions everywhere because it causes case sensitive sorting, which is confusing. `pip freeze` already sorts case-insensitive for us anyway. And when I tried including `sort` it reordered the existing requirements, which suggests it wasn't used previously.


servers/jm_utils/setup.py
```diff
@@ -0,0 +1,31 @@
...
+
+REQUIRES = []
+
+setup(
+    name=NAME,
+    version=VERSION,
+    description="Utilities for Job Monitor backends",
+    author_email="",
+    url="",
+    keywords=["Swagger", "Job Monitor Service"],
```
calbach: nit: monitor -> manager here and throughout this file

servers/cromwell/README.md
```diff
@@ -28,7 +28,13 @@ latter lists only direct dependencies. To regenerate run:
...
-pip freeze | sort > servers/cromwell/requirements.txt
+```
+Then, from the cromwell directory in this repo:
+```
+pip install -r requirements-to-freeze.txt
+pip freeze | sed 's/^jm-utils.*/\.\.\/jm_utils/g' > requirements.txt
 deactivate
 ```
+
+The sed command above replaces jm-utils=x.y.z with ../jm_utils, which is required
```
calbach: This is unfortunate, but I don't have a better suggestion. Perhaps @bfcrampton has an idea.

servers/dsub/README.md
```diff
@@ -51,7 +51,13 @@ latter lists only direct dependencies. To regenerate run:
 ```
 virtualenv --python=/usr/bin/python2 /tmp/dsub-server-requirements
 source /tmp/dsub-server-requirements/bin/activate
-pip install -r servers/dsub/requirements-to-freeze.txt
-pip freeze | sort > servers/dsub/requirements.txt
+```
+Then, from the dsub directory of this repo:
+```
+pip install -r requirements-to-freeze.txt
+pip freeze | sed 's/^jm-utils.*/\.\.\/jm_utils/g' > requirements.txt
```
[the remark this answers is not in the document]
dshiga: Ah, good to know. I've added it back but put `sort -f` for case-insensitive sorting. Without `-f` I get a different ordering than what I found in the requirement.txts.

servers/cromwell/README.md
dshiga: Yeah, I don't really like it either. Long term, I think a better way to share code would be to make `servers` one big installable Python package. That way we wouldn't need to put jm_utils into requirements.txt at all. I started exploring that but it requires big changes to the docker compose files, and needs more discussion. Assuming people want to go that way, I think it's more appropriate to make a separate ticket and PR to do it later.

samanehsan:
Is filtering intentionally disabled when QueryJobsResponse is paginated?

servers/cromwell/README.md
samanehsan: @dshiga for now I liked your suggestion to explicitly install `jm_utils` in the dockerfile instead of putting it in `requirements.txt`, because then we wouldn't need to do this sed command to fix the generated requirements

dshiga: It seemed attractive to just copy the jm_utils directory into the docker image, instead of pip installing it, but it doesn't mesh well with tox, which wants all dependencies installed via requirements files. We could put jm_utils in cromwell's and dsub's test-requirements.txt to get around that. I think it would work, but it feels risky to have dependencies installed one way in the Dockerfiles and another for unit tests. What do you think?

calbach:
@samanehsan no - pagination should have no side effects on the results of the query, beyond breaking it into pages. If you're referring to the dsub shim implementation, it filters in all cases.

dshiga:
In a separate discussion with @calbach, we determined that the filter is deliberately disabled when not all pages have been loaded into the ui, to avoid confusion about what's being filtered.
https://github.com/DataBiosphere/job-manager/blob/5d78e2ea7c86d5dc819ef11716f0ea36517f1dc0/ui/src/app/projects/projects.service.ts#L45
https://github.com/DataBiosphere/job-manager/blob/5d78e2ea7c86d5dc819ef11716f0ea36517f1dc0/ui/src/app/job-list/table/table.component.html#L5

If you click through until you completely run out of new pages, then filtering gets turned back on.

samanehsan:
Thanks for the clarification @dshiga @calbach!

servers/cromwell/README.md
samanehsan: That's a good point! As discussed in person, I think it's better to leave it as-is in this case for consistency.

bfcrampton: I looked into it a bit and couldn't come up with any better solutions. I agree the way it is now is probably best for consistency.

PR #156, opened 2017-12-05
Link subworkflow tasks to details page

samanehsan:
Link subworkflow tasks to their detail pages. E.g. for this workflow, "analysis" and "submit" would link to detail pages with more info about those subworkflows: 

screen shot 2017-12-05 at 3 50 41 pm


ui/src/app/job-details/panels/panels.component.html
```diff
@@ -19,7 +19,7 @@
       <p id="submitted" *ngIf="job.submission">Submitted {{ job.submission | jmShortDateTime }}</p>
       <p id="started" *ngIf="job.start">
         Started {{ job.start | jmShortDateTime }}
-        ({{ job.submission | jmDuration: job.end }})
+        ({{ job.start | jmDuration: job.end }})
```
rexwangcc: Not related to this PR: it seems the `job.status` and `{{ numCompletedTasks }} of {{ numTasks }} tasks completed` are in the same line now in the details page. I'm not sure is this what we want?

samanehsan: Good catch @rexwangcc. I think they should be on separate lines or there should be some separation between the two, e.g. "Succeeded: 7 out of 7 tasks completed"

samanehsan:
Thanks for taking a look @rexwangcc!

api/jobs.yaml
```diff
@@ -213,6 +213,9 @@ definitions:
       attempts:
         type: integer
         description: Number of times the task was run
+      subworkflowId:
```
calbach: There is no concept of a "workflow" in the spec currently. This should refer to a job instead... jobId, subJobId, childJobId? Something like that?

Note: jobId is already defined above to refer to the internal backend job ID - I wouldn't mind renaming that one to something like executionId though.

samanehsan: Good point! I agree with changing `jobId` to `executionId` since "jobs" mean something different in this context. And after some discussion with @dshiga, renaming`subworkflowId` to `jobId` would make sense here since it's the id of the job that corresponds to this task.

samanehsan:
Just made those changes @calbach!

ui/src/app/job-details/tasks/tasks.component.html
```diff
@@ -11,7 +11,10 @@
     <!-- Task name column -->
     <ng-container mdColumnDef="name">
       <md-header-cell *mdHeaderCellDef> Name </md-header-cell>
-      <md-cell *mdCellDef="let t">{{ t.name }}</md-cell>
+      <md-cell *mdCellDef="let t">
+        <a class="title-link" *ngIf="t.jobId" href="{{ '/jobs/' + t.jobId }}">{{ t.name }}</a>
+        <p *ngIf="!t.subworkflowId">{{ t.name }}</p>
```
calbach: update

servers/dsub/jobs/swagger/swagger.yaml
```diff
@@ -230,6 +230,9 @@ definitions:
       attempts:
         type: "integer"
         description: "Number of times the task was run"
+      jobId:
+        type: "string"
+        description: "Id of the job corresponding to this task"
```
calbach: Could you make this a bit more explicit? Too easy to confuse with the parent's job ID right now. Maybe:

..., if this task corresponds to a nested child job. Expresses a child:parent relationship between the provided job and the job containing this task.

ui/src/app/job-details/tasks/tasks.component.html
samanehsan: Ahh thanks I missed that one!

servers/dsub/jobs/swagger/swagger.yaml
samanehsan: Yeah it's confusing that the jobId corresponds to the child job/current task here, not the parent job

samanehsan:
I updated the description @calbach! Let me know what you think!

PR #162, opened 2017-12-11
[dsub] Fix missing requests import

bfcrampton:


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -20,6 +21,9 @@
 _DEFAULT_PAGE_SIZE = 64
 _MAX_PAGE_SIZE = 64
 
+NOT_FOUND_CODE = requests.codes.not_found
```
calbach: nit: I might just inline these rather than redeclaring constants

bfcrampton: Removed re-declaring and instead added a test to `test_jobs_controller_google` which exercises the error handling paths. This won't fail on CI yet but it should be fine for now.

PR #163, opened 2017-12-12
[dsub] Fix support for querying by job start

bfcrampton:
`dstat` takes in create_time as seconds from epoch which it directly passes to the pipelines API. We've been passing in a datetime which doesn't work.

### DIFF
--- a/servers/dsub/jobs/controllers/utils/query_parameters.py
+++ b/servers/dsub/jobs/controllers/utils/query_parameters.py
@@ -1,3 +1,5 @@
+import datetime
+from dateutil.tz import tzutc
 from dsub.lib import param_util
 
 import job_statuses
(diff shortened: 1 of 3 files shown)


servers/dsub/jobs/controllers/utils/query_parameters.py
```diff
@@ -13,11 +15,13 @@ def api_to_dsub(query):
...
-        'end_time': query.end,
-        'job_names': [query.name] if query.name else None,
-    }
+    dstat_params = {}
+
+    epoch = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=tzutc())
+    dstat_params['create_time'] = int(
+        (query.start - epoch).total_seconds()) if query.start else None
+
+    dstat_params['job_names'] = {query.name} if query.name else None
```
calbach: This syntax produces a list? Weird language.

servers/dsub/jobs/controllers/utils/query_parameters.py
```diff
@@ -13,11 +15,13 @@ def api_to_dsub(query):
...
     """
 
-    dstat_params = {
-        'create_time': query.start,
-        'end_time': query.end,
-        'job_names': [query.name] if query.name else None,
-    }
+    dstat_params = {}
+
+    epoch = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=tzutc())
```
calbach: Why tzutc() here, but tzlocal() above?

PR #164, opened 2017-12-14
Query Builder v1

alahwa:
Changes and new capabilities:
- Status tabs are gone, along with the statusGroup queryParam and all associated logic
- All search filters have been grouped under the 'q' queryParam (this is the generally accepted queryParam for a search query)
- Buttons for status conditionally appear if a status filter is not already present on the current list view, and perform the same function as the previous status tabs
- A header component that currently contains only the search bar is present on the list and details view
- Entering a QueryJobsRequest field or a label value creates a chip, which can be deleted or edited using the free text input field. The search button routes to the new query.

Changes pending unit tests


ui/src/app/job-list/table/table.component.html
```diff
@@ -26,7 +10,7 @@
     <md-cell *mdCellDef="let j">
       <md-checkbox class = "checkbox" [checked]="isSelected(j)" (change)="toggleSelect(j)">
       </md-checkbox>
-      <a md-button class="job-details-button" [routerLink]="[j.id]">{{ j.name }}</a>
+      <a md-button class="job-details-button" [routerLink]="[j.id]" [queryParams]="getQueryParams()">{{ j.name }}</a>
```
bfcrampton: Why do we want to pass the query parameters to the job-details page?

ui/src/app/shared/common.ts
```diff
@@ -1,21 +1,28 @@
...
+  start: 'start',
+  end: 'end',
 }
 
-export class LabelColumn {
-  header: string;
-  key: string;
-}
+/** The list of non-configurable primary columns that can be displayed and queried over. */
+export const primaryColumns: string[] = [
```
bfcrampton: nit: Uppercase `P` to match `QueryFields`?

ui/src/app/shared/common.ts
```diff
@@ -1,21 +1,28 @@
...
   Failed = <any> 'https://www.gstatic.com/images/icons/material/system/1x/close_grey600_24dp.png',
   Succeeded = <any> 'https://www.gstatic.com/images/icons/material/system/1x/done_grey600_24dp.png',
   Aborted = <any> 'https://www.gstatic.com/images/icons/material/system/1x/report_problem_grey600_24dp.png'
 }
 
-export enum StatusGroup {
-  Active = <any> 'active',
-  Failed = <any> 'failed',
-  Completed = <any> 'completed'
+/** Constant strings for non-configurable primary fields that can be queried over. Each should have a corresponding
```
bfcrampton: tiny nit: shorten this line so it doesn't wrap since you already have another line of comments.

ui/src/app/shared/header/header.component.ts
```diff
@@ -0,0 +1,148 @@
...
+      {queryParams: { q: URLSearchParamsUtils.encodeURLSearchParams(query)}}
+    );
+  }
+
+  removeChip(chipKey: string): void {
+    this.chips.delete(chipKey);
+  }
+
+  // TODO: Cut the dependency on string parsing to represent lists here
+  search(): void {
```
bfcrampton: Do we want to just automatically update after adding/removing a chip or is this the preferred behavior? I didn't think it was working until I looked at the code and saw we have to click search. Maybe we could make that button bigger?

ui/src/app/shared/header/header.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    );
+  }
+
+  setCurrentChip(chipKey: string): void {
+    this.currentChipKey = chipKey;
+    this.currentChipValue = this.chips.get(chipKey);
+  }
+
+  shouldDisplayStatusButtons(): boolean {
+    return !URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q'])[QueryFields.statuses];
```
bfcrampton: tiny nit: long line?

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      }
+      else if (key == QueryFields.start) {
+        queryRequest['start'] = new Date(urlSearchParams.get(QueryFields.start));
+      }
+      else if (key == QueryFields.end) {
+        queryRequest['end'] = new Date(urlSearchParams.get(QueryFields.end));
+      }
+      // Assume that any filters not matching a primary field should be interpreted as a label
+      else {
+        queryRequest.labels.set(key, urlSearchParams.get(key));
```
bfcrampton: This isn't getting serialized into the request properly. For example setting a `user-id` in the query field and clicking search doesn't change the request payload to the backend. Maybe its because this is a map rather than key-value?

ui/src/app/shared/header/header.component.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.filterOptions();
+    this.options = URLSearchParamsUtils.getQueryFields();
+    if (environment.additionalColumns) {
+      this.options = this.options.concat(environment.additionalColumns);
+    }
+  }
+
+  addChip(value: string): void {
+    if ((value || '').trim()) {
+      if (value.split(': ').length == 2) {
```
bfcrampton: I think if the user types `Statuses:Failed` we should auto translate that to `Statuses: Failed` so that it still works. Also maybe we should display some error for chips which do not parse to a queryable. Right now you can enter anything and there's no indication that it's invalid.

ui/src/app/shared/header/header.component.html
```diff
@@ -0,0 +1,52 @@
...
+             [mdChipInputFor]="chipList"
+             [mdChipInputSeparatorKeyCodes]="separatorKeysCodes"
+             [mdChipInputAddOnBlur]="false"
+             (mdChipInputTokenEnd)="addChip($event.value)"
+             [formControl]="control"/>
+      <md-autocomplete #auto="mdAutocomplete">
+        <md-option
+          *ngFor="let option of filteredOptions | async"
+          [value]="option"
+          (click)="addChip(option)">
```
bfcrampton: Is there a way to make it so that when you select an auto-completed key and the chip is created it goes directly to the input form to define what the value equals? It's a little annoying now to select a chip key from the list then have to click on it again to input the desired value.

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
+import {URLSearchParams} from '@angular/http';
+
+import {QueryJobsRequest} from "./model/QueryJobsRequest";
+import {JobStatus} from "./model/JobStatus";
+import {QueryFields} from "./common";
+
+/** Utilities for working with URLSearchParams*/
+export class URLSearchParamsUtils {
+
+  /** Accepts a QueryJobsRequest, and translates it reversibly into a string that can be passed as a QueryParam. */
```
bfcrampton: tiny nit: long line

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    if (request.labels) {
+      (request.labels as Map<String, String>).forEach((value: string, key: string) => {
+        urlSearchParams.set(key, value);
+      });
+    }
+    return urlSearchParams.toString();
+  }
+
+  /** Accepts a param map, and translates it reversibly into a string that can be passed as a QueryParam. */
```
bfcrampton: tiny nit: long line

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      if (values.length == 1) {
+        urlSearchParams.set(key, values[0]);
+      } else {
+        values.forEach(value => urlSearchParams.append(key, value));
+      }
+    });
+    return urlSearchParams.toString();
+  }
+
+  /** Accepts a string query that was generated via encodeUrlSearchParams and converts it back into a QueryJobsRequest. */
```
bfcrampton: tiny nit: long line

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+  public static unpackURLSearchParams(query: string): QueryJobsRequest {
+    let urlSearchParams = new URLSearchParams(query);
+    let queryRequest = {
+      labels: new Map(),
```
bfcrampton: Why did you chose a `Map` over key value object?

ui/src/app/shared/url-search-params.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+      }
+      else if (key == QueryFields.end) {
+        queryRequest['end'] = new Date(urlSearchParams.get(QueryFields.end));
+      }
+      // Assume that any filters not matching a primary field should be interpreted as a label
+      else {
+        queryRequest.labels.set(key, urlSearchParams.get(key));
+      }
+    });
+
```
bfcrampton: ✂️  blank line

ui/src/app/shared/url-search-params.utils.ts
alahwa: See other comment

ui/src/app/job-list/table/table.component.html
bfcrampton: Ah okay I didn't realize the query builder would be on the details page, but that makes sense.

ui/src/app/shared/common.ts
bfcrampton: Yup thanks.

ui/src/app/shared/header/header.component.html
bfcrampton: SGTM

ui/src/app/shared/header/header.component.ts
bfcrampton: Sounds good, definitely don't do a lot of work to make the search button bigger if we might remove it at some point.

ui/src/app/shared/url-search-params.utils.ts
bfcrampton: The dsub shim does have support for querying by labels and I'm pretty sure its working correctly––all the [tests](https://github.com/DataBiosphere/job-manager/blob/master/servers/dsub/jobs/test/base_test_cases.py#L288) are passing. Could you give a little bit more info about the 500 error you were seeing?

Which labels have you tried querying by, only `user-id` `job-id` and other `dsub` default ones ? `['task-id', 'job-id', 'user-id', 'job-name']` are reserved labels in dsub, so if I had to guess it is probably because we're trying to query them like normal labels, when we need to have a specific handling. A label like 'status-detail' should work, however.

I don't think this PR should fix the issue in the backend, but you should document the bug so that it can be fixed and the code which serializes labels into the request should function. Using the `Map` definitely will not work since its not even serialized into the request to the shim.

ui/src/app/shared/header/header.component.ts
bfcrampton: I don't really have any good suggestions, but I think that makes sense to have a UX discussion around it before implementing anything.

ui/src/app/shared/url-search-params.utils.ts
alahwa: I'm not familiar with the dsub repo and github tools for searching through code are non-existent so I haven't yet found the source of this error, but here's all the debugging information I have:

detail:"The server encountered an internal error and was unable to complete your request.  Either the server is overloaded or there is an error in the application."
status:500
title:"Internal Server Error"
type:"about:blank"

I can confirm this happens with status-detail as well as the reserved labels.

bfcrampton: Okay would you mind filing a github issue and adding a TODO wherever you disabled the non-supported fields. I'll try to figure out the issue and re-enable in another PR.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -32,18 +32,18 @@ export class JobListComponent implements OnInit {
   constructor(
     private readonly route: ActivatedRoute,
     private readonly router: Router,
-    private readonly JobManagerService: JobManagerService,
+    private readonly jobManagerService: JobManagerService,
     private readonly viewContainer: ViewContainerRef,
     private errorBar: MdSnackBar,
   ) {
-    this.jobStream = new JobStream(JobManagerService, StatusGroup.Active);
+  route.queryParams.subscribe(params => this.reloadJobs(params['q']));
```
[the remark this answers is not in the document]
alahwa: Done

ui/src/app/job-list/job-list.component.ts
```diff
@@ -32,18 +32,19 @@ export class JobListComponent implements OnInit {
...
   ) {
-    this.jobStream = new JobStream(JobManagerService, StatusGroup.Active);
+  route.queryParams.subscribe(
+      params => this.reloadJobs(params['q']));
   }
 
   ngOnInit(): void {
     if (this.route.snapshot.data['stream']) {
       this.jobStream = this.route.snapshot.data['stream'];
+      this.streamSubscription = this.jobStream.subscribe(resp => this.jobs.next(resp));
```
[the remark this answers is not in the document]
alahwa: Done

ui/src/app/job-list/table/table.component.html
```diff
@@ -95,7 +79,7 @@
 
 <div class="group-options" *ngIf="selectedJobs.length > 0">
   {{ selectedJobs.length }} jobs selected
-  <button md-raised-button color="accent" class="group-button group-abort" *ngIf="currentStatusGroup==statusGroup.Active"
+  <button md-raised-button color="accent" class="group-button group-abort"
```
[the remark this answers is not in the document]
alahwa: Went with the easiest fix for the latter behavior. Again, this will be a UX preference so we can circle back later if this isn't the preferred behavior.

ui/src/app/job-list/table/table.component.ts
```diff
@@ -41,35 +39,18 @@ import {environment} from '../../../environments/environment';
...
   // TODO(alanhwang): Allow these columns to be configured by the user
-  displayedColumns = [
-    'Job',
-    'Status',
-    'Submitted',
-  ];
+  displayedColumns = primaryColumns;
 
   @ViewChild(MdPaginator) paginator: MdPaginator;
   @ViewChild('filter') filter: ElementRef;
```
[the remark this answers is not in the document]
alahwa: Nope, removed

PR #165, opened 2017-12-14
Update page_tokens to support created_before + offset_id tokens

bfcrampton:
This is a prerequisite for implementing the fixed `dsub` pagination strategy which is based upon querying the Pipelines API with a CreatedBefore parameter.

The only alteration to the behavior for cromwell is the change in `encode` and `decode` method names, and that `page_tokens` returns `None` rather than `0` when trying to decode a `None`. I updated cromwell's `jobs_controller` to handle that. 

I also removed a duplication of:
```
if not page_size:
    page_size = _DEFAULT_PAGE_SIZE
```
in `page_from_offset`. It seems redundant to check that twice rather than do it at right away and know it is present from there forth. There was an associated test that `page_from_offset` can handle `None` which I removed.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -163,10 +163,8 @@ def query_jobs(body):
     """
     query = QueryJobsRequest.from_dict(body)
 
-    page_size = query.page_size
-    if not page_size:
-        page_size = _DEFAULT_PAGE_SIZE
-    offset = page_tokens.decode(query.page_token)
+    page_size = query.page_size or _DEFAULT_PAGE_SIZE
+    offset = page_tokens.decode_offset(query.page_token) or 0
     page = page_from_offset(offset, page_size)
     params_for_cromwell = cromwell_query_params(query, page, page_size)
 
@@ -179,14 +177,12 @@ def query_jobs(body):
     results.reverse()
 
     next_offset = offset + page_size
-    next_page_token = page_tokens.encode(next_offset)
+    next_page_token = page_tokens.encode_offset(next_offset)
 
     return QueryJobsResponse(results=results, next_page_token=next_page_token)
 
 
 def page_from_offset(offset, page_size):
-    if not page_size:
-        page_size = _DEFAULT_PAGE_SIZE
     return 1 + offset / page_size
 
 
--- a/servers/jm_utils/jm_utils/page_tokens.py
+++ b/servers/jm_utils/jm_utils/page_tokens.py
@@ -1,37 +1,38 @@
+"""
+We implement the pagination token via base64-encoded JSON s.t. tokens are
+opaque to clients and enable us to make backwards compatible changes to our
+pagination implementation. Base64+JSON are used specifically as they are
+easily portable across language.
+"""
 import base64
 import datetime
 import json
 import numbers
 
-# We implement the pagination token via base64-encoded JSON s.t. tokens are
-# opaque to clients and enable us to make backwards compatible changes to our
-# pagination implementation. Base64+JSON are used specifically as they are
-# easily portable across language.
-
 
 def _encode(dictionary):
-    """Encodes any arbitrary dictionary into a jobs pagination token.
+    """Encodes any arbitrary dictionary into a pagination token.
 
     Args:
         dictionary: (dict) Dictionary to basee64-encode
     
     Returns:
-        (string) encoded page token representing a page of jobs
+        (string) encoded page token representing a page of items
     """
     # Strip ugly base64 padding.
     return base64.urlsafe_b64encode(json.dumps(dictionary)).rstrip('=')
 
 
 def _decode(token):
-    """Decodes the jobs pagination token.
+    """Decodes the pagination token.
 
     Args:
         token: (string) Base64 encoded JSON pagination token
 
     Returns:
-        (dict) The token dictionary representing a page of jobs.
+        (dict) The token dictionary representing a page of items.
     """
-    if not token:
+    if token is None:
         return None
     # Pad the token out to be divisible by 4.
     padded_token = token + '=' * (4 - (len(token) % 4))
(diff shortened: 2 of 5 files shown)


servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
...
+    """Decodes the jobs offset-pagination token.
+
+    Args:
+        token: (string) base64 encoded JSON offset-pagination token
+
+    Returns:
+        (number) the pagination offset, defaults to 0 if token is None
+    """
+    token_dict = _decode(token)
+    if not token_dict:
```
calbach: I'd check against `None` explicitly; otherwise there's an edge case here for an input of `encode({})`

servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
...
+    if not isinstance(created_before, datetime.datetime):
+        raise ValueError(
+            'Invalid create time must be datetime: {}'.format(created_before))
+    if offset_id and not isinstance(offset_id, basestring):
+        raise ValueError(
+            'Invalid offset id must be string: {}'.format(offset_id))
+
+    epoch = datetime.datetime.utcfromtimestamp(0)
+    seconds_epoch = int((created_before - epoch).total_seconds())
+    token_dict = {'created-before': seconds_epoch}
```
calbach: Please use shorter key names (1-2 chars ideally), each character here adds ~1.33 characters to the token length

servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
...
+        raise ValueError('Invalid offset must integer > 0: {}'.format(offset))
+    return _encode({'of': offset})
+
+
+def encode_created_before(created_before, offset_id=None):
+    """Encodes a created_before and optional offset job ID into a jobs pagination token.
+    
+    Args:
+      created_before: (datetime) The create-time of the first job to include in the associated page.
+      offset_id: (str) The unique job-id of the first job to include in the associated page. Needed
```
calbach: Does this refer to the common Jobs API id? Or something else?

servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
 import base64
+import datetime
 import json
 import numbers
 
+# We implement the pagination token via base64-encoded JSON s.t. tokens are
```
calbach: nit: perhaps move into package comment instead (`"""` on the first line of the file)?

servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
...
+
+def decode_created_before(token):
+    """Decode a created_before and optional offset job ID pagination token.
+
+    Args:
+        token: (string) base64 encoded JSON offset-pagination token
+
+    Returns:
+        (datetime) The create-time of the first job to include in this page.
+        (str) The unique job-id of the first job to include in this page.
```
calbach: , if any

servers/jm_utils/jm_utils/page_tokens.py
```diff
@@ -1,44 +1,137 @@
...
+
+
+def decode_offset(token):
+    """Decodes the jobs offset-pagination token.
+
+    Args:
+        token: (string) base64 encoded JSON offset-pagination token
+
+    Returns:
+        (number) the pagination offset, defaults to 0 if token is None
```
calbach: update

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: Yup that's better.

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: I meant it as the API job-id, but really it can be any unique identifier for a pagination item. These page tokens really don't need to be job-specific, as they could definitely be re-utilized for things like paginating tasks. I removed all the job-specific language from the documentation in favor of generic `items`.

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: Right makes sense why offset is `of` now 😄

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: Fixed.

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: Done, here and everywhere else.

servers/jm_utils/jm_utils/page_tokens.py
bfcrampton: Added.

bfcrampton:
@rexwangcc PTAL

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -163,10 +163,8 @@ def query_jobs(body):
     """
     query = QueryJobsRequest.from_dict(body)
 
-    page_size = query.page_size
-    if not page_size:
-        page_size = _DEFAULT_PAGE_SIZE
-    offset = page_tokens.decode(query.page_token)
+    page_size = query.page_size or _DEFAULT_PAGE_SIZE
```
rexwangcc: Nice change here, we would change the Cromwell controller to this one-liner as well.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -163,10 +163,8 @@ def query_jobs(body):
...
+    offset = page_tokens.decode_offset(query.page_token) or 0
```
rexwangcc: Thanks for the changes here!

bfcrampton: :smile:

PR #166, opened 2017-12-14
[CI] Run google provider integration tests with service account

bfcrampton:


.circleci/config.yml
```diff
@@ -47,15 +45,20 @@ jobs:
...
           name: Run dsub server tests
           command: |
-            tox -c ./servers/dsub -- \
-              --exclude-test=jobs.test.test_jobs_controller_google.TestJobsControllerGoogle
+            if [ -z "$GCLOUD_SERVICE_KEY" ]; then
+              tox -c ./servers/dsub -- \
+                --exclude-test=jobs.test.test_jobs_controller_google.TestJobsControllerGoogle
+            else
+              echo $GCLOUD_SERVICE_KEY | base64 --decode --ignore-garbage > /home/circleci/gcloud-service-key.json
+              tox -c ./servers/dsub --
```
calbach: nit: no longer need `--`

bfcrampton:
@calbach slightly modified `base_test_cases.py`, you mind taking another quick glance?

.circleci/config.yml
bfcrampton: Done

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -306,12 +306,6 @@ def test_query_jobs_by_label(self):
             no_label_job = self.start_job('echo NO_LABEL', name='nolabeljob')
             no_label_job_id = self.get_api_job_id(no_label_job)
```
calbach: Would this change make the local provider flaky? Can't we not query local jobs until they've started? Another idea would be to support multiple statuses on wait_for_job_status

bfcrampton: I do not believe it will make them flakey. Querying by local jobs should now work even before they've started (after https://github.com/googlegenomics/dsub/issues/82 was resolved). This is separate from the issue from being unable to abort jobs local jobs which are not yet started.

PR #167, opened 2017-12-19
Add custom paginator logic for non-exhaustive job lists

calbach:


ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -168,6 +168,25 @@ describe('JobsTableComponent', () => {
       .toEqual("Precondition Failed (412): Job already in terminal status `FAILED` Dismiss");
   }))
 
+  it('should only show length for exhaustive job streams', async(() => {
```
alahwa: nit: this could be two unit tests

calbach: I wanted to explicitly cover the transition between non-exhaustive -> exhaustive to make sure the text updated appropriately. Added a comment.

PR #170, opened 2018-01-03
[dsub] Implement create-time based pagination strategy

bfcrampton:
Updated: Thanks to @calbach's suggestion of using the dstat job producer generator directly, this implementation has been cleaned up significantly.

The basic premise is:
- Paginate based on a job's `create-time`. `dstat` returns jobs in that order, but it is very much possible to have two jobs with an identical create time (only second granularity). In this case:
  - Get all the jobs with that `create-time`
  - Sort all those jobs by `job-id` + `task-id` (unique to a job)
  - Use this combined ID as an offset indicator for the next pagination (i.e. if there are multiple pages for a given `create-time`, always download all the jobs with that create-time and use the ID as an offset to know where you left off)
- If the query contains a `start` parameter, only paginate back to that time


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -119,55 +120,137 @@ def query_jobs(body):
...
+    token = page_tokens.encode_create_time_max(create_time_max, offset_id)
+    return QueryJobsResponse(results=results, next_page_token=token)
+
+    # If we're trying to paginate earlier than the query's minimum start time,
+    # don't return anything
+    if query.start and create_time_max and query.start > create_time_max:
+        return []
+
+
+def _query_dstat_jobs(provider, query, create_time_max, max_tasks=0):
```
calbach: OK, so here is how I originally was thinking this **could** work in my mind, I'm thinking this could simplify the above logic (though perhaps would sacrifice some performance):

API layer: Requests pageSize+2 from generator
Generator layer (this helper method): generator function which produces an ordered stream of dsub jobs (by created, job-id+taskid)

```
def _generate_dstat_jobs():
  jobs = dstat.producer()
  last_create_time = None
  job_buffer = []
  for j in jobs:
    job_buffer.append(j)
    if j['create-time'] != last_create_time:
      for b in sorted(job_buffer, key=...): yield b
      job_buffer = []
    last_create_time = j['create-time']
  for b in sorted(job_buffer, key=...): yield b
```

Possibly this layer could also filter out according to create_time_max/offset_id. A downside is that you'd be getting rid of `max_tasks`, which might result in excessive overscanning.

The upside is that the API layer should be drastically simplified, in that it needn't worry about sorting or doing a two-phase request.

Thoughts?

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -179,18 +262,28 @@ def _auth_token():
     return None
 
 
-def _client():
-    return current_app.config['CLIENT']
+def _get_offset_id_index(offset_id, sorted_jobs):
+    job_ids = [j['job-id'] + j.get('task-id', '') for j in sorted_jobs]
```
calbach: Extract to a helper? This job ID key is used in multiple places.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -49,13 +49,14 @@ def assert_query_matches(self, query_params, job_list):
...
             response = self.must_query_jobs(query_params)
-            self.assertEqual(len(response.results), len(job_list))
             sorted_results = sorted(
                 response.results, key=operator.attrgetter('id'))
+            self.assertEqual(len(response.results), len(job_list))
             sorted_job_list = sorted(job_list, key=self.get_api_job_id)
-            for result, job in zip(sorted_results, sorted_job_list):
-                self.assertEqual(result.id, self.get_api_job_id(job))
-                self.assertEqual(result.labels['user-id'], job['user-id'])
+            # for result, job in zip(sorted_results, sorted_job_list):
```
calbach: Uncomment or remove?

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Done

servers/dsub/jobs/test/base_test_cases.py
bfcrampton: Done

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -123,51 +124,127 @@ def query_jobs(body):
...
-    # in a subsequent page.
-    offset = page_tokens.decode_offset(query.page_token) or 0
-    max_tasks = offset + query.page_size + 1
+    create_time_max, offset_id = page_tokens.decode_create_time_max(
+        query.page_token) or (None, None)
 
     jobs = []
+    if offset_id and create_time_max:
+        if query.start and query.start > create_time_max:
+            raise ValueError(
```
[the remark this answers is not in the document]
bfcrampton: Done.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -123,51 +124,127 @@ def query_jobs(body):
...
-    # Request one extra job to confirm whether there's more data to return
-    # in a subsequent page.
-    offset = page_tokens.decode_offset(query.page_token) or 0
-    max_tasks = offset + query.page_size + 1
+    create_time_max, offset_id = page_tokens.decode_create_time_max(
+        query.page_token) or (None, None)
 
     jobs = []
+    if offset_id and create_time_max:
+        if query.start and query.start > create_time_max:
```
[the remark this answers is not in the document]
bfcrampton: Done.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Great suggestion! It required adding one additional method to `dstat` to expose the underlying job generator but very much worth it to greatly simplify the logic in the shim. This looks much cleaner and readable now, and likely much less error-prone.

Let me know what you think.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -114,63 +116,41 @@ def query_jobs(body):
...
-
-    results = [_query_result(j, query.parent_id) for j in jobs]
-    return QueryJobsResponse(results=results, next_page_token=next_page_token)
+    try:
+        next_job = job_generator.next()
+        next_ct = next_job['create-time']
+        last_ct = jobs[-1]['create-time']
+        offset_id = _JOB_SORT_KEY(next_job) if next_ct == last_ct else None
+        return _get_query_jobs_response(jobs, query.parent_id,
+                                        next_job['create-time'], offset_id)
```
calbach: next_ct

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -182,33 +162,62 @@ def _auth_token():
...
+        task_ids=dstat_params.get('task_ids'),
+        create_time_min=dstat_params.get('create_time'),
+        create_time_max=create_time_max,
+        job_names=dstat_params.get('job_names'),
+        labels=dstat_params.get('labels')))
+
+    last_create_time = None
+    job_buffer = []
+    for job in jobs:
+        job['create-time'] = job['create-time'].replace(microsecond=0)
```
calbach: Comment?

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -182,33 +162,62 @@ def _auth_token():
...
+        # If this job is from the last page, skip it and continue generating
+        if create_time_max and job['create-time'] == create_time_max:
+            if offset_id and _JOB_SORT_KEY(job) < offset_id:
+                continue
+
+        # Build up a buffer of jobs with the same create time. Once we get a
+        # job with an older create time we yeild all the jobs in the buffer
```
calbach: yield

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -182,33 +162,62 @@ def _auth_token():
...
+        # sorted by job-id + task-id
+        job_buffer.append(job)
+        if job['create-time'] != last_create_time:
+            for j in sorted(job_buffer, key=_JOB_SORT_KEY):
+                yield j
+            job_buffer = []
+        last_create_time = job['create-time']
+
+    # If we hit the end of the dstat job generator, ensure to yeild the jobs
```
calbach: yield

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -362,3 +363,58 @@ def test_query_jobs_by_start(self):
...
+            job2 = self.start_job('echo SECOND_JOB', name='job_y')
+            job3 = self.start_job('echo THIRD_JOB', name='job_x')
+            job4 = self.start_job('echo FOURTH_JOB', name='job_w')
+            job5 = self.start_job('echo FIFTH_JOB', name='job_v')
+            job6 = self.start_job('echo SIXTH_JOB', name='job_u')
+            job7 = self.start_job('echo SEVENTH_JOB', name='job_t')
+            job8 = self.start_job('echo EIGHTH_JOB', name='job_s')
+            job9 = self.start_job('echo NINTH_JOB', name='job_r')
+            job10 = self.start_job('echo TENTH_JOB', name='job_q')
+            time.sleep(2)
```
calbach: Comment? Is this for local dsub?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -362,3 +363,58 @@ def test_query_jobs_by_start(self):
                 QueryJobsRequest(start=third_time), [third_job, fourth_job])
             self.assert_query_matches(
                 QueryJobsRequest(start=fourth_time), [fourth_job])
+
+        def test_query_jobs_pagination(self):
```
calbach: 10 jobs might crowd our GCE quota; maybe move this down to the local provider only?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -362,3 +363,58 @@ def test_query_jobs_by_start(self):
...
+            # Jobs are sorted first by create-time then by job-id. We cannot
+            # guarentee these start at the exact same second, but we know some
```
calbach: guarantee

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -114,63 +116,41 @@ def query_jobs(body):
...
-
-    # Request one extra job to confirm whether there's more data to return
-    # in a subsequent page.
-    offset = page_tokens.decode_offset(query.page_token) or 0
-    max_tasks = offset + query.page_size + 1
-
+    create_time_max, offset_id = page_tokens.decode_create_time_max(
+        query.page_token) or (None, None)
+    query.page_size = min(query.page_size or _DEFAULT_PAGE_SIZE,
+                          _DEFAULT_PAGE_SIZE)
```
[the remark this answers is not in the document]
bfcrampton: Yup that's correct. I actually think it makes sense to modestly lower the default and raise the max, if you're fine with that.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Done.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Added

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Done

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Done

servers/dsub/jobs/test/base_test_cases.py
bfcrampton: Done.

servers/dsub/jobs/test/base_test_cases.py
bfcrampton: Actually it's not needed, good catch.

servers/dsub/jobs/test/base_test_cases.py
bfcrampton: Trimmed down to 5 jobs and page size of 2, which I think should still be sufficient to cover the edge cases with less pressure on our quota. Let me know if you think it's still too many.

calbach: Seems good

bfcrampton:
I still need to make a new dsub build before merging, but fixed the generator issue so ready for another look.

bfcrampton:
The performance is actually quite good now! Much improved compared to the previous pagination strategy. We can now successfully paginate to the end of the project used for these integration tests (5000+ jobs) with very little delay on each request.

Updated the PR description.

PR #175, opened 2018-01-05
[dsub] Support querying by reserved labels and URL-prefixed projects fix

### LINKED ISSUE #157: Details page fails on organization prefixed projects
e.g. google.com:foo ->

```
"Job ID format for google provider is: <project-id>:<job-id>[:<task-id>]?", status: 400,…}
```

Also, I didn't see an error pop-up for this issue.

### LINKED ISSUE #169: [dsub] Allow querying by user-id, job-id, task-id

bfcrampton:
### Resolves #169 
- Support querying by `user-id`, `job-id`, and `task-id` labels. These are considered first class fields in `dstat`, but are jammed into `labels` in the common API. We have to translate back when querying by them.
- Add tests validating behavior and ensuring querying multi-tasks jobs works
- Also add `job-id` and `task-id` columns to the jobs list page. This will allow users to more easily construct queries for say all the running tasks left on a single dsub job. 
### Resolves #157 
- Use a `+` rather than `:` to join components since projects with a URL prefix use a `:` to join.

### DIFF
--- a/servers/dsub/jobs/controllers/utils/query_parameters.py
+++ b/servers/dsub/jobs/controllers/utils/query_parameters.py
@@ -29,18 +29,13 @@ def api_to_dsub(query):
     if query.name:
         dstat_params['job_names'] = {query.name}
     if query.labels:
-        dstat_params['job_ids'] = {
-            v
-            for (k, v) in query.labels.items() if k == 'job-id'
-        }
-        dstat_params['task_ids'] = {
-            v
-            for (k, v) in query.labels.items() if k == 'task-id'
-        }
-        dstat_params['user_ids'] = {
-            v
-            for (k, v) in query.labels.items() if k == 'user-id'
-        }
+        if query.labels.get('job-id'):
+            dstat_params['job_ids'] = {query.labels['job-id']}
+        if query.labels.get('task-id'):
+            dstat_params['task_ids'] = {query.labels['task-id']}
+        if query.labels.get('user-id'):
+            dstat_params['user_ids'] = {query.labels['user-id']}
+
         dstat_params['labels'] = {
             param_util.LabelParam(k, v)
             for (k, v) in query.labels.items()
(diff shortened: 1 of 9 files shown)


servers/dsub/jobs/controllers/utils/query_parameters.py
```diff
@@ -17,20 +17,34 @@ def api_to_dsub(query):
...
+    if query.start:
+        epoch = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=tzutc())
+        dstat_params['create_time'] = int(
+            (query.start - epoch).total_seconds())
+    if query.name:
+        dstat_params['job_names'] = {query.name}
+    if query.labels:
+        dstat_params['job_ids'] = {
+            v
+            for (k, v) in query.labels.items() if k == 'job-id'
```
calbach: Isn't query.labels a dict? Why not just index into it by `'job-id'`?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -73,8 +71,14 @@ def expected_log_files(self, job_id):
             }
 
         def get_api_job_id(self, dsub_job):
-            if self.testing_project:
-                return '{}:{}'.format(self.testing_project, dsub_job['job-id'])
+            if self.testing_project and dsub_job.get('task-id'):
```
calbach: Can you call job_ids.dsub_to_api here?

servers/dsub/jobs/controllers/utils/query_parameters.py
bfcrampton: Good point, I was thinking there could be multiple job-id's, but that isn't supported by our API right now. Switched to index directly for each reserved label.

servers/dsub/jobs/test/base_test_cases.py
bfcrampton: Yup, good call fixed.

PR #177, opened 2018-01-08
Only list parent jobs in jobs list

samanehsan:
The cromwell query endpoint returns both parent workflows and subworkflows, but as a user it is clearer to only see top-level workflows in the jobs list view and then click on job details to see its subworkflows/tasks.

This PR updates the query jobs endpoint in the cromwell shim layer so that cromwell includes the `parentWorkflowId` for jobs in the returned metadata. Then only jobs without a `parentWorkflowId` are returned. 

Note: This isn't an issue since we're using pagination, but when specifying fields to include in the metadata response with `additionalQueryResultFields`, cromwell looks up that field in its database for each item in the query result. And if there are over 1,000 results it will throw a 400 error because the database can't handle that many requests.


api/jobs.yaml
```diff
@@ -346,6 +346,9 @@ definitions:
       labels:
         type: object
         description: Custom job labels with string values
+      parentJobId:
```
bfcrampton: I think I would prefer this to be `parentId` to match the field on `QueryJobsRequest`, unless you think this is representing something fundamentally different. This way the field isn't tied to the identity of the parent resource. For example `dsub` jobs could be populated with the parent project ID, but there isn't really a concept of parent 'jobs' per se for `dsub`.

bfcrampton:
Looks good just one naming question. Also looks like you'll need a rebase and CI is failing because you need a `yapf` lint.

api/jobs.yaml
samanehsan: @bfcrampton does `parentId` for dsub jobs refer to the google cloud project id? @dshiga mentioned some teams at the Broad divide cromwell jobs between different google projects as a way to distribute load so we may want to include that field for cromwell jobs as well.

bfcrampton: @samanehsan That would be totally fine with me! I would just note that if you expect to need querying by a job's parent Job ID then we may need to add `parentJobId` to `QueryJobsRequest` as well. We do not currently need for `parentProjectId` on `QueryJobsResult` since all jobs have to be scoped by project.

samanehsan: Sounds good @bfcrampton! I'd rather rename it in a follow-up PR if that's fine with you to keep those changes separate. 

I can't think of a use case where a user would want all jobs with a certain parent id, since you can see that when you go to a parent job's details page. Also this PR will filter out all subworkflows from the jobs list so if a user tried that query it wouldn't show them any results.

Also are you able to restart the Circle CI build? It's still giving me yapf linting errors but when I run the lint command there's nothing to fix!

bfcrampton: Yup that sounds good to me, thanks! I am able to restart it but I don't think that's the issue. Most likely you have an older version of `yapf` and CI always gets the newest (we should probably be pinning to a yapf version, but we aren't). Would you mind you try updating `yapf` locally to see if that fixes it?

samanehsan: Updating `yapf` fixed it! Thanks! Also I put a ticket in the backlog so we don't forget to rename the `parentId` field later.

samanehsan:
The CI is finally passing so I'll go ahead and merge this! :octocat:

PR #184, opened 2018-01-16
[dsub] Implement create-time based pagination strategy

bfcrampton:
Re-opening of #170 now that [`dsub v0.1.5`](https://github.com/googlegenomics/dsub/releases/tag/v0.1.5) has been released.


servers/dsub/requirements.txt
```diff
@@ -4,7 +4,7 @@ chardet==3.0.4
 click==6.7
 clickclick==1.2.2
 connexion==1.1.13
-dsub==0.1.4
+dsub==0.1.5
```
calbach: This is the only change from the last PR right? If so, LGTM

bfcrampton: Yup (as well as the change in requirements-to-freeze). This is the same branch I just amended one commit on top with the version bump.

PR #186, opened 2018-01-17
[dsub] Implement querying by end date

### LINKED ISSUE #73: Support querying by start and end times
- [x] Query by `start` (actually done by create not submission https://github.com/bvprivate/job-monitor/issues/54)
- [ ] Query by `end` (client side until supported by dsub)
- [ ] Query by `last-update` (client side)

bfcrampton:
### Resolves #73:
- Implement querying by end datetime
- Minor cleanup in tests to shorten a couple function names and make it more readable

### DIFF
--- a/servers/dsub/jobs/controllers/jobs_controller.py
+++ b/servers/dsub/jobs/controllers/jobs_controller.py
@@ -172,8 +172,8 @@ def _generate_dstat_jobs(provider, query, create_time_max=None,
     # If create_time_max is not set, but we have to client-side filter by
     # end-time, set create_time_max = query.end because all the jobs with
     # create_time >= query.end cannot possibly match the query.
-    if not create_time_max and query.start:
-        create_time_max = query.start
+    if not create_time_max and query.end:
+        create_time_max = query.end
 
     dstat_params = query_parameters.api_to_dsub(query)
     jobs = execute_redirect_stdout(lambda: dstat.lookup_job_tasks(
@@ -190,7 +190,7 @@ def _generate_dstat_jobs(provider, query, create_time_max=None,
     last_create_time = None
     job_buffer = []
     for job in jobs:
-        if query.end and job['end-time'] > query.end:
+        if query.end and ('end-time' not in job or job['end-time'] > query.end):
             continue
 
         # The LocalJobProvider returns datetimes with milliescond granularity.
--- a/servers/dsub/jobs/controllers/jobs_controller.py
+++ b/servers/dsub/jobs/controllers/jobs_controller.py
@@ -129,14 +129,20 @@ def query_jobs(body):
             microsecond=0)
         if create_time_max and query.start > create_time_max:
             raise BadRequest(
-                "Invalid pagination token with query start parameter.")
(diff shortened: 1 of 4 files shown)


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -129,14 +129,21 @@ def query_jobs(body):
...
+            raise BadRequest(
+                "Invalid query: start date must precede end date.")
 
     job_generator = _generate_dstat_jobs(provider, query, create_time_max,
                                          offset_id)
     jobs = []
     try:
         for job in job_generator:
-            jobs.append(job)
+            if not query.end or job.get(
```
calbach: nit: I would probably invert and `continue` instead, which should set up better for additional filter criteria.

calbach: I think you want this inside the job generator function instead. Consider the following test case:

request: pageSize of 2, end=10
jobs: {end: 4}, {end: 5}, {end: 11}

want: page of 2 elements
got: page of 2 elements, page of 0 elements

calbach:
Per offline discussion, should also set `create_time_max` on the initial request if `query.end` is given.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Done (in generator function)

bfcrampton: Good point didn't think about it returning page tokens when it shouldn't. Fixed.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -164,6 +169,12 @@ def _auth_token():
 
 def _generate_dstat_jobs(provider, query, create_time_max=None,
                          offset_id=None):
+    # If create_time_max is not set, but we have to client-side filter by
+    # end-time, set create_time_max = query.end because all the jobs with
+    # create_time >= query.end cannot possibly match the query.
+    if not create_time_max and query.start:
```
calbach: hmmm.. I thought this was supposed to be query.end?

Do the tests fail?

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -179,6 +190,9 @@ def _generate_dstat_jobs(provider, query, create_time_max=None,
     last_create_time = None
     job_buffer = []
     for job in jobs:
+        if query.end and job['end-time'] > query.end:
```
calbach: Wouldn't this fail for running jobs, i.e. without an end time?

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Whoops you're definitely should be `query.end`, should've run the tests again to make sure they still pass.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Yup, fixed.

PR #187, opened 2018-01-18
Fix submission time in UI, always populate Cromwell shim submission

calbach:
Fallback to the current time for workflow submission time as a last resort, if submission or start time are not available from Cromwell.

Possible related Cromwell issues:
- Cromwell queryJobs swagger marks `started` as required, which clearly cannot be the case
- Cromwell should probably return `submission` from queryJobs
- Cromwell should probably guarantee return of `submission` time, even from subworkflows

Alternatives considered:
- If submission time is missing, exclude it from the results altogether -> this seemed less favorable to faking the submission time
- Make submission time not required in the common Jobs API -> I think this is a minimal requirement we should make in the API as it affects other aspects such as sorting order

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -104,10 +104,10 @@ def get_job(id):
     start = _parse_datetime(job.get('start'))
     submission = _parse_datetime(job.get('submission'))
     if submission is None:
-      # Submission is required by the common jobs API. Submission may be missing
-      # for subworkflows in which case we fallback to the workflow start time
-      # or, if not started, the current time.
-      submission = start or datetime.utcnow()
+        # Submission is required by the common jobs API. Submission may be missing
+        # for subworkflows in which case we fallback to the workflow start time
+        # or, if not started, the current time.
+        submission = start or datetime.utcnow()
     return JobMetadataResponse(
         id=id,
         name=job.get('workflowName'),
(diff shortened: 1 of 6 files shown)


rexwangcc:
This is the issue we talked about this afternoon. @jsotobroad

samanehsan:
Thanks for making these changes! I just have a few comments about the possible cromwell issues you mentioned in the PR description: 

> Cromwell queryJobs swagger marks started as required, which clearly cannot be the case

That's a good point, since jobs are not immediately started in Cromwell and that field isn't always there. Maybe we can make a ticket to go back and change this?

> Cromwell should probably return submission from queryJobs

I think this would be possible to fix in Cromwell and is definitely much better than faking a submission time.

> Cromwell should probably guarantee return of submission time, even from subworkflows

I opened a github issue in Cromwell about including submission time in subworkflows, and there was some discussion around how subworkflows are considered to be more like workflow tasks that get called:
https://github.com/broadinstitute/cromwell/issues/2968

samanehsan:
I think this is good to merge after updating with the master branch

calbach:
> That's a good point, since jobs are not immediately started in Cromwell and that field isn't always there. Maybe we can make a ticket to go back and change this?

Sent https://github.com/broadinstitute/cromwell/pull/3168

> > Cromwell should probably return submission from queryJobs
>
> I think this would be possible to fix in Cromwell and is definitely much better than faking a submission time. But I think the solution in your PR would work as a temporary fix.

Filed https://github.com/broadinstitute/cromwell/issues/3167 and addded comment.

> > Cromwell should probably guarantee return of submission time, even from subworkflows
>
> I opened a github issue in Cromwell about including submission time in subworkflows, and there was some discussion around how subworkflows are not submitted in the same way that a workflow is submitted: broadinstitute/cromwell#2968. I think the submission time would always be the same as the start time for subworkflows/tasks so it's not as much of an issue in this case.

Thanks - added this to the comment.

PR #188, opened 2018-01-19
[dsub] Support querying by 'Submitted' status

### LINKED ISSUE #140: QueryJobs filter should account for pending/running
Just realized that the latest dsub upgrade changed the status, but didn't have the corresponding client-side filtering to distinguish pending from running via the QueryJobs method: https://github.com/bvprivate/job-monitor/pull/134

@bfcrampton fyi

bfcrampton:
Resolves #140

### DIFF
--- a/servers/dsub/jobs/controllers/jobs_controller.py
+++ b/servers/dsub/jobs/controllers/jobs_controller.py
@@ -22,7 +22,6 @@ from jobs.models.query_jobs_result import QueryJobsResult
 
 _DEFAULT_PAGE_SIZE = 32
 _MAX_PAGE_SIZE = 256
-_JOB_SORT_KEY = lambda j: j['job-id'] + j.get('task-id', '')
 
 
 def abort_job(id):
@@ -136,8 +135,7 @@ def query_jobs(body):
             raise BadRequest(
                 "Invalid query: start date must precede end date.")
 
-    job_generator = _generate_dstat_jobs(provider, query, create_time_max,
-                                         offset_id)
+    job_generator = _generate_jobs(provider, query, create_time_max, offset_id)
     jobs = []
     try:
         for job in job_generator:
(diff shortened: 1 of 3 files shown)


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -175,6 +175,16 @@ def _generate_dstat_jobs(provider, query, create_time_max=None,
     if not create_time_max and query.end:
         create_time_max = query.end
 
+    # If the query explicitly requests submitted jobs which aren't running
```
calbach: opt: I'm not fully sure  what the impact of this would be on the rest of the code, but an alternative approach to this might be to switch this generator to yielding converted common Jobs API jobs; then after converting in this function you could do a simpler check

```
if query.statuses and api_job.status not in query.statuses:
  continue
```

If it would cause issues with the job ID token creation, for example, then I wouldn't bother.

bfcrampton: Good suggestion, ended up taking some minor refactoring but it ends up making things much cleaner working with a job model rather than a dict.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -257,13 +260,9 @@ def _provider_type():
...
-    return QueryJobsResult(
-        id=job_ids.dsub_to_api(project_id, job['job-id'], job.get(
-            'task-id', '')),
-        name=job['job-name'],
-        status=job_statuses.dsub_to_api(job),
-        submission=job['create-time'],
-        start=job.get('start-time'),
-        end=job['end-time'],
-        labels=labels.dsub_to_api(job))
+def _query_jobs_response(jobs,
```
calbach: opt_nit: This helper no longer has much value - I'd probably just inline it.

bfcrampton: Good idea, fixed.

PR #195, opened 2018-01-29
Datepicker for start and end time querying

alahwa:


ui/src/app/shared/common.ts
```diff
@@ -10,15 +10,19 @@ export enum JobStatusImage {
   Aborted = <any> 'https://www.gstatic.com/images/icons/material/system/1x/report_problem_grey600_24dp.png'
 }
 
+export const start = 'start';
```
calbach: nit: I might call these `startCol`, `endCol`

ui/src/app/shared/header/header.component.html
```diff
@@ -10,14 +10,26 @@
              src="https://www.gstatic.com/images/icons/material/system/1x/cancel_grey600_18dp.png">
       </mat-chip>
       <mat-menu #menu="matMenu" [overlapTrigger]="false" class="wide-menu">
-        <mat-form-field class="query-input">
+        <mat-form-field *ngIf="isFreeText()" class="query-input">
```
calbach: opt: `*ngSwitch` might be nice here and would ensure mutual exclusion for these being displayed

ui/src/app/shared/header/header.component.spec.ts
```diff
@@ -1,15 +1,23 @@
...
 } from "@angular/material";
 import {FormsModule, ReactiveFormsModule} from "@angular/forms";
 import {RouterTestingModule} from "@angular/router/testing";
 import {BrowserAnimationsModule} from "@angular/platform-browser/animations";
 
+import {HeaderComponent} from "./header.component";
+import {start} from "../common";
+
 
 describe('HeaderComponent', () => {
```
calbach: Unit test for date picking?

ui/src/app/shared/common.ts
alahwa: Done

ui/src/app/shared/header/header.component.spec.ts
alahwa: Added

PR #198, opened 2018-02-02
Show Cromwell labels in list view

rexwangcc:
In this PR:

1. Make changes to `jobs_controller` so it can query and load labels from Cromwell.
2. Add `.DS_Store` to gitignore


ui/README.md
```diff
@@ -1,5 +1,5 @@
 # Job Manager UI
-This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 2.2.3.
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 1.6.5.
```
rexwangcc: Not sure if this change is correct, but since the latest release of angular-cli is v1.6.7, I guess 2.2.3 should be a typo?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -267,3 +269,7 @@ def _get_base_url():
 def _get_user_auth():
     return HTTPBasicAuth(current_app.config['cromwell_user'],
                          current_app.config['cromwell_password'])
+
+
+def _is_parent_workflow(job):
+    return not job.get('parentWorkflowId')
```
rexwangcc: Extract the logic into a function so it could be more clear.

.gitignore
```diff
@@ -45,3 +45,6 @@ venv/
 # e2e
 /ui/e2e/*.js
 /ui/e2e/*.map
+
+# OS generated files
+.DS_Store
```
bfcrampton: tiny nit: newline so we don't see the ugly :no_entry_sign:

ui/README.md
bfcrampton: Yup good catch, I believe this was a mix up with the swagger-cli which is version 2.2.3

.gitignore
rexwangcc: Thanks, will do! Don't like the 🚫  either :(

ui/src/environments/environment.cromwell.staging.ts
```diff
@@ -4,7 +4,7 @@ export const environment = {
   production: false,
   requiresAuth: false,
   scope: '',
-  additionalColumns: null,
+  additionalColumns: ['cromwell-workflow-name', 'cromwell-workflow-id', 'other-labels'],
```
samanehsan: What would go in the `other-labels` column?

samanehsan:
Please bring this up to date with the master branch!

ui/src/environments/environment.cromwell.staging.ts
rexwangcc: The `other-labels` for now is just a literal placeholder, we have a ticket in backlog about determining what kind of labels we want to add to workflows, such as Sample id, Workflow type, Workflow version (e.g. v1.2.1), HCA bundle uuid and so on. I'm glad to remove it if that is confusing!

PR #200, opened 2018-02-05
List Cromwell jobs from newest to oldest 

samanehsan:
The Cromwell `query` endpoint returns jobs from oldest to newest, but we want to display jobs from newest to oldest in the UI. 

This PR updates the `query` endpoint in the translation layer to: 
1. Make a request to the query endpoint to get the `totalResultsCount`. 
2. Calculate which page to request from the API based on the total number of pages and the current page in the UI, so that the results are retrieved from newest to oldest.
3. Query the Cromwell API (filtering out subworkflows) and continue getting results if the initial amount returned is less than `query.page_size`. 

Notes:
- This request can take 30+ seconds if there are a lot of subworkflows in the page of results that is requested, so this PR also increases the `timeout` in gunicorn to 60 seconds (the default is 30s)
- Including the ability to filter out subworkflows in Cromwell would allow us to make fewer requests to get the number of jobs we want to display. Here's the open issue in the Cromwell repo: https://github.com/broadinstitute/cromwell/issues/3240

**This requires first updating the staging Cromwell to the latest hotfix that includes the `totalResultsCount` key.**

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -161,7 +161,8 @@ def update_key_names(metadata):
 
 def query_jobs(body):
     """
-    Query jobs by various filter criteria. Returned jobs are ordered from newest to oldest submission time.
+    Query jobs by various filter criteria. Additional jobs are requested if the number of results is less than the
+    requested page size. The returned jobs are ordered from newest to oldest submission time.
 
     :param body:
     :type body: dict | bytes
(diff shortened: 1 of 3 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -185,19 +220,13 @@ def query_jobs(body):
...
 
-    # Only list parent jobs
-    now = datetime.utcnow()
-    results = [
-        format_job(job, now) for job in response.json()['results']
-        if _is_parent_workflow(job)
-    ]
-    # Reverse so that newest jobs are listed first
-    results.reverse()
+    return response.json()['totalResultsCount']
```
rexwangcc: Since we can get `totalResultsCount` now, do you think we could replace the "many" on the page?
screen shot 2018-02-06 at 4 17 25 pm

bfcrampton: This is still not possible for dsub, however it could potentially be somehow encoded in the `/capabilities` end point in the future.

rexwangcc: I see, thanks for the explanation here! @bfcrampton

samanehsan:
I just updated the version of our staging cromwell and am waiting for the HCA integration test to pass before merging this in.

PR #204, opened 2018-02-09
Move paginator to the header

calbach:
- Refactor paginator from the table component, to the header; required restructuring the data flow between the job stream, data source, pagination control, table updates
- The job table no longer gets momentarily cleared out on navigation
- Refactor table to use the CDK model selection, including indeterminate state, e.g. if we have a mix of checked/unchecked -> select all now only selects the currently visible jobs
- Only show the aborted button if at least one selected job can be aborted
- Dropped the status tabs from the job details page (need to fully address the filter bar being there, but this was an easy incremental change to make in this PR)
- Move @types/angular to dev deps; it's odd that we need it at all since it just references Angular 1 stuff

### DIFF
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -207,7 +207,7 @@ describe('JobListComponent', () => {
   }));
 
   @Component({
-    selector: 'app',
+    selector: 'jm-test-app',
     template: '<router-outlet></router-outlet>'
   })
   class AppComponent {}
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -36,7 +36,7 @@ describe('JobListComponent', () => {
       status: JobStatus.Running,
       submission: new Date('2015-04-20T20:00:00')
     };
-    return (new Array(5)).map((_, i) => {
+    return (new Array(5)).fill(null).map((_, i) => {
       return {
         ...base,
         id: `JOB${i}`
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -124,7 +124,7 @@ describe('JobListComponent', () => {
     expectJobsRendered(fakeJobService.jobs.slice(0, 3));
   }));
 
-  it('paginates properly', fakeAsync(() => {
+  it('paginates forward', fakeAsync(() => {
     fakeJobService.jobs = testJobs(5);
     tick();
     fixture.detectChanges();
@@ -134,6 +134,15 @@ describe('JobListComponent', () => {
 
     // Page 2.
     expectJobsRendered(fakeJobService.jobs.slice(3, 5));
+  }));
+
+  it('paginates backwards', fakeAsync(() => {
+    fakeJobService.jobs = testJobs(5);
+    tick();
+    fixture.detectChanges();
+    const de: DebugElement = fixture.debugElement;
+    de.query(By.css('.mat-paginator-increment')).nativeElement.click();
+    fixture.detectChanges();
 
     // Back to page 1, which should have the first 3 jobs.
     de.query(By.css('.mat-paginator-decrement')).nativeElement.click();
@@ -162,16 +171,21 @@ describe('JobListComponent', () => {
     tick();
     fixture.detectChanges();
 
+    // Filter by active to filter out jobs after they've been aborted.
     let de: DebugElement = fixture.debugElement;
     de.query(By.css('.active-button')).nativeElement.click();
     tick();
     fixture.detectChanges();
 
+    // We select the first page (3 jobs) and abort. Jobs 3 and 4 are unaffected.
     de.query(By.css('jm-job-list-table')).componentInstance.toggleSelectAll();
     fixture.detectChanges();
     de.query(By.css('.group-abort')).nativeElement.click();
     tick();
     fixture.detectChanges();
+
+    // Jobs 3 and 4 should be the only jobs displayed, as they are the only
+    // remaining active jobs.
     expectJobsRendered(jobs.slice(3, 5));
   }));
 
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -19,10 +19,10 @@ import {RouterTestingModule} from '@angular/router/testing';
 import {JobListComponent} from "./job-list.component"
 import {JobsTableComponent} from "./table/table.component"
 import {JobManagerService} from '../core/job-manager.service';
+import {JobListResolver} from './job-list-resolver.service';
 import {FakeJobManagerService} from '../testing/fake-job-manager.service';
 import {SharedModule} from '../shared/shared.module';
-import {JobStream} from "../shared/job-stream";
-import {ActivatedRoute} from "@angular/router";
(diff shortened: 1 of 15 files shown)


ui/src/app/shared/header/header.component.css
```diff
@@ -20,3 +20,9 @@
   width: 7rem;
   margin-right: .5rem;
 }
+
```
alahwa: Is it simple to fix the height of the table and give it an inline scroll? I recall seeing examples of material tables doing that. If it's not a simple style thing, fine to leave this as a TODO for another ticket.

ui/src/app/shared/header/header.component.html
```diff
@@ -67,9 +67,18 @@
...
-    <button mat-raised-button class="status-button" (click)="showFailedJobs()">FAILED</button>
-    <button mat-raised-button class="status-button" (click)="showCompletedJobs()">COMPLETED</button>
+  <div class="search-table-controls">
+    <div *ngIf="shouldDisplayStatusButtons()">
+      <button mat-raised-button class="status-button" (click)="showActiveJobs()">ACTIVE</button>
+      <button mat-raised-button class="status-button" (click)="showFailedJobs()">FAILED</button>
+      <button mat-raised-button class="status-button" (click)="showCompletedJobs()">COMPLETED</button>
+    </div>
+
+    <mat-paginator *ngIf="jobs" #paginator
```
alahwa: Is it possible to get this paginator to right-align regardless of whether the status buttons are present or not? The current behavior is a little jarring.

ui/src/app/shared/header/header.component.spec.ts
```diff
@@ -12,17 +12,33 @@ import {
...
+import {BehaviorSubject} from 'rxjs/BehaviorSubject';
 
+import {JobListView} from "../job-stream";
 import {HeaderComponent} from "./header.component";
 import {startCol, statusesCol} from "../common";
+import {QueryJobsResult} from '../model/QueryJobsResult';
 import {JobStatus} from "../model/JobStatus";
 
 
 describe('HeaderComponent', () => {
```
alahwa: + a test for should show/hide the paginator depending on jobs present

ui/src/app/shared/header/header.component.ts
```diff
@@ -174,7 +228,8 @@ export class HeaderComponent implements OnInit {
   }
 
   shouldDisplayStatusButtons(): boolean {
-    return !URLSearchParamsUtils.unpackURLSearchParams(
+    // jobs is only populated on multi-job views
+    return this.jobs && !URLSearchParamsUtils.unpackURLSearchParams(
```
alahwa: Predicating whether to show UI elements (both here and with the paginator) on the nullness of the jobs input is somewhat unintuitive and sort of breaks the component modularity agreement. Even though it might be redundant, consider providing a separate bool showControls input and then setting it to true for JobsList and false for JobDetails. That make this behavior a lot more transparent and increases the useability of this module.

ui/src/app/shared/header/header.component.css
calbach: Looked into this a bit, but it didn't look like a trivial change so I'm going to punt for now as the PR is already large/complex.

ui/src/app/shared/header/header.component.spec.ts
calbach: Done, along with other tests.

ui/src/app/job-list/table/table.component.html
```diff
@@ -3,12 +3,18 @@
   <!-- Job name column -->
   <ng-container matColumnDef="Job">
     <mat-header-cell *matHeaderCellDef>
-      <mat-checkbox class = "checkbox" [checked]="allSelected" (change)="toggleSelectAll()">
+      <mat-checkbox class = "checkbox"
+                    (change)="$event ? toggleSelectAll() : null"
```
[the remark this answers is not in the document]
alahwa: Style guide reference: https://angular.io/guide/styleguide#put-presentation-logic-in-the-component-class

I'll leave it to your judgment, it's a pretty simple operation so I don't have an overwhelming preference either way.

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -97,59 +103,77 @@ describe('JobsTableComponent', () => {
...
     expect(dsubColumns.length).toEqual(2);
     expect(dsubColumns[0].nativeElement.textContent)
-      .toEqual(testJob1.labels['user-id']);
+      .toEqual(jobs[0].labels['user-id']);
     expect(dsubColumns[1].nativeElement.textContent)
-      .toEqual(testJob1.labels['status-detail']);
+      .toEqual(jobs[0].labels['status-detail']);
   }));
 
   it('displays the abort button for selected active jobs', async(() => {
```
alahwa: Regardless of whether you decide to split out the test I mentioned in job-list-spec above, here I think it does make sense to have these be 4 separate unit tests.

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -61,42 +72,135 @@ describe('JobListComponent', () => {
...
+    de.query(By.css('.active-button')).nativeElement.click();
+    tick();
+    fixture.detectChanges();
+
+    de.query(By.css('jm-job-list-table')).componentInstance.toggleSelectAll();
+    fixture.detectChanges();
+    de.query(By.css('.group-abort')).nativeElement.click();
+    tick();
+    fixture.detectChanges();
+    expectJobsRendered(jobs.slice(3, 5));
```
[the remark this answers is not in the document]
calbach: Added comments, let me know if this is still unclear.

ui/src/app/job-list/table/table.component.html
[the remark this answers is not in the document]
calbach: Thanks for the link. I don't think it should be a hard and fast rule, but I've applied it in the checked/indeterminate cases, so that I can at least comment the logic there.

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -61,42 +72,135 @@ describe('JobListComponent', () => {
...
+    de.query(By.css('.mat-paginator-increment')).nativeElement.click();
+    fixture.detectChanges();
+
+    // Page 2.
+    expectJobsRendered(fakeJobService.jobs.slice(3, 5));
+
+    // Back to page 1, which should have the first 3 jobs.
+    de.query(By.css('.mat-paginator-decrement')).nativeElement.click();
+    fixture.detectChanges();
+    expectJobsRendered(fakeJobService.jobs.slice(0, 3));
```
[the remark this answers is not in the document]
calbach: Fair point, I think it's a reasonable guideline to follow. Definitely agree I should make these test cases more targeted. In general a lot of this is me defensively making assertions due to lack of familiarity with what Angular is going to do or not do when I call `fixture.detectChanges()`. The "assert" phase here was really just the last line.

Anyways, split this out. Slightly concerned that there's a chance that "paginates backwards" could still pass without having paginating forwards or backwards. But I suppose the two tests in combination should cover this.

ui/src/app/job-list/table/table.component.spec.ts
calbach: Done

calbach:
@alahwa https://github.com/DataBiosphere/job-manager/pull/205 is included in this PR. Do you still want to review that separately, or should I drop that?

PR #205, opened 2018-02-12
replace unused mock service with fake

calbach:


calbach:
Circle failure is a flake on a dsub integration test

PR #206, opened 2018-02-13
[API] Add capabilities endpoint

bfcrampton:
Initial API changes outlined in this [design doc](https://docs.google.com/document/d/1F6y-5HOw7wXsEMYAsOrpeOBcj4DnLqqQx72UOpZfRDM/edit?ts=5a7894a3#). Will follow up in a separate PR with changes to `JobMetadataResponse` and `QueryJobsResult` models, along with implementation changes.


api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
...
 definitions:
+  AuthenticationCapability:
+    description: Specification of the backend's authentication capabilities
+    required:
+      - required
+    properties:
+      required:
+        type: boolean
+        description: Whether or not authentication is required
+      client_id:
```
calbach: Per doc -> mv this one back to environment.ts

api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
       tags:
         - Jobs
 definitions:
+  AuthenticationCapability:
+    description: Specification of the backend's authentication capabilities
+    required:
+      - required
```
calbach: nit: I think I'd prefer to avoid required here and for the authentication tuple, i.e. if you don't care about authentication, you don't need to return anything.

api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
...
+  CapabilitiesResponse:
+    description: Specification of the backend's capabilities
+    required:
+      - authentication
+    properties:
+      authentication:
+        $ref: '#/definitions/AuthenticationCapability'
+      display_fields:
+        type: object
+        description: Fields on QueryJobsResult returned from POST /jobs/query
```
calbach: nit: `|` or `>`?

api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
...
+          populated on some or all jobs. The fields are mapped to their display
+          names, in order of importance. The UI will attempt to display as
```
calbach: For separation of concerns, I'd drop this UI sentence. It's up to the UI to decide what to do with this information.

api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
...
+          labels can be included, such as 'label.foo' or
+          'extended_fields.user_id'
+      label_query_fields:
+        type: object
+        description: Labels on QueryJobsRequest which are queryable, mapped to
+          their display names.
+      extended_query_fields:
+        type: object
+        description: Fields on ExtendedQueryFields which are queryable, mapped
+          to their display names.
```
calbach: Actually, I don't think there's a reason to encode display names here. These fields are all hardcoded and have existing definitions; these shouldn't vary across backends. I'd just make this an array.

api/jobs.yaml
```diff
@@ -116,6 +131,45 @@ paths:
       tags:
         - Jobs
 definitions:
+  AuthenticationCapability:
+    description: Specification of the backend's authentication capabilities
+    required:
+      - required
+    properties:
+      required:
```
samanehsan: nit: What do you think about naming this `is_required` since it's a boolean?

api/jobs.yaml
bfcrampton: Good point, done.

api/jobs.yaml
bfcrampton: I think that makes sense, switched.

api/jobs.yaml
bfcrampton: Done.

api/jobs.yaml
bfcrampton: Done.

api/jobs.yaml
bfcrampton: Done.

api/jobs.yaml
bfcrampton: Yeah I've been going back and forth on that, think I agree its better to use their hardcoded names. I made the change for label_query_fields too, let me know it you think there's a reason we should have those mapped still.

api/jobs.yaml
```diff
@@ -116,6 +131,40 @@ paths:
...
+        $ref: '#/definitions/AuthenticationCapability'
+      display_fields:
+        type: object
+        description: >
+          Fields on QueryJobsResult returned from POST /jobs/query
+          populated on some or all jobs. The fields are mapped to their 
+          display names, in order of importance. Extended fields and
+          labels can be included, such as 'label.foo' or
+          'extended_fields.user_id'
+      label_query_fields:
```
calbach: Per comment I just added to the design doc, I think this naming should be generalized and description updated.

bfcrampton: Done.

api/jobs.yaml
```diff
@@ -116,6 +131,40 @@ paths:
       tags:
         - Jobs
 definitions:
+  AuthenticationCapability:
+    description: Specification of the backend's authentication capabilities
+    properties:
+      is_required:
```
calbach: Actually, we do use camelCase elsewhere, should use that here as well.

e.g. parentId

bfcrampton: Good point, didn't notice that. Switched all snake_case to camelCase

api/jobs.yaml
```diff
@@ -116,6 +131,40 @@ paths:
...
+          labels can be included, such as 'label.foo' or
+          'extendedFields.userId'
+      commonLabels:
+        type: array
+        description: Common labels which are present on most jobs returned
+        items:
+          type: string
+      extendedQueryFields:
+        type: array
+        description: Fields on ExtendedQueryFields which are queryable
```
samanehsan: Should this say "Fields on QueryJobsResult which are queryable"?

bfcrampton: This really only applies to fields on the [ExtendedQueryFields](https://github.com/DataBiosphere/job-manager/pull/208/files#diff-4945231733bba41c658978333f49c3dfR273) model. It is a bit confusing submitting this before that model has been added but I think it should be fine in the short term.

samanehsan: Oh, I see! Thanks for the clarification

PR #208, opened 2018-02-15
[API] Update job models with extended fields

bfcrampton:
I'm going to use this as the base PR for making the rest of the changes in the [Job Manager Capabilities Doc](https://docs.google.com/document/d/1F6y-5HOw7wXsEMYAsOrpeOBcj4DnLqqQx72UOpZfRDM/edit?ts=5a7894a3#). 

After this is API change is approved I will open separate PRs with the necessary implementation changes to both backends and the UI with this branch as the base.

- [x] dsub server changes (#211)
- [x] cromwell server changes (#215)
- [x] UI changes (#222)


api/jobs.yaml
```diff
@@ -201,85 +201,12 @@ definitions:
...
-      - failure
-    properties:
-      failure:
-        type: string
-        description: The failure message
-      timestamp:
-        type: string
-        format: date-time
-        description: The time at which this failure occurred
+      extendedFields:
```
calbach: opt: I'm on the fence, but looking at this again, "fields" feels slightly redundant. We could consider making this `extended` instead or `extensions` for more readable references:

`job.extendedFields.userId`
`job.extended.userId`
`job.extensions.userId`

Both here and for query (though the model name could remain as-is). WDYT?

api/jobs.yaml
```diff
@@ -341,15 +268,26 @@ definitions:
...
+      projectId:
         type: string
         description: >
-          Returns only jobs with the given parent ID. This parameter may be
-          unsupported for some API implementations or required for others
-          depending on whether there exists a logical noun above a job
-          in the resource hierarchy (for example, a cloud project). If the
-          presence of this parameter is incompatible with the server, it may
-          return a 400 HTTP status.
+          Returns only jobs with the specified project ID. If specified by the
```
calbach: nit: I'd probably say "belonging to the specified  project". And at the end "; all jobs have exactly one parent project id."

bfcrampton:
Tests are going to fail until I update the broken uses of these models.

PR #209, opened 2018-02-16
[dsub] Implement capabilities endpoint

bfcrampton:


servers/dsub/jobs/controllers/capabilities_controller.py
```diff
@@ -1,7 +1,43 @@
...
+                'https://www.googleapis.com/auth/genomics',
+                'https://www.googleapis.com/auth/cloudplatformprojects.readonly'
+            ]),
+        display_fields={
+            'name': 'Job',
+            'status': 'Status',
+            'submission': 'Submitted',
+            'labels.job_id': 'Job ID',
+            'labels.task_id': 'Task ID',
+            'extensions.user_id': 'User ID',
```
calbach: This should be camelCased, I think. Here and throughout

servers/dsub/jobs/controllers/capabilities_controller.py
```diff
@@ -1,7 +1,43 @@
...
+            'extensions.status_detail': 'Status Detail'
+        },
+        common_labels=['job_id', 'task_id'],
+        extended_query_fields=['project_id', 'user_id', 'submission'])
+
+    if _provider_type() == ProviderType.GOOGLE:
+        capabilities.authentication = AuthenticationCapability(
```
calbach: Did you mean to leave this unset above?

calbach: Ah.. beat me to it.

bfcrampton: Yup sorry ... noticed that right as I sent it for review

servers/dsub/jobs/controllers/capabilities_controller.py
bfcrampton: That's a good point since the typescript models are camelCase. Updated, I guess we leave labels snake_case right?

calbach: Actually, they should be hyphenated. It won't be by style convention here - it will need to exactly match the casing of the actual label keys we expect to be returning.

bfcrampton: Right, fixed.

servers/dsub/jobs/controllers/capabilities_controller.py
```diff
@@ -1,7 +1,37 @@
...
     Returns:
         CapabilitiesResponse: Response containing this backend's capabilities
     """
-    raise NotImplemented('Not yet supported.')
+    capabilities = CapabilitiesResponse(
+        display_fields={
+            'name': 'Job',
+            'status': 'Status',
+            'submission': 'Submitted',
+            'labels.job_id': 'Job ID',
```
calbach: hyphen here and below

bfcrampton: Whoops, thanks

PR #210, opened 2018-02-16
[cromwell] Implement capabilities endpoint

bfcrampton:


servers/cromwell/jobs/controllers/capabilities_controller.py
```diff
@@ -1,7 +1,20 @@
...
+    return CapabilitiesResponse(
+        display_fields={
+            'name': 'Job',
+            'status': 'Status',
+            'submission': 'Submitted',
+            'labels.cromwell-workflow-name': "Workflow Name",
+            'labels.cromwell-workflow-id': "Workflow ID",
+            'labels.comment': "Comment"
+        },
+        common_labels=[],
```
samanehsan: This list should include `cromwell-workflow-name`, `cromwell-workflow-id` and `comment`. I think anything that we're displaying as a column should be included in the drop-down that lists queryable fields.

samanehsan: Also we expect to have different labels for different job manager deployments, so we'll need to make these configurable. But that can definitely happen in a separate PR.

bfcrampton: Good point, didn't realize those can be queried by, added!

Yup definitely will need to be configured by deployment. This was just to bring it up to parody with our current environment files.

I was thinking an easy way to configure this would be too add a folder of deployment configurations with JSON files and then this method can just read a file based on an ENV variable or something. In any case I'll leave that up to you guys when its needed.

PR #211, opened 2018-02-16
[dsub] Implement changes to returned job models for capabilities update

bfcrampton:


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -245,17 +233,33 @@ def _handle_http_error(error, proj_id):
...
 
 
 def _provider_type():
     return current_app.config['PROVIDER_TYPE']
+
+
+def _query_jobs_result(job, project_id=None):
+    return QueryJobsResult(
+        id=job_ids.dsub_to_api(project_id, job['job-id'], job.get(
+            'task-id', '')),
```
calbach: nit: I'd probably make it accept `None` for task-id

servers/dsub/jobs/controllers/utils/extensions.py
```diff
@@ -0,0 +1,19 @@
...
+
+        Args:
+            job: A dict with dsub job metadata
+
+        Returns:
+            ExtendedFields: Populated extensions on the job
+    """
+    return ExtendedFields(
+        user_id=job['user-id'],
+        status_detail=job['status-detail'],
```
calbach: Are `status-detail` and `last-update` sure things? Will they ever be `None`? Maybe safer to just use `get()` instead

servers/dsub/jobs/controllers/utils/labels.py
```diff
@@ -10,15 +10,8 @@ def dsub_to_api(job):
     # Put any dsub specific information into the labels. These fields are
```
calbach: nit: drop comment

servers/dsub/jobs/controllers/utils/labels.py
```diff
@@ -1,4 +1,4 @@
-def dsub_to_api(job):
+def get_labels(job):
```
calbach: nit: I like the previous naming better I think, it's basically just stripping out things that are not labels in the API.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Yup, already does not sure why it was like this.

servers/dsub/jobs/controllers/utils/extensions.py
bfcrampton: Good point, changed to `get()`

servers/dsub/jobs/controllers/utils/labels.py
bfcrampton: Done.

servers/dsub/jobs/controllers/utils/labels.py
bfcrampton: Done.

PR #212, opened 2018-02-20
[CI] Ensure building for production succeeds

bfcrampton:


calbach:
hm, not the expected failure message. 

```
#!/bin/bash -eo pipefail
cd ui
ng build --target=production --environment=dev
/bin/bash: line 1: ng: command not found
Exited with code 127
```

PR #213, opened 2018-02-20
Fix ng build visibility issues

calbach:


ui/src/app/job-list/job-list.component.html
```diff
@@ -4,5 +4,5 @@
   [pageSize]="pageSize"></jm-header>
 <jm-job-list-table
   [dataSource]="dataSource"
-  (onJobsChanged)="handleJobsChanged(jobs)">
+  (onJobsChanged)="handleJobsChanged($event)">
```
alahwa: Doesn't handleJobsChanged still accept a QueryJobsResult[]? This change seems unnecessary given the one below

calbach: Correct. The emitter just sends the subset of jobs that were modified (not all jobs). I gather that this parameter needs to be specially called `$event`.

That said, the job list component doesn't actually even look at this information currently. Just seemed like a reasonable thing to emit.

alahwa: In that case, would it be better to just remove that parameter entirely?

calbach: Yeah, good point. I've left the emitter value, but dropped this callback configuration.

PR #214, opened 2018-02-20
UI updates for the table

alahwa:
- Automatically launch search on completed chips
- Fix a bug where searching with an empty status chip would throw a 500 error
- Change the mat-button job details link to clickable text (this fix caused style alignment issues which was best resolved by moving the details dropdown to a separate column, which probably makes the most sense anyway)


ui/src/app/shared/header/header.component.ts
```diff
@@ -115,13 +115,16 @@ export class HeaderComponent implements OnInit, AfterViewInit, OnDestroy {
...
+        this.search();
       }
       else {
         // Parse as just the key
         this.deleteChipIfExists(value);
         this.chips.set(value.trim(), '');
       }
       this.inputValue = "";
+    } else {
+      this.search();
```
bfcrampton: Why is this needed?

ui/src/app/shared/header/header.component.ts
```diff
@@ -115,13 +115,16 @@ export class HeaderComponent implements OnInit, AfterViewInit, OnDestroy {
         let keyVal: string[] = value.split(':');
         this.deleteChipIfExists(keyVal[0].trim());
         this.chips.set(keyVal[0].trim(), keyVal[1].trim());
+        this.search();
       }
       else {
         // Parse as just the key
         this.deleteChipIfExists(value);
         this.chips.set(value.trim(), '');
```
bfcrampton: Would it be easy to trigger showing the equals/date drop down menu from this? It still feels odd to add a blank chip

ui/src/app/shared/header/header.component.ts
alahwa: Not needed

ui/src/app/shared/header/header.component.ts
alahwa: Agreed that this should be the behavior, and I plan that this will be a part of the effort to open/close menus and auto focus where needed. Still figuring out the best way to do this, since it will likely involve tightly coupling the DOM to the app logic which generally degrades code quality.

PR #217, opened 2018-02-21
[API] Update CapabilitiesResponse.displayFields to be an array

bfcrampton:
This is needed in order to maintain priority ordering of the fields. An `object` (or Map) does not have  guaranteed ordering.


api/jobs.yaml
```diff
@@ -165,6 +167,11 @@ definitions:
         description: Fields on ExtendedQueryFields which are queryable
         items:
           type: string
+  DisplayField:
+    description: Map of field name to display name
```
calbach: ultranit: I'd say description of a display field; could extend this with more later potentially

api/jobs.yaml
```diff
@@ -201,85 +208,12 @@ definitions:
       labels:
         type: object
         description: Custom job labels with string values
```
calbach: How'd all this other stuff move?

api/jobs.yaml
bfcrampton: Done.

api/jobs.yaml
bfcrampton: Oops I copied some stuff from my PR with job models updates. Removed.

PR #218, opened 2018-02-22
Log user email on dsub interactions

calbach:


calbach:
Anything logged on the UI is completely lost in the void currently. I considered trying to put this only on a more singular "entrypoint" request, but there's no such endpoint and this might miss certain usage patterns.

I think the overhead here should be pretty negligible, but if it becomes a problem I can dig. I think lets get something in ASAP so we can start monitoring usage.

bfcrampton:
SGTM

PR #220, opened 2018-02-22
Remove job list header on details page

dmohs:
This is my proposed solution to https://elastc.com/c/4pbVV8SQ/216-filter-bar-appears-on-job-details-page.

Looking at the header component, my judgement is that it isn't so much a general header as a job list header, and so does not belong on the job details page. If that is true, I'd suggest a follow-up PR to rename as such. I think it would be appropriate to create a separate header for the job details page if/when that seems like a good idea (e.g., to add a link back to the jobs list).

### DIFF
--- a/ui/src/app/job-details/job-details.component.html
+++ b/ui/src/app/job-details/job-details.component.html
@@ -1,4 +1,3 @@
-<jm-header [showControls]="false"></jm-header>
 <jm-panels [job]="job"></jm-panels>
 <!-- Only show the tasks table if there is at least 1 task. -->
 <jm-tasks *ngIf="hasTasks()" [tasks]="job.tasks" [jobId]="job.id"></jm-tasks>


calbach:
Agree with your judgement in the PR description - composing the query builder into different headers (if needed) would be cleaner.

RE: actual proposed change, I'm a bit on the fence. I'm +.5 on dropping the query builder from the details view, but -1 from losing our only breadcrumb back to the query page. If we do this I think ideally we'd replace this with a back button. If others are currently complaining about the pop-up query builder issue, then I'm +1 to going ahead with this as-is (for a net of +.5); were you able to reproduce this?

For more context, the current guidance from Marty w.r.t. the query builder being shown on the details page is as follows:

  - On the details page, add another chip e.g. job-id which scopes it to a particular job; the query builder is now in a reasonable looking state on this page (albeit slightly confusing, since had you formed that query on your own - we'd need to decide which page you're going to land on and what the logic would be)
  - If the above introduces too many issues, remove the query builder from the details page entirely

kbergin:
There's also a ticket for Marty to design what we might want for breadcrumbs. https://elastc.com/c/pYeTXDi7/205-breadcrumbs-and-location so we could put in something to hold us over?

calbach:
@bfcrampton and I chatted yesterday and I convinced myself that I'm in favor of going with the "id chip" approach for now. I think it is actually nice to have the search bar as a persistent navigation element, and I think it makes the search/click in cycle more seamless. This was also Marty's top preference.

Adding something like that is probably out of scope for this PR, and I will plan to take that on once I'm able.

But tl;dr for this PR is that I think we want to keep a header component on the details page for now.

dmohs:
@calbach I'm going to close this PR because I don't think it is the right solution, but I do feel this component needs some cleanup. First, the phrase "Filter jobs..." suggests a list that I can whittle down, which is confusing when I am looking at the job details. Second, the main way to go back (as far as I can tell) is to click the magnifying glass icon, which opens the filter bar before navigating. That's a bit confusing. I think that is what lead me to believe this component wasn't meant to be displayed on this page, which I now know is incorrect.

PR #221, opened 2018-02-22
Fix initial non-exhaustive paginator rendering

calbach:
The bug was that if you had a long list of jobs, the paginator would show "1-50 of 100" rather than "1-50 of many", until you interacted with the component, e.g. hover.

The cause of this issue is that `afterViewInit` gets run after the initial rendering. Modifying `_intl` also doesn't directly mutate anything that is rendered in the template, so ng doesn't pick up the update. Explicitly notifying the paginator of the change on initialization fixes this.

### DIFF
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -49,7 +49,7 @@ describe('JobListComponent', () => {
   let fixture: ComponentFixture<TestJobListComponent>;
   let fakeJobService: FakeJobManagerService;
 
-  beforeEach(async(() => {
+  beforeEach(fakeAsync(() => {
     fakeJobService = new FakeJobManagerService(testJobs(5));
 
     TestBed.configureTestingModule({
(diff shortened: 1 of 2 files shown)


calbach:
That didn't work, apparently it is not OK to update things during this phase of view initialization. Working on revising the approach..

calbach:
SHould now be working, with some unfortunate extra ticks...

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -147,6 +156,7 @@ describe('JobListComponent', () => {
     // Back to page 1, which should have the first 3 jobs.
     de.query(By.css('.mat-paginator-decrement')).nativeElement.click();
     fixture.detectChanges();
+    tick();
```
bfcrampton: Just a note: I've found that have a `tick()` followed by `fixture.detectChanges()` works best for me (sometimes doesn't work when ordered the other way). Not sure if that was specific to my case or a general thing.

calbach: I think this is good for cases where it's:

- run some view initialization which triggers some async promises
- tick()
- render (promise values now available)

In my case, the problem was actually some random library that paginator depends on setting a timeout after it gets rendered, annoyingly.

PR #222, opened 2018-02-23
[UI] Implement changes for capabilities update and job model changes

bfcrampton:
Final implementation changes for Job Manager Capabilities. Once this PR is merged into #208 that should be ready for submission.


bfcrampton:
As a note: I decided that the "Jobs" column (job name) should always be displayed (and not specified in `capabilities.display_fields`. This makes a number of things simpler and I can't think of a real use case for not showing the job name. However, if others think all the columns should be configurable I can change it back.

ui/src/app/core/auth.service.ts
```diff
@@ -30,28 +31,30 @@ export class AuthService {
...
 
-  constructor(private zone: NgZone) {
-    if (environment.requiresAuth) {
-      this.initAuthPromise = new Promise<void>( (resolve, reject) => {
-        gapi.load('client:auth2', {
-          callback: () => this.initAuth().then(() => resolve()),
-          onerror: () => reject(),
+  constructor(private zone: NgZone, capabilitiesService: CapabilitiesService) {
+    capabilitiesService.getCapabilities().then(capabilities => {
+      if (capabilities.authentication && capabilities.authentication.isRequired) {
```
calbach: nit: I'd invert and return early to reduce indentation.

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    private readonly authService: AuthService,
+    private readonly capabilitiesService: CapabilitiesService,
+    private readonly router: Router) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
+    return this.capabilitiesService.getCapabilities()
+      .then(capabilities =>
+        this.handleAuthCapabilities(
+          capabilities, route.routeConfig.path, state.url))
+      .then(result => result[0] == false
```
calbach: `.then([authenticated, capabilities] => ...`

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,72 @@
...
+          ? Promise.resolve([false, result[1]])
```
calbach: SHould be equivalent to directly return result here

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,72 @@
...
+        return [true, capabilities];
+      }
+
+      return this.authService.isAuthenticated().then( (authenticated) => {
+        if (!authenticated) {
+          this.router.navigate(['sign_in'], {
+            queryParams: { returnUrl: url }
+          });
+        }
+        return [authenticated, capabilities] as [boolean, CapabilitiesResponse];
```
calbach: wdyt about erroring the promise instead in the unauthenticated case, e.g. by throwing an error here, or returning a `Promise.error('foo')`

ui/src/app/job-details/panels/panels.component.ts
```diff
@@ -27,28 +27,32 @@ export class JobPanelsComponent implements OnChanges {
...
-
   }
 
   getInputResourceURL(key: string): string {
     return ResourceUtils.getResourceBrowserURL(this.job.inputs[key]);
   }
 
   getLogResourceURL(key: string): string {
-    return ResourceUtils.getResourceURL(this.job.logs[key]);
+    return ResourceUtils.getResourceURL(this.job.extensions.logs[key]);
```
calbach: guard for null extensions?

ui/src/app/job-details/tasks/tasks.component.spec.ts
```diff
@@ -10,28 +10,37 @@ import {
...
-    attempts: 1,
-    failures: [],
-    returnCode: 0,
-    stderr: 'gs://test-bucket/stderr.txt',
-    stdout: 'gs://test-bucket/stdout.txt',
-    inputs: {},
-    jobId: 'subworkflow123'
-  };
-  let testTasks: TaskMetadata[] = [task];
+            name: 'task1',
```
calbach: nit: unindent

ui/src/app/job-list/job-list.module.ts
```diff
@@ -10,7 +10,7 @@ import {
   MatTableModule,
   MatTooltipModule,
 } from '@angular/material';
-import {NgModule} from '@angular/core';
+import {NgModule, LOCALE_ID} from '@angular/core';
```
calbach: unused?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -28,37 +33,40 @@ import {environment} from '../../../environments/environment';
...
   private mouseoverJob: QueryJobsResult;
 
-  public additionalColumns: string[] = [];
+  public displayFields: DisplayField[];
   public selection = new SelectionModel<QueryJobsResult>(/* allowMultiSelect */ true, []);
   public jobs: QueryJobsResult[] = [];
 
   // TODO(alanhwang): Allow these columns to be configured by the user
-  displayedColumns = primaryColumns.slice();
+  displayedColumns: string[] = ["Job", "Details"];
```
calbach: what is "Details"?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -84,6 +92,15 @@ export class JobsTableComponent implements OnInit {
       .catch((error) => this.handleError(error));
   }
 
+  displayStatusColumn(): boolean {
```
calbach: I don't see this used, is it?

ui/src/app/job-list/table/table.component.ts
calbach: Is this array still needed? Could displayFields subsume this?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -101,11 +118,24 @@ export class JobsTableComponent implements OnInit {
...
+  isStatusField(df: DisplayField): boolean {
+    return df.field == "status";
+  }
+
+  getFieldValue(job: QueryJobsResult, df: DisplayField): any {
+    // Handle nested fields, separated by '.', i.e. extensions.userId
+    let fields = df.field.split(".");
+    let value = job;
+    for (let field of fields) {
+      value = value[field];
```
calbach: Slightly worried about a tough-to-debug error here as a result of returning a malformed field path from the capabilities endpoint. Perhaps exit early if value is null.

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -11,23 +12,38 @@ export class URLSearchParamsUtils {
...
+
+    if (request.extensions) {
+      if (request.extensions.projectId) {
+        urlSearchParams.set('projectId', request.extensions.projectId);
+      }
+      if (request.extensions.userId) {
+        urlSearchParams.set('userId', request.extensions.userId);
+      }
+      if (request.extensions.submission) {
+        urlSearchParams.set('submission', request.extensions.submission.toISOString());
```
calbach: Did you meant o include the submission-related stuff in this PR? Is that supported on the server yet?

ui/src/app/core/capabilities-activator.service.ts
bfcrampton: Not supported in typscript unfortunately :(

ui/src/app/core/capabilities-activator.service.ts
bfcrampton: Done

ui/src/app/core/capabilities-activator.service.ts
bfcrampton: I'd probably prefer not too, why you think that'd be better than just returning false for the `canActivate` and navigating to the sign in page. IMO navigating to a page you are not yet authenticated for isn't really an error, it just triggers a different application flow to authenticate you before returning to the page you were on.

ui/src/app/job-details/panels/panels.component.ts
bfcrampton: Done.

ui/src/app/job-details/tasks/tasks.component.spec.ts
bfcrampton: Whoops, fixed.

ui/src/app/job-list/job-list.module.ts
bfcrampton: rm

ui/src/app/job-list/table/table.component.ts
bfcrampton: It's the Job details dropdown column [here](https://github.com/DataBiosphere/job-manager/pull/222/files/f3a063f766bd8dd9b94803e647f172242d0dbcb9#diff-906b98c1103b300e8cf6d2c5eb1a56edR24).

bfcrampton: This determines all the columns displayed on screen, including the default "Job" and details drop down columns. One issue is that displayFields is not always immediately available, so these are the default columns shown until the capabilities endpoint returns. We could probably  have a function which does this but it seems simpler to just keep track of it in a variable.

ui/src/app/job-list/table/table.component.ts
bfcrampton: Nope, sorry thought I removed it.

ui/src/app/job-list/table/table.component.ts
bfcrampton: Good point, done.

ui/src/app/shared/utils/url-search-params.utils.ts
bfcrampton: Yup, I did. I think the UI should support populating the request with query fields and query extension fields as soon as they are added to the model. Our dsub backend does not yet support submission, but our capaibilites endpoint also doesn't return it as a `query_extensions` so it won't be displayed (that actually wasn't true but I just changed it :+1:)

cromwell-compose-dev.yml
```diff
@@ -4,7 +4,7 @@ services:
     extends:
       file: common-compose.yml
       service: ui
-    command: ["npm", "run-script", "ng", "--", "serve", "--host", "ui", "--env=cromwell_dev"]
+    command: ["npm", "run-script", "ng", "--", "serve", "--host", "ui", "--env=cromwell_dev", "--aot"]
```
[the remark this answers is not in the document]
bfcrampton: I've found that it makes the development UI much faster with fewer webpack requests with minimal impact on compilation time (and still has hot-reloading). However, this is just based on qualitative experience though and it is out of scope of this PR so I'll remove it and may follow up separately.

bfcrampton:
Per offline discussion, updated `capabilities-activator.service` to reject promises when the route cannot be activated and handle those at the end. Ended up being much cleaner thanks @calbach.

ui/src/app/core/capabilities.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/** Service wrapper for accessing the capabilities endpoint. */
+@Injectable()
+export class CapabilitiesService {
+
+  private capabilitiesResponse: Promise<CapabilitiesResponse>;
+
+  constructor(private http: Http) {}
+
+  getCapabilities(): Promise<CapabilitiesResponse> {
```
calbach: Per offline discussion, I would also support a synchronous getter for this which can be used from components, to simplify the initialization life cycle (throw error if not populated).

ui/src/app/job-details/job-details.component.html
```diff
@@ -1,4 +1,4 @@
 <jm-header [showControls]="false"></jm-header>
 <jm-panels [job]="job"></jm-panels>
 <!-- Only show the tasks table if there is at least 1 task. -->
-<jm-tasks *ngIf="hasTasks()" [tasks]="job.tasks" [jobId]="job.id"></jm-tasks>
+<jm-tasks *ngIf="hasTasks()" [tasks]="job.extensions.tasks" [job]="job"></jm-tasks>
```
samanehsan: We could consider just passing in `job` here since the tasks are available on the job. But that doesn't need to happen in this PR since it's already such a large change!

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,77 @@
...
+export class CapabilitiesActivator implements CanActivate {
+
+  constructor(
+    private readonly authService: AuthService,
+    private readonly capabilitiesService: CapabilitiesService,
+    private readonly router: Router) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
+    return this.capabilitiesService.getCapabilities().then(cap => {
+      return this.handleAuthCapabilities(cap, route.routeConfig.path, state.url)
```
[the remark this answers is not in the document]
bfcrampton: Done

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,77 @@
...
+        .then(() => this.handleProjectCapabilities(cap, route))
+        .catch(error => {
+          // Handle all not-activated errors by just returning false for
+          // this promise. All others, re-throw them.
+          if (error == 'not-activated') {
+            return false;
+          }
+          Promise.reject(error);
```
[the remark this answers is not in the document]
bfcrampton: Done.

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,77 @@
...
+        });
+    });
+  }
+
+  private handleAuthCapabilities(capabilities: CapabilitiesResponse, path: String, url: String): Promise<void> {
+    if (capabilities.authentication && capabilities.authentication.isRequired) {
+      if (this.authService.authenticated.getValue() || path == 'sign_in') {
+        return Promise.resolve();
```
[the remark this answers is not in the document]
bfcrampton: To use chaining I need to return a `Promise<CapabilitiesResponse>` updated to `Promise.resolve(capabilities)`

ui/src/app/core/capabilities.service.ts
bfcrampton: Done.

ui/src/app/job-details/job-details.component.html
bfcrampton: Hmm that's a good point, I think that would work but mind if I leave that to another PR?

ui/src/app/shared/header/header.component.ts
```diff
@@ -209,7 +212,7 @@ export class HeaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
   removeChip(chipKey: string): void {
     this.chips.delete(chipKey);
-    if (chipKey == statusesCol) {
+    if (chipKey == "statuses") {
```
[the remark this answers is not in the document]
bfcrampton: Done.

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -95,11 +124,17 @@ export class URLSearchParamsUtils {
...
-    let fields: string[] = [];
-    for (let field in queryFields) {
-      fields.push(queryFields[field]);
+  public static getQueryFields(capabilities: CapabilitiesResponse): Map<string, FieldDataType>  {
+    // We could make commonLabels a map of label to FieldDataType to support
+    // non-text label types.
+    var labelFields = new Map<string, FieldDataType>();
+    if (capabilities.commonLabels) {
+      labelFields = new Map<string, FieldDataType>(
+        capabilities.commonLabels.map(l => [l, FieldDataType.Text] as [string, FieldDataType]));
```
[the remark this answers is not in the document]
bfcrampton: Done.

ui/src/app/testing/mock-capabilities.service.ts
```diff
@@ -0,0 +1,15 @@
+import {Injectable} from '@angular/core';
+
+import {CapabilitiesResponse} from '../shared/model/CapabilitiesResponse';
+import {CapabilitiesService} from '../core/capabilities.service';
+
+@Injectable()
+export class MockCapabilitiesService extends CapabilitiesService {
```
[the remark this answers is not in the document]
bfcrampton: Done.

servers/dsub/jobs/controllers/utils/query_parameters.py
```diff
@@ -30,13 +30,14 @@ def api_to_dsub(query):
...
         if query.labels.get('task-id'):
             dstat_params['task_ids'] = {query.labels['task-id']}
-        if query.labels.get('user-id'):
-            dstat_params['user_ids'] = {query.labels['user-id']}
-
         dstat_params['labels'] = {
             param_util.LabelParam(k, v)
             for (k, v) in query.labels.items()
-            if k not in ['job-id', 'task-id', 'user-id']
+            if k not in ['job-id', 'task-id']
```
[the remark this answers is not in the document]
bfcrampton: I think its happening correctly: https://github.com/DataBiosphere/job-manager/pull/222/files/91f0314db7f6f2cd347566b2d2abedd180e033f7#diff-edd5389e0640397bba1845481d51aeaaR101

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -95,11 +124,17 @@ export class URLSearchParamsUtils {
...
     }
-    return fields;
+    return new Map<string, FieldDataType>([
+      ...Array.from(queryDataTypes),
```
samanehsan: What is this `...` doing?

ui/src/app/job-details/job-details.component.html
samanehsan: I think that's fine!

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,76 @@
...
+          // Handle all not-activated errors by just returning false for
+          // this promise. All others, re-throw them.
+          if (reason == 'not-activated') {
+            return false;
+          }
+          throw reason;
+        });
+  }
+
+  private handleAuthCapabilities(capabilities: CapabilitiesResponse, path: String, url: String): Promise<CapabilitiesResponse> {
```
calbach: A `then` accepts a callback which can return a value, or a `Promise`; in this case since nothing async is happening, it would make more sense to just return the value directly, and throw to trigger the error path.

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,76 @@
...
+      })
+    } else if (path == 'sign_in') {
+      // Do not allow navigation to the sign in page when authentication is
+      // not required.
+      this.router.navigate(['']);
+      return Promise.reject('not-activated')
+    }
+  }
+
+  private handleProjectCapabilities(capabilities: CapabilitiesResponse, route: ActivatedRouteSnapshot): Promise<boolean> {
```
calbach: Same as above

ui/src/app/core/capabilities-activator.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    private readonly router: Router) {}
+
+  canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): Promise<boolean> {
+    return this.capabilitiesService.getCapabilities()
+      .then(cap => this.handleAuthCapabilities(cap, route.routeConfig.path, state.url))
+      .then(cap => this.handleProjectCapabilities(cap, route))
+      .catch(reason => {
+          // Handle all not-activated errors by just returning false for
+          // this promise. All others, re-throw them.
+          if (reason == 'not-activated') {
```
calbach: nit: constant

ui/src/app/core/capabilities.service.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  getCapabilitiesSynchronous(): CapabilitiesResponse {
+    if (this.capabilitiesResponse) {
+      return this.capabilitiesResponse;
+    }
+    throw new Error("CapabilitiesResponse has not been retrieved yet.")
+  }
+
+  getCapabilities(): Promise<CapabilitiesResponse> {
+    if (this.capabilitiesResponse) {
```
calbach: It would actually be good to still cache the promise here. In the event that two callers hit `getCapabilities()` on startup, we won't dispatch multiple requests.

ui/src/app/shared/utils/url-search-params.utils.ts
bfcrampton: `...` is the spread operator https://github.com/tc39/proposal-object-rest-spread 

Basically its just merging all three maps into a new map.

samanehsan: Oh that’s interesting! Thanks for clarifying!

ui/src/app/core/capabilities-activator.service.ts
bfcrampton: Done.

ui/src/app/core/capabilities-activator.service.ts
bfcrampton: This function actually does make an async call to `this.authService.isAuthenticated()`. But I changed the function to return `Promise<CapabilitiesResponse>|CapabilitiesResponse` so that we can return directly in cases where the async call is not made, and throw errors instead of rejecting the promise.

I changed it to return the value directly for `handleProjectCapabilities` because that does nothing async.

ui/src/app/core/capabilities.service.ts
bfcrampton: Done.

PR #224, opened 2018-02-26
Fix issues for Cromwell API deployments

rexwangcc:
There are a few issues with Cromwell API server that prevent us from deploying it painlessly:

- Cromwell API Server won't stand up without a config JSON file, even when one isn't required.
- The Cromwell shim will blow up if we want to stand up it for a locally hosted Cromwell instance, keeps complaining `Connection Refused`
- Invalid username-password pairs will make Cromwell shim blow up with a lot of "401" errors.
- The docker compose files are not good to use for a Kubernetes deployment.

This PR aims at solving the first 3 problems by:
1. Moving the entry point of Cromwell shim from `__main__.py` to `__init__.py`; adding several checkers before starting the API server. If the checks fail, the API server won't start.
2. Adding special instructions for local Cromwell instances.

For the last problem, a set of YAML files for Kubernetes deployment is now living in a separate "deploy" repository, we might migrate generalized version of them to this repository once we have a centralized "deploy" folder.


servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
...
     app.app.config.update(config)
+
 app.app.config['cromwell_url'] = args.cromwell_url
 app.app.json_encoder = JSONEncoder
 app.add_api('swagger.yaml', base_path=args.path_prefix)
 
-if __name__ == '__main__':
-    app.run(host='0.0.0.0', port=args.port)
+
+def run():
```
bfcrampton: Does this need to be in a separate method? If you put these checks on the top-level and left the entrypoint as `jobs.__main__:app` does that not work? Fine if not just wanted to check.

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
...
 config_path = os.environ.get('CROMWELL_CREDENTIALS')
-with open(config_path) as f:
-    config = json.load(f)
+
+# Check if the credentials are provided properly
+try:
+    with open(config_path) as f:
+        config = json.load(f)
+except (IOError, TypeError):
+    print(LOADER_ERR_MSG)
```
bfcrampton: Nit maybe add a bit more context:
```
print 'Failed to load config: {}'.format(LOADER_ERR_MSG)
```

bfcrampton: Also, any reason to print this error rather than raise an exception and stop execution?

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
     args, _ = parser.parse_known_args()
 
 app = connexion.App(__name__, specification_dir='./swagger/', swagger_ui=False)
+DEFAULT_CROMWELL_CREDENTIALS = {"cromwell_user": "", "cromwell_password": ""}
```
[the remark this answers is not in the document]
rexwangcc: Consistency is always great, will replace all double quotes, thanks!

servers/cromwell/jobs/__main__.py
rexwangcc: Sorry for the confusion here, I should have call it `LOADER_WARN_MSG`. The reason is that different Cromwell instances are using different authentication. Our team is using HTTP basic auth, which explicitly requires the `config.json`(credentials). But there are some other Cromwell instances using different proxies, for example, no basic auth required, but can only be accessible within a certain intranet. For those cases, config is not need to be provided by the user/developer, and stoping execution may confuse the users.

rexwangcc: Will change it to `LOADER_WARN_MSG` to eliminate the confusion.

servers/cromwell/jobs/__main__.py
rexwangcc: The goal here is to provide a startup check, if it fails, the Cromwell shim shouldn't even start. 

It might because I haven't found a good way for this, but if we point the gunicorn to`jobs.__main__:app`, even the checks failed, gunicorn can still find the application(global var `app.app`) and start it. But the first request to the shim API will then invoke the checks and the shim layer will blow up with thousands of same error messages :/

Flask's Resolvers may deal with this case but it looks more complicated at the first glance. So I moved the `app` var to `__init__.py` here to provide a feasible startup check. Please comment on here if there is a better way to deal with this.

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
...
+        response = requests.head(
+            args.cromwell_url,
+            auth=HTTPBasicAuth(app.app.config["cromwell_user"],
+                               app.app.config["cromwell_password"]),
+            timeout=5)
+        if response.status_code == 401:
+            raise requests.exceptions.HTTPError(
+                "Invalid credentials for the Cromwell: {0}".format(
+                    args.cromwell_url))
+        else:
```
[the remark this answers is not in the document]
rexwangcc: Ah, good catch, thanks!

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
...
+# Load credentials for cromwell
 config_path = os.environ.get('CROMWELL_CREDENTIALS')
-with open(config_path) as f:
-    config = json.load(f)
+
+# Check if the credentials are provided properly
+try:
+    with open(config_path) as f:
+        config = json.load(f)
+except (IOError, TypeError):
```
samanehsan: If there is a config.json, I think we want to check if it contains the `"cromwell_user"` and `"cromwell_password" keys, otherwise this would still throw an error when trying to check the credentials later on.

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
...
+                               app.app.config["cromwell_password"]),
+            timeout=5)
+        if response.status_code == 401:
+            raise requests.exceptions.HTTPError(
+                "Invalid credentials for the Cromwell: {0}".format(
+                    args.cromwell_url))
+        else:
+            return app.app
+    except requests.exceptions.RequestException as err:
+        print(err)
```
samanehsan: How about using logging instead of printing these errors?

samanehsan:
Thanks for making these changes @rexwangcc! I left a few comments on your PR!

servers/cromwell/jobs/__main__.py
bfcrampton: Gotcha, thanks for the explanation that sounds good!

servers/cromwell/jobs/__main__.py
bfcrampton: Gotcha, didn't realize the app would still run if an exception was thrown in `__main__`. This sounds like a reasonable solution to me, thanks for the explanation!

servers/cromwell/jobs/__main__.py
```diff
@@ -31,13 +33,42 @@
     args, _ = parser.parse_known_args()
 
 app = connexion.App(__name__, specification_dir='./swagger/', swagger_ui=False)
+DEFAULT_CROMWELL_CREDENTIALS = {"cromwell_user": "", "cromwell_password": ""}
+LOADER_ERR_MSG = "Failed to load config.json, using the default config: {0}".format(
```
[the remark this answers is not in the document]
rexwangcc: Good point, will do, thanks!

servers/cromwell/jobs/__main__.py
rexwangcc: That's a good catch! I should have caught this... will do!

servers/cromwell/jobs/__main__.py
rexwangcc: That's a good idea! Will refactor it, thanks!

PR #225, opened 2018-02-27
[dsub] Implement querying by start time

### LINKED ISSUE #54: QueryJobs by start queries by submission
This is a relatively minor difference, but it's not working as documented in the API. Pipelines API actually only has support for querying by creation time, not start time, so we'd need to add support or change the API.

bfcrampton:
#### Resolves #54 
- Make `query.start` actually query by start time instead of submission
- Add support for `query.extensions.submission`

### DIFF
--- a/servers/dsub/jobs/controllers/jobs_controller.py
+++ b/servers/dsub/jobs/controllers/jobs_controller.py
@@ -121,21 +121,29 @@ def query_jobs(body):
         query.page_token) or (None, None)
     query.page_size = min(query.page_size or _DEFAULT_PAGE_SIZE,
                           _MAX_PAGE_SIZE)
+
+    query.start = query.start.replace(tzinfo=tzlocal()).replace(
+        microsecond=0) if query.start else None
+    query.end = query.end.replace(tzinfo=tzlocal()).replace(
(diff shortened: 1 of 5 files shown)


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -121,21 +121,27 @@ def query_jobs(body):
...
-            raise BadRequest(
-                "Invalid query: start date is invalid with pagination token.")
-    if query.end:
-        query.end = query.end.replace(tzinfo=tzlocal()).replace(microsecond=0)
-        if query.start and query.start >= query.end:
-            raise BadRequest(
-                "Invalid query: start date must precede end date.")
+    if query.submission and query.start and query.submission >= query.start:
+        raise BadRequest(
+            "Invalid query: submission date must precede start date.")
```
calbach: I'd probably go with <= instead. "submitted and started after time t" seems like it should be a valid query, and there's nothing semantically incorrect about submit->start taking <1s.

bfcrampton: Good point, done.

PR #226, opened 2018-02-27
[UI] Fix incorrect duration calculation

bfcrampton:


ui/src/app/shared/pipes/duration.pipe.ts
```diff
@@ -12,7 +12,7 @@ export class DurationPipe implements PipeTransform {
     } else {
       duration = new Date().getTime() - start.getTime();
     }
-    return Math.round(duration/3600000) + "h " +
-      Math.round(duration/60000)%60 + "m";
+    return Math.floor(duration/3600000) + "h " +
+      Math.floor(duration/60000)%60 + "m";
```
rexwangcc: Sometimes there will be a slight difference after the change here: 
For `duration=990000`, `Math.round(duration/3600000) + "h " +Math.round(duration/60000)%60 + "m"` will return "0h 17m" while `Math.floor(duration/3600000) + "h " +Math.floor(duration/60000)%60 + "m"` would return "0h 16m". But I don't think that is a big deal as this is for the rough "duration" column, so it looks good to me.

bfcrampton: 990000 milliseconds is equivalent to 16.5 minutes. I think the previous behavior of rounding this  up to 17 minutes is incorrect because it is misrepresenting the information. Additionally, if we added seconds to this, the rounding up of minutes would give the incorrect time: "17 m 30s" when it should be "16m 30s". Truncating over rounding here always makes sense.

rexwangcc: Yeah, this makes sense to me. Thanks for the explanation here!

PR #227, opened 2018-02-27
[CI] Split UI and backend tests, cache dependencies

bfcrampton:


.circleci/config.yml
```diff
@@ -4,49 +4,69 @@
...
 version: 2
 jobs:
-  build:
+  ui:
+    machine:
+      image: circleci/classic:latest
+    steps:
+      - checkout
+      - restore_cache:
+          key: node-modules-{{ .Branch }}--{{ checksum "ui/package.json" }}
```
calbach: I'd include the secondary key here and below as well (with just the branch); this will mean quicker cached builds when we change package.json.

.circleci/config.yml
```diff
@@ -4,49 +4,69 @@
...
-          name: Lint angular code with ng lint
-          command: |
-            nvm use 8.5.0
-            cd ui
-            ng lint --type-check
-      # Unit and Integration Tests
-      # TODO(calbach): Cache the tox venvs per https://discuss.circleci.com/t/cached-venv-in-tox/1499.
       - run:
           name: Run dsub server tests
           command: |
```
calbach: opt: if you feel like including it in this PR, it should be a relatively easy change to make google tests only run on master:

```
if [ -z "${GCLOUD_SERVICE_KEY}"  || "${CIRCLE_BRANCH}" != "master" ]; then
  ...
```

or you could even exclude the check for gcloud_service_key I guess (and assume master implies presence of the key).

.circleci/config.yml
bfcrampton: Per offline discussion, this could cause issues when removing / downgrading a dependency.

.circleci/config.yml
bfcrampton: Good call, done.

PR #228, opened 2018-02-28
Add general loading + jobs table specific loading

calbach:
- During all navigation, show an unobtrusive "Loading..." snackbar; this is a low-fidelity solution which doesn't cover the screen but provides user feedback that something is happening.
  - This primarily covers projects -> jobs list, jobs list -> job details transitions
- During job list reloads, show a mat-spinner + white transparent overlay on the table only; this behaves specially due to the following requirements:
  - In our ideal UX we'd like to leave the filter controls / query builder available during a query, e.g. if someone is typing in a series of chips, we don't want to block entry of the next chip on page load.
  - The job list component is special anyways, since its a singleton and navigation to itself doesn't block on data load
- Harden the jobs list towards multiple concurrent requests, which are easy to trigger with the query builder

### DIFF
--- a/ui/src/app/core/notify-loading.service.ts
+++ b/ui/src/app/core/notify-loading.service.ts
@@ -1,10 +1,6 @@
 import {
-  Component,
-  ElementRef,
   Injectable,
   NgZone,
-  OnInit,
-  ViewChild,
 } from '@angular/core';
 import {
   Event as RouterEvent,
@@ -15,7 +11,6 @@ import {
   Router,
 } from '@angular/router';
 import {MatSnackBar, MatSnackBarRef, SimpleSnackBar} from '@angular/material';
-import {Observable} from 'rxjs/Observable';
 
 @Injectable()
 export class NotifyLoadingService {
--- a/ui/src/app/core/notify-loading.service.ts
+++ b/ui/src/app/core/notify-loading.service.ts
@@ -1,0 +1,57 @@
+import {
+  Component,
+  ElementRef,
+  Injectable,
+  NgZone,
+  OnInit,
+  ViewChild,
+} from '@angular/core';
+import {
+  Event as RouterEvent,
+  NavigationCancel,
+  NavigationEnd,
+  NavigationError,
+  NavigationStart,
+  Router,
+} from '@angular/router';
+import {MatSnackBar, MatSnackBarRef, SimpleSnackBar} from '@angular/material';
+import {Observable} from 'rxjs/Observable';
+
+@Injectable()
+export class NotifyLoadingService {
+  private snackBarRef: MatSnackBarRef<SimpleSnackBar>;
+
+  constructor(
+    private zone: NgZone,
+    private snackBar: MatSnackBar,
+    router: Router,
+  ) {
+    router.events.subscribe((e: RouterEvent) => {
+      if (e instanceof NavigationStart) {
+        this.showSpinner();
+      } else if (e instanceof NavigationEnd
+                 || e instanceof NavigationError
+                 || e instanceof NavigationCancel) {
+        this.hideSpinner();
+      }
+    });
+  }
+
+  private showSpinner() {
+    this.zone.runOutsideAngular(() => {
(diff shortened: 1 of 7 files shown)


ui/src/app/job-list/job-list.component.ts
```diff
@@ -65,15 +66,23 @@ export class JobListComponent implements OnInit {
 
   reloadJobs(query: string) {
     if (this.streamSubscription) {
+      this.loading = true;
       this.streamSubscription.unsubscribe();
-      this.jobStream = new JobStream(this.jobManagerService,
+      const nextStream = new JobStream(this.jobManagerService,
           URLSearchParamsUtils.unpackURLSearchParams(query));
-      this.jobStream.loadAtLeast(initialBackendPageSize)
+      nextStream.loadAtLeast(initialBackendPageSize)
```
bfcrampton: Is there any way to cancel previous requests so that they don't block the next one?

calbach: This is probably a larger change. It might involve switching our Promise usage over to Observables, as you unfortunately can't cancel a Promise.

Alternatively, we could return a cancellation callback (or accept that as a parameter); would need to plumb down to the job manager service.

Added a TODO.

calbach:
Hm test failing on Circle (though it worked locally) - I'll dig into that when I can. PTAL in the meantime.

ui/src/app/core/notify-loading.service.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  Component,
```
samanehsan: It looks like several of these imports are not being used and can be removed including `Component`, `ElementRef`, `OnInit`, `ViewChild` and `{Observable}`

calbach: Done.

calbach:
Testcase is now fixed as well.

@alahwa thoughts on the behavior?

ui/src/app/job-list/job-list.component.html
```diff
@@ -2,7 +2,12 @@
   [showControls]="true"
   [jobs]="jobs"
   [pageSize]="pageSize"></jm-header>
-<jm-job-list-table
-  [dataSource]="dataSource"
-  (onJobsChanged)="handleJobsChanged()">
-</jm-job-list-table>
+<div class="table-container">
```
alahwa: Ideally, we'd have a directive here that changes the opacity of the table component. That would natively work with any theme and avoid the potential for bugs like mispositioned overlays. Material components can be a pain when it comes to styling, but have you tried this here?

calbach: Could you expand on why you think that approach is ideal? Seems a bit unnatural to me, but I can see advantages e.g. this would look wrong if we changed the app's background color.

It also seems reasonable to block clicks to the table rows, which the overlay currently achieves, though this is not strictly necessary.

alahwa: In the past, I've run into bugs implementing overlays like this in Android. Stuff like: underlying content is clickable if you press exactly at the edges of the overlay, overlay appears misaligned in certain environments, content appears to flash on load in/out and might be briefly clickable depending on the order in which the framework constructs/tears down views. In the end I just replaced all the overlays with my aforementioned solution. I have no reason to believe that we'll see any of these bugs in particular, but opting for fewer moving parts exposes us to less HTML/CSS trickery. If you'd like the table to be unclickable while loading, directly disabling pointer actions on the table is also going to be safer than blocking access with an overlay to absorb mouse events.

This is more of a preemptive suggestion than a blocking change, so we can leave it as is unless it does actually cause problems.

calbach: Done, kind of.

I moved the opacity setting onto the table element; lets try that out, I like the idea of not applying a particular color on the overlay.

Secondly though, I didn't remove the overlay. My primary purpose for that is to control the positioning of the spinner element (horizontally centered). The continues to have a secondary role of blocking clicks, which I don't feel particularly strongly about.

PR #232, opened 2018-03-12
Support ng test with chrome headless, use for Circle

calbach:


.circleci/config.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           command: |
             /opt/circleci/.nvm/nvm.sh use 8.5.0
             cd ui
-            ng test --watch=false -sm=false
+            ng test --watch=false -sm=false --browsers=ChromeHeadless
```
bfcrampton: tiny nit: Would you mind actually change `--watch=false` to `--single-run`. I added the watch flag but I think single run is much clearer.

calbach: Done.

calbach:
I think there may be an issue with the dsub tests, mine just took 15m which seems way too long if we excluded google tests properly. There's also a suspicious bash error:
https://circleci.com/gh/DataBiosphere/job-manager/819?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Will look if I have the chance

pgrosu:
@calbach Regarding the Bash issue, for your `if` statement just turn your single-brackets to double-brackets `[[ ... ]]`  and the error should disappear, as follows:

```Bash
if [[ -z "$GCLOUD_SERVICE_KEY" || "${CIRCLE_BRANCH}" != "master" ]]; then
  detox -n 8 -c ./servers/dsub -- \
    --exclude-test=jobs.test.test_jobs_controller_google.TestJobsControllerGoogle
else
  echo $GCLOUD_SERVICE_KEY | base64 --decode --ignore-garbage > /home/circleci/gcloud-service-key.json
  detox -n 8 -c ./servers/dsub
  rm /home/circleci/gcloud-service-key.json
fi
```

Paul

pgrosu:
Regarding the `dsub` tests performance hit, it probably is `dtox` that might be the culprit.  Wouldn't hurt to do some profiling to see where the bottleneck might be.

~p

bfcrampton:
Hmmm yeah it definitely should not be taking that long without google provider tests. Maybe we should try switching back to `tox` so we can see more details.

Thanks for the fix @pgrosu! Fixed here: https://github.com/DataBiosphere/job-manager/pull/234

pgrosu:
Hi Bryan (@bfcrampton),

Sure thing :)  Yes, `tox` should be faster than the distributed version under some conditions.  `dtox` just performs a `GreenPool` spawn request with a `waitall()` as a semaphore, which could block at capacity.

~p

PR #233, opened 2018-03-12
[UI] Display extended fields on job details page

bfcrampton:


ui/src/app/job-details/panels/panels.component.ts
```diff
@@ -45,6 +50,8 @@ export class JobPanelsComponent implements OnChanges {
     this.inputs = Object.keys(this.job.inputs || {}).sort();
     this.outputs = Object.keys(this.job.outputs || {}).sort();
     this.labels = Object.keys(this.job.labels || {}).sort();
+    this.extensions = Object.keys(this.job.extensions || {}).sort();
```
calbach: Perhaps it would make sense to do filtering by type here in addition to or instead of the blacklisting? Wouldn't `tasks` get rendered with the current code?

I would also not be opposed to a whitelst approach, or alternatively just hardcoding extension fields into the template; not sure if we plan to give any of these special treatment.

bfcrampton: Good point, I think whitelist is a better approach so that any new extensions aren't automatically displayed (in case they are the wrong type). Changed to that.

ui/src/app/job-details/panels/panels.component.ts
```diff
@@ -16,13 +16,18 @@ import {ResourceUtils} from '../../shared/utils/resource-utils';
   styleUrls: ['./panels.component.css'],
 })
 export class JobPanelsComponent implements OnChanges {
+  // Filter out 'envs' and 'logs' from extendedFields since these are maps and
```
calbach: Update comment

ui/src/app/job-details/panels/panels.component.ts
```diff
@@ -16,13 +16,18 @@ import {ResourceUtils} from '../../shared/utils/resource-utils';
...
+  // require special treatment for display.
+  private readonly extensionsWhiteList: string[] = ['userId', 'statusDetail', 'lastUpdate', 'parentJobId'];
```
calbach: I'd do one of two things here:

- Use this array to inform the order on the details page
- Use a Set instead

ui/src/app/job-details/panels/panels.component.ts
bfcrampton: Done.

ui/src/app/job-details/panels/panels.component.ts
bfcrampton: Done, also cleaned up a few small things.

PR #234, opened 2018-03-13
[CI] Fix bash error in circle config

bfcrampton:


calbach:
1m38s, that's more what I expected. Nice - thanks.

PR #235, opened 2018-03-13
[UI] Job details change inputs/outputs into table rather than drop downs

bfcrampton:
Also adds some nice functionality for the input/output links to a GCS browser. Now, for non-wildcard paths, the filename will be embedded as a filter 'prefix' in the URL. This is useful when the directory on GCS has a lot of files, the link will now filter to the specific one you clicked on (plus any files which are prefixed with the filename, i.e. if you click `foo.bar` and `foo.bar.gz` exists in the bucket it will still show up).

Before merging this PR I will open another which adds the inlined logs/scripts, since this one removes logs from the details page completely. Also going to go over this with marty tomorrow.

### DIFF
--- a/ui/src/app/job-details/panels/panels.component.css
+++ b/ui/src/app/job-details/panels/panels.component.css
@@ -38,7 +38,11 @@ p {
 }
 
 .card.input-output {
-  width: 30;
+  width: 30%;
+}
+
+.mat-tab-labels{
+  justify-content: center;
 }
 
 .input-output-table {
--- a/ui/src/app/job-details/panels/panels.component.css
+++ b/ui/src/app/job-details/panels/panels.component.css
@@ -26,10 +26,21 @@ p {
   margin-right: 3rem;
 }
 
-.mat-card {
-  width: 25%;
+.card {
   margin-right: 1rem;
   margin-top: 2rem;
   float: left;
   min-height: 8.7rem;
+}
+
+.card.status {
+  width: 20%;
+}
+
+.card.input-output {
+  width: 30;
+}
+
+.input-output-table {
+  overflow: hidden;
 }
--- a/ui/src/app/job-details/panels/resources-table/resources-table.component.ts
+++ b/ui/src/app/job-details/panels/resources-table/resources-table.component.ts
@@ -1,0 +1,36 @@
+import {
+  Component,
+  Input,
+  OnChanges,
+  SimpleChanges
+} from '@angular/core';
+import {ResourceUtils} from '../../../shared/utils/resource-utils';
+
+
+@Component({
+  selector: 'jm-resources-table',
+  templateUrl: './resources-table.component.html',
+  styleUrls: ['./resources-table.component.css'],
+})
+export class ResourcesTableComponent implements OnChanges {
+  private readonly resourcesColumns: string[] = ["key", "value"];
+
+  @Input() entries: Object;
+  entryKeys: Array<string>;
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.entryKeys = Object.keys(this.entries || {}).sort();
+  }
+
+  getResourceURL(key: string): string {
+    return ResourceUtils.getResourceBrowserURL(this.entries[key]);
+  }
+
+  getResourceFileName(key: string): string {
+    return ResourceUtils.getResourceFileName(this.entries[key]);
+  }
+
+  isResourceURL(key: string): boolean {
+    return ResourceUtils.isResourceURL(this.entries[key]);
+  }
+}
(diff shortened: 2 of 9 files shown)


ui/src/app/job-details/panels/panels.component.css
```diff
@@ -26,10 +26,21 @@ p {
...
   float: left;
   min-height: 8.7rem;
 }
+
+.card.status {
+  width: 20%;
+}
+
+.card.input-output {
+  width: 30;
```
calbach: Units?

ui/src/app/job-details/panels/panels.component.css
```diff
@@ -26,10 +26,21 @@ p {
...
+}
+
+.input-output-table {
```
calbach: Should the table / card contents be scrollable? Is it already?

ui/src/app/job-details/panels/resources-table/resources-table.component.ts
```diff
@@ -0,0 +1,36 @@
...
+import {ResourceUtils} from '../../../shared/utils/resource-utils';
+
+
+@Component({
+  selector: 'jm-resources-table',
+  templateUrl: './resources-table.component.html',
+  styleUrls: ['./resources-table.component.css'],
+})
+export class ResourcesTableComponent implements OnChanges {
+  readonly resourcesColumns: string[] = ["key", "value"];
```
calbach: This will produce a table with headers of "key" and "value" right? If so may be worth going over with Marty whether we should try and just hide that entirely.

ui/src/app/shared/utils/resource-utils.ts
```diff
@@ -3,19 +3,34 @@ export class ResourceUtils {
...
+    let parts = ResourceUtils.validateGcsURLGetParts(url);
     // This excludes the object from the link to show the enclosing directory.
     // This is valid with wildcard glob (bucket/path/*) and directories
     // (bucket/path/dir/) as well, the * or empty string will be trimmed.
-    return parts ? ResourceUtils.browserPrefix + parts.slice(2,-1).join("/") : undefined;
+    if (!parts) {
+      return undefined;
+    }
+
+    var browserUrl = ResourceUtils.browserPrefix + parts.slice(2,-1).join("/");
```
calbach: nit: single quotes for consistency

ui/src/app/shared/utils/resource-utils.ts
```diff
@@ -3,19 +3,34 @@ export class ResourceUtils {
...
+    if (url.indexOf('*') == -1) {
+      browserUrl += "?prefix=" + ResourceUtils.getResourceFileName(url);
```
calbach: Cool

ui/src/app/shared/utils/resource-utils.ts
```diff
@@ -32,14 +47,15 @@ export class ResourceUtils {
   }
 
   /** Parse file name from gs link */
-  public static getResourceFileName(value: string): string {
-    let parts = ResourceUtils.validateGcsURLGetParts(value);
-    let formattedValue = value;
-    if (parts && parts.length > 3 && parts[parts.length - 1]) {
+  public static getResourceFileName(url: string): string {
+    let parts = ResourceUtils.validateGcsURLGetParts(url);
```
calbach: nit const? Here and for parts above?

ui/src/app/job-details/panels/panels.component.css
bfcrampton: Whoops, added.

ui/src/app/job-details/panels/resources-table/resources-table.component.ts
bfcrampton: It actually does not since we don't have a `mat-header-row` in the template using those columns. I thought it looked pretty bad with those headers. 

However,  this is basically unnecessary it can just be inlined into the template since its used once, so removed.

ui/src/app/shared/utils/resource-utils.ts
bfcrampton: Done.

ui/src/app/shared/utils/resource-utils.ts
bfcrampton: Yeah I forgot to mention this addition in the description. Its actually quite nice when there's a bunch a junk all in the same directory this URL will filter it. Unfortunately doesn't work with wildcards though.

ui/src/app/job-details/panels/panels.component.css
bfcrampton: This disables the scroll bar which was weirdly appearing even when there was nothing to scroll. I'm going to ask marty today about whether he thinks it should be scrollable. 

I imagine we should show some default number of rows and either have a button to expand it or make it scrollable.

ui/src/app/shared/utils/resource-utils.ts
bfcrampton: Done.

bfcrampton:
This is superseded by https://github.com/DataBiosphere/job-manager/pull/239

PR #238, opened 2018-03-19
Add support for cromwell-as-a-service

samanehsan:
Support making API requests to cromwell-as-a-service using a bearer token (requires google sign in to be enabled).

### DIFF
--- a/cromwell-compose-caas.yml
+++ b/cromwell-compose-caas.yml
@@ -4,7 +4,7 @@ services:
     extends:
       file: common-compose.yml
       service: ui
-    command: ["npm", "run-script", "ng", "--", "serve", "--host", "ui", "--env=cromwell_staging"]
+    command: ["npm", "run-script", "ng", "--", "serve", "--host", "ui", "--env=cromwell_caas"]
   cromwell:
     build:
       context: .
@@ -13,7 +13,8 @@ services:
       file: common-compose.yml
       service: cromwell
     environment:
-      - USE_CAAS=False
+      - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1}
+      - USE_CAAS=True
   jobs-proxy:
     extends:
       file: common-compose.yml
--- a/servers/cromwell/jobs/__main__.py
+++ b/servers/cromwell/jobs/__main__.py
@@ -60,7 +60,7 @@ finally:
     app.app.config.update(config)
 
 app.app.config['cromwell_url'] = args.cromwell_url
-app.app.config['use_caas'] = args.use_caas.lower() == 'true'
+app.app.config['use_caas'] = args.use_caas and args.use_caas.lower() == 'true'
 app.app.json_encoder = JSONEncoder
 app.add_api('swagger.yaml', base_path=args.path_prefix)
 
--- a/servers/cromwell/jobs/__main__.py
+++ b/servers/cromwell/jobs/__main__.py
@@ -19,6 +19,11 @@ parser.add_argument(
     help='Url for fetching data from cromwell',
     default=os.environ.get('CROMWELL_URL'))
 parser.add_argument(
+    '--use_caas',
+    type=str,
+    help='Whether the cromwell backend is using cromwell-as-a-service',
+    default=os.environ.get('USE_CAAS'))
+parser.add_argument(
     '--path_prefix',
     type=str,
     help='Path prefix, e.g. /api/v1, to serve from',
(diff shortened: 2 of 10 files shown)


common-compose.yml
```diff
@@ -36,6 +36,7 @@ services:
       - PATH_PREFIX=/api/v1
       - CROMWELL_CREDENTIALS=/app/jobs/config.json
       - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1}
+      - USE_CAAS=0
```
bfcrampton: nit: Since these environment vars now differ between cromwell compose files, can you move them out of the common compose and only into the extended compose files? Our general pattern has been to only add environment vars which are not overriden in extended compose files.

bfcrampton: tiny nit: I may also just call this `REQUIRES_AUTH` to match the dsub implementation. It looks like you only switch on this flag for returning the capabilities response and getting the bearer token from the header, which isn't really specific to CAAS. 

We may be able to extract some of the auth token stuff out to jm_utils in the future.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -29,9 +29,13 @@ def abort_job(id):
 
     :rtype: None
     """
+    auth_token = request.headers.get('Authentication')
```
bfcrampton: This is going to keep the `Bearer ` piece on. Is that intended? i.e. does cromwell expect a bearer auth token?

cromwell-compose-caas.yml
```diff
@@ -0,0 +1,24 @@
...
+  cromwell:
+    build:
+      context: .
+      dockerfile: servers/cromwell/Dockerfile
+    extends:
+      file: common-compose.yml
+      service: cromwell
+    environment:
+      - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1}
+      - USE_CAAS=1
```
dshiga: maybe true/false would be better than 0/1 here, just a little clearer

servers/cromwell/jobs/controllers/capabilities_controller.py
```diff
@@ -23,3 +25,12 @@ def get_capabilities():
             'cromwell-workflow-name', 'cromwell-workflow-id', 'comment'
         ],
         query_extensions=[])
+    if current_app.config['use_caas']:
+        capabilities.authentication = AuthenticationCapability(
+            is_required=True,
+            scopes=[
+                'profile', 'email', 'openid',
```
dshiga: How did you determine what scopes to use here? It looks pretty different from the scopes used on the dsub side, which is a bit surprising to me because both dsub and cromwell ultimately interact with GCP via Pipelines API.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -47,10 +51,14 @@ def update_job_labels(id, body):
 
     :rtype: UpdateJobLabelsResponse
     """
+    auth_token = request.headers.get('Authentication')
```
dshiga: It might be nice to break these two lines into a function that returns auth_token and auth, or even a decorator so you don't have to repeat as much. This might also make it possible to write tests for the auth code.

servers/cromwell/jobs/controllers/jobs_controller.py
samanehsan: Yes, cromwell-as-a-service expects a bearer auth token

common-compose.yml
samanehsan: I kept this specific to using caas for the future case where we want users to log in with a gmail account to the job manager so they can see their preferences, but still connect to a cromwell instance that requires just a username/password. In that case, `REQUIRES_AUTH` would be true but we would not send a bearer token to cromwell.

servers/cromwell/jobs/controllers/capabilities_controller.py
samanehsan: Thanks for pointing that out Dave! Looking at this again, I believe we want to use these scopes: https://github.com/broadinstitute/firecloud-tools/blob/740ea4289a2234ac7a232704f0b4c4d00ba7a5b2/register_service_account/register_service_account.py#L18

common-compose.yml
bfcrampton: Gotcha, makes sense!

servers/cromwell/jobs/__main__.py
```diff
@@ -20,7 +20,7 @@
     default=os.environ.get('CROMWELL_URL'))
 parser.add_argument(
     '--use_caas',
-    type=int,
+    type=str,
```
dshiga: Would it work to use type boolean here or else use bool() below like you were doing before?

servers/cromwell/jobs/controllers/utils/auth.py
```diff
@@ -0,0 +1,29 @@
+from flask import current_app, request
+from requests.auth import HTTPBasicAuth
+from werkzeug.exceptions import Unauthorized
+
+
+def requires_auth(fn):
```
dshiga: I like this! Could you test using mock request and either decorating a function in your tests or else calling requires_auth directly?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -30,12 +31,14 @@ def abort_job(id):
...
         cromwell_url=_get_base_url(), id=id)
-    response = requests.post(url, auth=_get_user_auth())
+    response = requests.post(
+        url, auth=kwargs.get('auth'), headers=kwargs.get('auth_headers'))
     if response.status_code == NotFound.code:
         raise NotFound(response.json()['message'])
 
 
-def update_job_labels(id, body):
+@requires_auth
```
bfcrampton: Very cool I like it

servers/cromwell/jobs/__main__.py
samanehsan: I believe using type boolean here would be the same as doing `bool(value)`. But, if the value is the string `False` it will still get evaluated to true.

dshiga: Ah okay, so better to keep this as str here.

samanehsan:
This is ready for another look @dshiga!

PR #239, opened 2018-03-21
[UI] Revamp Job Details Page

bfcrampton:
Major update is adding resources expansion panel to details page. This includes inputs, outputs, script, and log files. All tabs are only shown if available. If none of the resources tabs are available the entire panel will not be shown. If there is no tasks panel, the resources panel will be always expanded.
See screenshot below.

API change:
- Add `script` to extended fields

More minor UI fixes:
- Input/output file links now open up the GCS browser with a prefix to filter to the the relevant file
- Labels are now separated horizontally along the top per wire-frames (with no wrapping labels across lines)
- New "Details" card to display white-listed extended fields (if present). 
  - Long fields which are more than a line are truncated with an ellipsis.
- Fix to "Graph View" tab so that click it won't change the visually selected tab (since it currently opens a new tab when clicked)


bfcrampton:
screen shot 2018-03-21 at 11 15 01 am
screen shot 2018-03-21 at 11 15 17 am
screen shot 2018-03-21 at 11 15 23 am

servers/cromwell/jobs/swagger/swagger.yaml
```diff
@@ -371,25 +371,28 @@ definitions:
...
         type: "string"
         format: "date-time"
-        description: "Job last update datetime in ISO8601 format"
+        description: "Job last update datetime in ISO8601 format."
       envs:
         type: "object"
-        description: "Map of ENV variables associated with the job"
+        description: "Map of ENV variables key values associated with the job."
         properties: {}
+      script:
```
calbach: Per offline discussion, perhaps consider making this something more generic such as `sourceFile`? This could then more easily be be used to, for example, display a Cromwell WDL file; even though I guess we don't have a way of displaying that currently.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  hasOutputs(): boolean {
+    return this.outputs.length > 0;
+  }
+
+  hasScript(): boolean {
+    return this.job.extensions && !!this.job.extensions.script;
+  }
+
+  keepExpanded(): boolean {
```
calbach: nit: `alwaysExpanded`?

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  script: string = '';
+  inputs: Array<string> = [];
+  logFileData: Map<string, string> = new Map();
+  logFiles: Array<string> = [];
+  outputs: Array<string> = [];
+  currentResourcesTab: string = '';
+
+  constructor(private readonly gcsService: GcsService) {}
+
+  ngOnChanges(changes: SimpleChanges) {
```
calbach: Why `onChanges` (vs `onInit`) here? Is the job metadata response not necessarily present at initialization time?

ui/src/app/job-details/tasks/tasks.component.ts
```diff
@@ -58,13 +59,15 @@ export class TaskDetailsComponent implements OnInit, OnChanges {
...
-       return '/jobs/' + this.job.id;
-     }
-   }  
+  hasTimingUrl(): boolean {
+    return this.job.extensions && !!this.job.extensions.timingUrl;
+  }
+
+   tabChanged(event: MatTabChangeEvent) {
+    event.tab.isActive = false;
+    debugger;
```
calbach: rm

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  hasScript(): boolean {
+    return this.job.extensions && !!this.job.extensions.script;
+  }
+
+  keepExpanded(): boolean {
+    return !this.job.extensions || !this.job.extensions.tasks;
+  }
+
+  tabChanged(event: MatTabChangeEvent) {
+    this.currentResourcesTab = event.tab.textLabel;
```
calbach: Is there something you could use in lieu of the display text as the identifier here? If a backend returned "Script" or any of the other hardcoded tabs as a log file name, there would be a collision here.

servers/cromwell/jobs/__init__.py
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env python
-
-from __main__ import run
```
samanehsan: Did you mean to remove this? The cromwell server won't run without this 😞

servers/dsub/jobs/controllers/utils/extensions.py
```diff
@@ -11,9 +11,16 @@ def get_extensions(job):
         Returns:
             ExtendedFields: Populated extensions on the job
     """
+    envs = job['envs']
+    script = None
+    if envs and envs['_SCRIPT']:
+        script = envs['_SCRIPT']
+        del envs['_SCRIPT']
```
samanehsan: Why do you need to delete `envs['_SCRIPT']` here?

samanehsan:
Not sure if this is within the scope of this PR, but inputs that are arrays of gs resource urls are not getting linked:
 
screen shot 2018-03-22 at 2 59 01 pm

ui/src/app/job-details/resources/resources-table/resources-table.component.html
```diff
@@ -0,0 +1,13 @@
+<mat-table class="resources-table" #table [dataSource]="entryKeys">
+  <ng-container matColumnDef="key">
+    <mat-cell *matCellDef="let key"> {{key}} </mat-cell>
```
samanehsan: nit: Please add spaces around the `key` variable for consistency

samanehsan:
This looks really good @bfcrampton! I just have a few comments on your PR

ui/src/app/job-details/resources/resources.component.ts
bfcrampton: Really good point, its a bit tough since the tabs are dynamic based on what data is present so we can't just use an index. I came up with a solution I'm okay with which basically gives each tab an "ID" mapped to its title. For top-level fields this is the same as the name, and for log files it is prefixed with `"log-"`.

Let me know what you think.

ui/src/app/job-details/resources/resources-table/resources-table.component.html
bfcrampton: Done.

bfcrampton:
@samanehsan hmmm I hadn't realized that it was possible to have an input be an array of file paths. 

Seems like we'll need to think about the UX for that a bit, so mind if we push that to another PR since this ones big already?

servers/dsub/jobs/controllers/utils/extensions.py
samanehsan: Ok, thanks for the clarification!

samanehsan:
@bfcrampton, I think it's fine to handle arrays of file paths in a separate PR!

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -18,78 +17,64 @@ import {GcsService} from '../../core/gcs.service';
...
-      if (this.job.extensions.script) {
-        this.script = this.job.extensions.script;
-      }
-    }
+  constructor(private readonly gcsService: GcsService) {}
 
+  ngOnInit() {
     if (this.job.inputs) {
       this.inputs = Object.keys(this.job.inputs).sort();
+      this.tabTitles.set('Inputs', 'inputs');
```
calbach: Isn't this going in the wrong direction?

ui/src/app/job-details/resources/resources.component.ts
calbach: Ah, this was likely a carry-over from the pre-Resolver days where the job would late-load in.

ui/src/app/job-details/panels/panels.component.css
```diff
@@ -22,14 +22,28 @@ p {
...
-  margin-top: 2rem;
+.card {
   float: left;
-  min-height: 8.7rem;
+  width: 25%;
+  height: 10rem;
+  margin: 1rem;
+}
+
+/* TODO(bryancrampton): Add hover-over to display full detail field, if truncated */
```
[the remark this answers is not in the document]
calbach: Sounds good

ui/src/app/job-details/resources/resources.component.ts
calbach: Hmmm, I think this approach still has the same issue, since you are keying the map by the tab name.

Perhaps array of tab "ID" is the right approach. Noting that you'd need to also splice out log data if they fail to load. currentResourcesTab would then have the tab ID value

ui/src/app/core/gcs.service.ts
```diff
@@ -0,0 +1,33 @@
...
+  readObject(bucket: string, object: string): Promise<string> {
+    return this.authService.isAuthenticated().then( authenticated => {
+      if (authenticated) {
+        return gapi.client.request({
+          path: `${GcsService.apiPrefix}/b/${bucket}/o/${object}`,
+          params: {alt: 'media'}
+        })
+        .then(response => response.body)
+         // If an error occurs ignore it, the file may not exist
+        .catch(response => "")
```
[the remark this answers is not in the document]
calbach: Good point. Can we detect that case specifically here? Would be nice to continue doing what you are now for "does not exist" (note, this might be a 403 rather than a 404, don't remember what GCS does) and give some notification or show some text here for "random transient error".

ui/src/app/job-details/resources/resources.component.ts
calbach: (see other comment on the tabId approach)

bfcrampton: Fixed with updated strategy.

bfcrampton:
Added error handling when the file fails to load other than for 404s, and updated tab switching approach per comments.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -0,0 +1,138 @@
...
+  tabs(): string[] {
+    return this.tabIds.map(id => this.tabTitles.get(id));
+  }
+
+  logFiles(): string[] {
+    return Array.from(this.logFileData.keys());
+  }
+
+  tabChanged(event: MatTabChangeEvent) {
+    this.currentTabId = this.tabIdForTitle(event.tab.textLabel);
```
calbach: I still don't think this problem is resolved, but let me know if I'm missing something. I had meant to imply that you would start using the tab index here instead. The properties are a bit confusing, but I think `this.tabIds[event.tab.origin + event.tab.position]` gives you the right result?

bfcrampton: Good point, you're right that doesn't fix it since we're still looking up by title.

I'm not sure what `tab.origin` and `tab.position` are, but they do not seem to represent the index of the tab––curiously it almost always seems to be 1 and 0, respectively and sometimes -1 and 0.

Luckily this is actually available on the event, `event.index`, so switched to that.

calbach: Oops - I was looking at `MatTab` rather than `MatTabChangeEvent`. Nice, thanks!

PR #241, opened 2018-03-21
Drop the header from job-details, in favor of an "X"

calbach:


ui/src/app/job-details/panels/panels.component.html
```diff
@@ -1,13 +1,17 @@
...
-      <span *ngFor="let l of labels">
-        <p class="text-info">{{ l + ': ' + job.labels[l] }}</p>
-      </span>
-      <span *ngFor="let f of displayedExtensions">
-        <p class="text-info">{{ f + ': ' + job.extensions[f] }}</p>
-      </span>
+    <p class="header">{{ job.name }}</p>
+    <button mat-button class="close"
+        (click)="handleClose()">
+      &#x274C;
```
bfcrampton: neat

calbach: This actually rendered as a giant red X on my pixelbook for some reason. Went with the multiplication symbol instead which renders well on both of my chromes.

PR #243, opened 2018-03-24
[UI] Don't show input/output tabs when there are none

bfcrampton:
Also add the log file tabs once we download them and verify they exist.

### DIFF
--- a/ui/src/app/job-details/resources/resources.component.ts
+++ b/ui/src/app/job-details/resources/resources.component.ts
@@ -64,7 +64,7 @@ export class JobResourcesComponent implements OnInit {
           .then(entries => {
             // Sort log files by the file name (tuples will be converted to
             // string and compared).
-            for (let [file, data] of entries.filter(l => !!l).sort()) {
+            for (let [file, data] of entries.filter(e => !!e).sort()) {
               if (data) {
                 this.logFileData.set(file, data);
                 this.tabIds.push('log-' + file);
--- a/ui/src/app/job-details/resources/resources.component.ts
+++ b/ui/src/app/job-details/resources/resources.component.ts
@@ -59,17 +59,19 @@ export class JobResourcesComponent implements OnInit {
       }
 
       if (this.job.extensions.logs) {
-        let files = Object.keys(this.job.extensions.logs);
+        const files = Object.keys(this.job.extensions.logs);
         Promise.all(files.map(file => this.readResourceFile(file)))
           .then(entries => {
-            for (let [file, data] of entries.sort()) {
+            // Sort log files by the file name (tuples will be converted to
+            // string and compared).
+            for (let [file, data] of entries.filter(l => !!l).sort()) {
               if (data) {
                 this.logFileData.set(file, data);
                 this.tabIds.push('log-' + file);
                 this.tabTitles.set('log-' + file, file);
               }
             }
-          }).catch(error => this.handleError(error));
+          });
       }
     }
 
@@ -110,10 +112,11 @@ export class JobResourcesComponent implements OnInit {
     let bucket = ResourceUtils.getResourceBucket(this.job.extensions.logs[file]);
     let object = ResourceUtils.getResourceObject(this.job.extensions.logs[file]);
     return this.gcsService.readObject(bucket, object)
-      .then(data => [file, data] as [string, string]);
+      .then(data => [file, data] as [string, string])
+      .catch(error => this.handleError(error));
   }
 
-  private handleError(error: any) {
+  private handleError(error: any): any {
     this.errorBar.open(
       new ErrorMessageFormatterPipe().transform(error),
       'Dismiss',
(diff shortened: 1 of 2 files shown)


ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -43,28 +43,23 @@ export class JobResourcesComponent implements OnInit {
...
     }
-    if (this.job.outputs) {
+    if (this.job.outputs && Object.keys(this.job.outputs).length > 0) {
       this.outputs = Object.keys(this.job.outputs).sort();
       this.tabIds.push('outputs');
     }
 
     if (this.job.extensions) {
-      if (this.job.extensions.sourceFile) {
+      if (this.job.extensions.sourceFile && this.job.extensions.sourceFile.length > 0) {
```
calbach: This addition doesn't look necessary me; empty string is falsy in JS -  unless typescript is doing something I'm not aware of.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -43,28 +43,23 @@ export class JobResourcesComponent implements OnInit {
...
       }
 
       if (this.job.extensions.logs) {
-        for (let file of Object.keys(this.job.extensions.logs)) {
-          this.readResourceFile(file);
-          let tabId = 'log-' + file;
-          this.tabIds.push(tabId);
-          this.tabTitles.set(tabId, file);
-        }
+        Object.keys(this.job.extensions.logs).forEach(file => this.readResourceFile(file))
```
calbach: I would rework this slightly. The issue is that the log files could load in in an arbitrary order. Using a a `Promise.all()` on the `readResourceFile` promises, then pushing tabIds in a for loop in a single callback would resolve this.

bfcrampton: Good point, switched to that approach.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -59,12 +59,17 @@ export class JobResourcesComponent implements OnInit {
       }
 
       if (this.job.extensions.logs) {
-        for (let file of Object.keys(this.job.extensions.logs)) {
-          this.readResourceFile(file);
-          let tabId = 'log-' + file;
-          this.tabIds.push(tabId);
-          this.tabTitles.set(tabId, file);
-        }
+        let files = Object.keys(this.job.extensions.logs);
```
calbach: nit: const

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -59,12 +59,17 @@ export class JobResourcesComponent implements OnInit {
...
+        Promise.all(files.map(file => this.readResourceFile(file)))
```
calbach: Actually, rereading the `Promise.all` docs, you don't exactly want this as it will short-circuit reject. You'd need to catch immediately after `this.readResourceFile`, `handleError` there, then `Promise.all` on the return value there. You'd then need to filter out "empty" results below.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -59,12 +59,17 @@ export class JobResourcesComponent implements OnInit {
...
+          .then(entries => {
+            for (let [file, data] of entries.sort()) {
```
calbach: How do string arrays get sorted? I think you'd probably want to just sort on key names.

ui/src/app/job-details/resources/resources.component.ts
bfcrampton: Done.

ui/src/app/job-details/resources/resources.component.ts
bfcrampton: Done.

ui/src/app/job-details/resources/resources.component.ts
bfcrampton: Per offline discussion, it will `toString` the tuples and sort that in ascending, ASCII character order.

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -59,12 +59,19 @@ export class JobResourcesComponent implements OnInit {
...
-          let tabId = 'log-' + file;
-          this.tabIds.push(tabId);
-          this.tabTitles.set(tabId, file);
-        }
+        const files = Object.keys(this.job.extensions.logs);
+        Promise.all(files.map(file => this.readResourceFile(file)))
+          .then(entries => {
+            // Sort log files by the file name (tuples will be converted to
+            // string and compared).
+            for (let [file, data] of entries.filter(l => !!l).sort()) {
```
calbach: nit: not a fan of `l` as a var name, would prefer `e` here, or anything else.

bfcrampton: Right good call, fixed.

PR #245, opened 2018-03-26
[UI] Auto select top project on drop down

bfcrampton:
Also add sorting to the projects list. This is necessary in the case where you have two projects with one name being the prefix of the other (ex: `bvdp-jmui-testing` and `bvdp-jmui-testing-2`). We'd like the shorter to be first so that if you type exactly `bvdp-jmui-testing` the exact match is auto-selected.

### DIFF
--- a/ui/src/app/projects/projects.component.ts
+++ b/ui/src/app/projects/projects.component.ts
@@ -78,7 +78,9 @@ export class ProjectsComponent implements OnInit {
     this.projectsObservable = Observable.fromPromise(
       this.projectsService.listProjects(filter)
         .then(listView => {
-          this.projects = listView.results.sort() ? listView.results : [];
+          // Sort projects alphabetically so that the shorter, matching project
+          // shows up first.
+          this.projects = (listView ? listView.results : []).sort();
           // If the currently entered string is a valid project ID, hide the
           // autocomplete menu so that the user can click the button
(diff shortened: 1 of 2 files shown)


ui/src/app/projects/projects.component.ts
```diff
@@ -79,13 +78,14 @@ export class ProjectsComponent implements OnInit {
     this.projectsObservable = Observable.fromPromise(
       this.projectsService.listProjects(filter)
         .then(listView => {
-          this.projects = listView.results ? listView.results : [];
+          this.projects = listView.results.sort() ? listView.results : [];
```
calbach: FYI: I can't tell what order CRM is returning these in by default, but it's definitely not ideal since I found the prefix case: `google.com:verily` search does not provide itself as the first result, but instead some other longer project is given.

calbach: nit: I realize it's equivalent but I'd probably split out the `sort()` onto a different line, or else put it on the latter part of this ternary expression. Putting it in the predicate makes the intent a little bit more ambiguous IMO.

bfcrampton: Acknowledged, yeah I couldn't tell either but sorting seems to give us a reasonable order.

bfcrampton: Moved outside ternary expression and added comment about sorting.

PR #246, opened 2018-03-26
[UI] Fix jobs table links ctrl+click new tab

bfcrampton:
Also fix link colors on job details page to be black.

### DIFF
--- a/ui/src/app/job-details/resources/resources-table/resources-table.component.css
+++ b/ui/src/app/job-details/resources/resources-table/resources-table.component.css
@@ -1,3 +1,0 @@
-.resource-button {
-  color: black;
-}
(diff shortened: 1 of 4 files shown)


ui/src/app/job-details/resources/resources-table/resources-table.component.css
```diff
@@ -0,0 +1,3 @@
+.resource-button {
+  color: black;
```
calbach: Background on this? I tried this change in Chrome inspector and I think I prefer having standard link styling.

bfcrampton: Switched back to standard.

PR #247, opened 2018-03-29
Fix buggy multi-abort error handling

calbach:


ui/src/app/job-list/table/table.component.ts
```diff
@@ -140,17 +140,33 @@ export class JobsTableComponent implements OnInit {
   onAbortJobs(): void {
     const aborts: Promise<void>[] = [];
     const selected = this.selection.selected.slice();
+
+    let numErrs = 0;
+    const ref = this.errorBar.open(
```
bfcrampton: nit: maybe rename from error bar since its not only for errors now?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -140,17 +140,33 @@ export class JobsTableComponent implements OnInit {
...
+      `Aborting jobs...`, /* action */ '',
```
bfcrampton: nit: Any reason why tickmarks vs. single quotes?

ui/src/app/core/job-manager.service.ts
```diff
@@ -57,17 +57,25 @@ export class JobManagerService {
...
+    try {
+      return response.json();
+    } catch (_) {
+      return {};
+    }
+  }
+
+  private getErrorTitle(response: Response): string {
+    const json = this.respToJson(response);
+    if (json["title"]) {
```
bfcrampton: Won't this throw a key error? I think you want if `if "title" in json`

calbach: You're thinking of Python. JS will return `undefined` in this case.

ui/src/app/job-list/table/table.component.ts
calbach: Good idea, done.

ui/src/app/job-list/table/table.component.ts
calbach: Thanks; just copy-paste (back-ticks needed for interpolation elsewhere). Fixed.

ui/src/app/core/job-manager.service.ts
bfcrampton: Yes I am sorry got confused 😕

PR #248, opened 2018-03-31
Open and focus chips on add

alahwa:
When new chips are added, open their menus and (if free text) focus the text box by default. This change was much larger than expected for a feature of this size.

Context: Binding to the DOM and trying to manually trigger menu open, close, and focus events was causing buggy behavior primarily due to MatMenus being unable to handle having multiple MatMenuTriggers (one per chip). Fixing this behavior required restructuring the Header, splitting out chips into components, each handling their own MatMenu and associated Trigger. Functionality was further subdivided by chip type.

Because MatChips are nested in FilterChip components, we nest those in another layer of MatChips to prevent losing all of the functionality of having MatChips in a MatChipList.


ui/src/app/shared/header/chips/status-selection.component.ts
```diff
@@ -0,0 +1,32 @@
+import {JobStatus} from "../../model/JobStatus";
+import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
+
+@Component({
+  selector: 'jm-status-selection',
+  templateUrl: './status-selection.component.html',
+})
+export class StatusSelectionComponent implements OnInit {
+  @Input() chipKey: string;
```
calbach: chipkey is unused here , rm?

ui/src/app/shared/header/chips/filter-chip.component.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  ngOnInit(): void {
+    this.options = URLSearchParamsUtils.getQueryFields(
+      this.capabilitiesService.getCapabilitiesSynchronous());
+    this.currentChipValue = this.initialChipValue;
+  }
+
+  focusInput() {
+    if (this.freeTextInput) {
+      setTimeout(() => this.freeTextInput.nativeElement.focus(), 100);
```
calbach: The `100` isn't ideal, does 0 not work? If it's needed, could you add a comment here explaining?

ui/src/app/shared/header/chips/filter-chip.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  getDisplayValue() {
+    return this.chipKey + ': ' + this.currentChipValue;
+  }
+
+  removeThisChip(): void {
+    this.removeChip.emit(null);
+  }
+
+  setChipValue(value: string): void {
+    if (value != "") {
```
calbach: nit: `if (value) {` would be simpler

ui/src/app/shared/header/header.component.ts
```diff
@@ -1,61 +1,58 @@
...
 })
-export class HeaderComponent implements OnInit, AfterViewInit, OnDestroy {
+export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked, OnDestroy {
   @Input() jobs: BehaviorSubject<JobListView>;
   @Input() pageSize: number;
+  @ViewChildren(FilterChipComponent) chipElements: QueryList<FilterChipComponent>;
+  @ViewChild(MatAutocompleteTrigger) autocompleteTrigger: MatAutocompleteTrigger;
   @ViewChild(MatPaginator) paginator: MatPaginator;
   public pageSubject: Subject<PageEvent> = new Subject<PageEvent>();
+  private chipToExpand: string;
```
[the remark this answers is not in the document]
alahwa: Done

ui/src/app/shared/header/header.component.ts
```diff
@@ -218,9 +175,10 @@ export class HeaderComponent implements OnInit, AfterViewInit, OnDestroy {
 
   removeChip(chipKey: string): void {
     this.chips.delete(chipKey);
-    if (chipKey == 'statuses') {
-      this.selectedStatuses = [];
-    }
+    this.search();
+  }
```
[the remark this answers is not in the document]
alahwa: Done

ui/src/app/shared/header/chips/filter-chip.component.ts
alahwa: Oops, this was a leftover workaround from an earlier bug with multiple triggers. Removed.

ui/src/app/shared/header/chips/status-selection.component.ts
```diff
@@ -0,0 +1,32 @@
+import {JobStatus} from "../../model/JobStatus";
+import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
+
+@Component({
+  selector: 'jm-status-selection',
+  templateUrl: './status-selection.component.html',
+})
+export class StatusSelectionComponent implements OnInit {
```
calbach: When I enter a new "statuses" chip directly, looks like it's getting initialized with null or undefined; the API call then fails when I try to enter it.

I wasn't planning to do this during the demo anyways, but something to fix for this PR.

alahwa: Good catch, fixed

ui/src/app/shared/header/chips/datepicker-input.component.ts
```diff
@@ -0,0 +1,32 @@
+import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
```
calbach: Doesn't have to be in this PR, but one nice future improvement might be to auto-open the datepicker when you open the mat menu.

ui/src/app/shared/header/header.component.css
```diff
@@ -26,3 +20,10 @@
   justify-content: space-between;
   align-items: center;
 }
+
+.chipShell {
+  width: fit-content !important;
```
calbach: I'm guessing the `!important` is needed, but would be great if it weren't. Otherwise maybe add a comment here to explain why it is needed.

ui/src/app/shared/header/chips/datepicker-input.component.ts
alahwa: Done. I think this is a more legitimate use of setTimeout since we're waiting on a timed UI animation rather than for lifecycle code to execute, and the calendar doesn't position properly before the dropdown is finished redrawing for the last time. I'm on the fence about this, so if you think this is messy, we can remove the timeout and keep the behavior as is (calendar opens immediately before card expansion animation is complete and overlays the card instead of positioning below it).

ui/src/app/shared/header/header.component.css
alahwa: Actually only the `padding = 0` is necessary, removed

ui/src/app/shared/header/chips/datepicker-input.component.ts
alahwa: I tested this using menuOpened() but that callback seems to also trigger before the animation completes on the menu.

PR #249, opened 2018-04-02
[Cromwell] Add capabilities loader for Cromwell shim layer

rexwangcc:
This PR enables Cromwell shim layer to load capabilities config from a JSON file and override pre-defined capabilities configurations. This aims at simplifying the deployment process and complying with the [The Twelve-Factor App](https://12factor.net/config) which encourages separating config files from source code.


servers/cromwell/jobs/util.py
```diff
@@ -144,3 +144,26 @@ def _deserialize_dict(data, boxed_type):
...
+    target_keys = set(target_keys) if target_keys else set(dic.keys())
+    res_dic = {}
+
+    for key, val in dic.items():
+        if isinstance(val, dict) and key in target_keys:
+            for inner_key, inner_val in val.items():
+                res_dic['{0}{1}{2}'.format(key, sep, inner_key)] = inner_val
+        else:
+            res_dic[key] = val
+    return res_dic
```
rexwangcc: This function is relatively expensive, but since the structure of the `capabilities_config.json` is relatively simple, and the computation occurs in starting process instead of controllers, I think the performance won't be affected that much.

servers/cromwell/jobs/controllers/capabilities_controller.py
```diff
@@ -10,26 +10,38 @@ def get_capabilities():
...
+    }
+
+    capabilities_config = current_app.config['capabilities_config']
+    if capabilities_config:
+        capabilities.update(capabilities_config)
+
+    display_fields, common_labels, query_extensions = [], set([]), set([])
+
+    # Construct display_fields
+    for field, display in capabilities['display_fields'].items():
```
rexwangcc: This loop provides the flexibility to add any display_field to the UI, but sacrifices the ability to control the order of columns. In order to control the order of the columns, we have three ways to do it:

1. Make changes to the UI so users can modify the order of columns in their browsers.
2. Control part of the orders(status and submission) but keep the flexibility:
```python
display_fields = [capabilities['display_fields'].pop('status'), capabilities['display_fields'].pop('submission')]

for field, display in capabilities['display_fields'].items():
        display_fields.append(DisplayField(field=field, display=display))
```
3. Control all of the orders:
Change the structure of ` capabilities_config.json` to:
```json
{
      "display_fields": {
        "status": "Status",
        "submission": "Submitted",
        "labels": [
          {"cromwell-workflow-id": "Workflow ID"},
          {"workflow-name": "Workflow Name"},
          {"comment": "Comment"}
        ]
      },
      "common_labels": [
        "cromwell-workflow-id",
        "workflow-name",
        "comment"
      ],
      "query_extensions": []
    }
```
and modify the `flatten` function(much more expensive) to:
```python
def flatten(dic, nest_keys, sep='.'):
    nest_keys = set(nest_keys)
    res_dic ={}
    for key, val in dic.items():
        if key in nest_keys and isinstance(val, list):
            res_dic[key] = []
            for inner_dic in val:
                for inner_key, inner_val in inner_dic.items():
                    res_dic[key].append({key + sep + inner_key:inner_val})
        else:
            res_dic[key] = val
    return res_dic
```
and then change here to:
```python
for dics in flatten(dic['display_fields'], nest_keys=['labels'])['labels']:
    display_fields.append(DisplayField(field=list(dics.keys())[0], display=list(dics.values())[0]))
```

rexwangcc: Among all of the solutions, I personally prefer the current one for the most.

servers/cromwell/jobs/util.py
samanehsan: Should this be able to handle the case where the value is a list of dictionaries?

rexwangcc: @samanehsan thanks, I will definitely add test cases for this function once we decide which way to go. Currently I'm hesitating between different implementations here, see my comments on the `capabilities_controller.py`, whether or not to handle the `list` cases is also depends on the structure we want to have for the `capabilities_config.json`.

servers/cromwell/jobs/controllers/capabilities_controller.py
```diff
@@ -10,26 +10,36 @@ def get_capabilities():
...
+        display_fields.append(DisplayField(field=field, display=display))
+
+    # Construct common_labels, which is used by the query builder
+    common_labels.update(capabilities['common_labels'])
```
samanehsan: Instead of setting `common_labels` to an empty set and updating its value, could we do `common_labels = set(capabilities['common_labels'])`? And also with `query_extensions` below?

servers/cromwell/jobs/controllers/capabilities_controller.py
samanehsan: I like your third suggestion, since it would allow us to set a default order for the columns.

servers/cromwell/jobs/util.py
samanehsan: Oh, okay! I commented on that as well above

servers/cromwell/jobs/__main__.py
```diff
@@ -47,18 +48,33 @@
...
+# Try to load the capabilities config file
+try:
+    with open(capabilities_path) as f:
+        capabilities_config = json.load(f)
+    capabilities_config['display_fields'] = flatten_dict(capabilities_config['display_fields'], target_keys=['labels'])
+except (IOError, TypeError, KeyError):
+    logger.warning(
+        'Failed to load capabilities config, using default display fields.')
+    capabilities_config = None
+finally:
```
bfcrampton: nit: Do you need the `finally` here? Since you're catching the exception and not re-throwing I think it should just continue.

rexwangcc:
@bfcrampton Your comment makes 100% sense to me. And it is true that I was completely rebuilding the wheels (👎 :( ) since I somehow didn't realize the generated models come with deserialize methods. Just re-implemented the controller based on your suggestion.

servers/cromwell/jobs/__main__.py
```diff
@@ -47,18 +48,33 @@
...
+try:
+    with open(capabilities_path) as f:
+        capabilities_config = json.load(f)
+    logger.info('Successfully loaded the custom capabilities config.')
+    app.app.config['capabilities'] = CapabilitiesResponse.from_dict(
+        capabilities_config)
+except IOError:
+    logger.warning(
+        'Failed to load capabilities config, using default display fields.')
+    app.app.config['capabilities'] = None
```
bfcrampton: nit: I wouldn't explicitly set this to None, just leave it unset.

servers/cromwell/jobs/controllers/capabilities_controller.py
```diff
@@ -10,26 +9,19 @@ def get_capabilities():
     Returns:
         CapabilitiesResponse: Response containing this backend's capabilities
     """
+    # Check if a capabilities config is given
+    capabilities = current_app.config['capabilities']
```
bfcrampton: And switch this to:
```
if 'capabilities' in current_app.config:
    return capabilities['capabilities']
```

rexwangcc:
Thanks for all the review and comments!!!

PR #250, opened 2018-04-03
Document how to run workflows in Cromwell

samanehsan:
This PR includes: 
- Updates to the Cromwell README with the command for starting jobs in Cromwell 
- A test workflow that has a subworkflow and all of the other files needed to run the job in Cromwell (thanks @rexwangcc!). I can also take this out of the repo if there's a better place to put these files.


servers/cromwell/jobs/test/test_workflows/stress_test/inputs.json
```diff
@@ -0,0 +1,21 @@
...
+  "StressTestAdapterSmartSeq2SingleCell.genome_ref_fasta": "gs://hca-dcp-mint-test-data/reference/GRCh38_Gencode/GRCh38.primary_assembly.genome.fa",
+  "StressTestAdapterSmartSeq2SingleCell.hisat2_ref_trans_index": "gs://hca-dcp-mint-test-data/reference/HISAT2/gencode_v27_trans_rsem.tar.gz",
+  "StressTestAdapterSmartSeq2SingleCell.rsem_ref_index": "gs://hca-dcp-mint-test-data/reference/GRCh38_Gencode/gencode_v27_primary.tar",
+  "StressTestAdapterSmartSeq2SingleCell.gene_ref_flat": "gs://hca-dcp-mint-test-data/reference/GRCh38_Gencode/GRCh38_gencode.v27.refFlat.txt",
+  "StressTestAdapterSmartSeq2SingleCell.hisat2_ref_name": "genome_snp_tran",
+  "StressTestAdapterSmartSeq2SingleCell.stranded":"NONE",
+  "StressTestAdapterSmartSeq2SingleCell.reference_bundle": "bf51d668-3e14-4843-9bc7-5d676fdf0e01",
+  "StressTestAdapterSmartSeq2SingleCell.format_map": "gs://hca-dcp-mint-test-data/adapters/file_format_map.json",
+  "StressTestAdapterSmartSeq2SingleCell.dss_url": "https://dss.dev.data.humancellatlas.org/v1",
+  "StressTestAdapterSmartSeq2SingleCell.submit_url": "http://api.ingest.dev.data.humancellatlas.org/",
```
rexwangcc: How about replacing all the actual urls in this file with fake url strings? Although there are no credentials, we may want to limit the user groups to our public bucket? (Even just for reducing egress cost lol)

servers/cromwell/jobs/test/test_workflows/stress_test/labels.json
```diff
@@ -0,0 +1,4 @@
+{
+    "comment": "stress-test",
```
rexwangcc: I'd replace the "stress-test" with "workflow-test"

servers/cromwell/jobs/test/test_workflows/stress_test/inputs.json
samanehsan: That's a good point! I think it would be useful to keep at least one url in case we wanted to test the behavior of the link in the inputs list on the details page. What do you think?

rexwangcc: You are right, it is better to have one realistic url case.

servers/cromwell/README.md
```diff
@@ -3,23 +3,37 @@
...
-```
-export CROMWELL_URL=https://cromwell.test-cromwell.broadinstitute.org/api/workflows/v1
-```
 
-**Note:** If you want to setup this API Server against a locally hosted Cromwell instance, you need to explicitly provide the ip address (inet if the Cromwell is hosted on the same machine) to the Cromwell with port numbers, for example:
-```
-export CROMWELL_URL=http://192.168.0.106:8000/api/workflows/v1
-```
+    ```
+    export CROMWELL_URL=https://cromwell.test-cromwell.broadinstitute.org/api/workflows/v1
```
calbach: Sorry - which one should I have access to? @dshiga gave us access to the "Verily" instance, which I appear to be able to access via the browser. However, using the config.json my server just fails:

```
HTTPError: 405 Client Error: Method Not Allowed for url: https://cromwell-verily.dsp-techops.broadinstitute.org/api/v1/workflows/query
```

servers/cromwell/README.md
```diff
@@ -3,23 +3,37 @@
...
+    {
+      "cromwell_user" : "USERNAME",
+      "cromwell_password" : "PASSWORD"
+    }
+    ```
+
+The Job Manager does not currently support launching jobs. Cromwell jobs can be launched by sending a `POST` request to the `/api/workflows/{version}` endpoint. For example:
+```
+cd servers/cromwell/jobs/test/test_workflow
+curl -X POST "https://cromwell.test-cromwell.broadinstitute.org/api/workflows/v1" \
```
calbach: nit: replace with `"${CROMWELL_URL}"` to make this more general

servers/cromwell/README.md
```diff
@@ -3,23 +3,37 @@
...
-  "cromwell_user" : "USERNAME",
-  "cromwell_password" : "PASSWORD"
-}
+    ```
+    {
+      "cromwell_user" : "USERNAME",
+      "cromwell_password" : "PASSWORD"
+    }
+    ```
+
```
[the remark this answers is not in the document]
samanehsan: The docker-compose step is documented in the main [README](https://github.com/DataBiosphere/job-manager/blob/master/README.md#prerequisite). But that's a good point and I think it makes more sense to add it to this README. 

I think the dev and staging configs were different from when the environment files contained a link to the api url for the timing diagram. Good catch! Is it okay if we consolidate those in a separate ticket?

servers/cromwell/README.md
samanehsan: As discussed over slack, the issue here was that the CROMELL_URL needs to be `https://cromwell-verily.dsp-techops.broadinstitute.org/api/workflows/v1`

servers/cromwell/README.md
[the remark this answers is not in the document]
calbach: Yes, definitely fine as a follow-up - sorry I forgot to respond on this.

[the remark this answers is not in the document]
samanehsan: Ok great! I created a ticket for it as a reminder

samanehsan:
Thanks for the suggestions @calbach! Let me know if this looks good to you!

PR #254, opened 2018-04-11
fixed issue with cromwell swagger codegen

rsasch:
Needed to add line to cromwell/.swagger-codegen-ignore to cover it

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ see [servers/cromwell](servers/cromwell/README.md#Development).
 #### Notes
 1. Websocket reload on code change does not work in docker-compose (see
 https://github.com/angular/angular-cli/issues/6349).
-1. Changes to `package.json` or `requirements.txt` require a rebuild with:
+1. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
   ```
   docker-compose up --build
   ```
--- a/README.md
+++ b/README.md
@@ -67,12 +67,12 @@ see [servers/cromwell](servers/cromwell/README.md#Development).
 
 ### Run Locally
 1. Run `docker-compose up` from the root of the repository:
-1. Navigate to http://localhost:4200.
+2. Navigate to http://localhost:4200.
 
 #### Notes
 1. Websocket reload on code change does not work in docker-compose (see
 https://github.com/angular/angular-cli/issues/6349).
-1. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
+2. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
   ```
   docker-compose up --build
   ```
@@ -93,13 +93,13 @@ update the server implementations:
     # macOS
     brew install swagger-codegen
     ```
-1. Clear out existing generated models:
+2. Clear out existing generated models:
     ```
     rm ui/src/app/shared/model/*
     rm servers/dsub/jobs/models/*
     rm servers/cromwell/jobs/models/*
     ```
-1. Regenerate both the python and angular definitions.
+3. Regenerate both the python and angular definitions.
     ```
     java -jar swagger-codegen-cli.jar generate \
       -i api/jobs.yaml \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -72,7 +72,7 @@ see [servers/cromwell](servers/cromwell/README.md#Development).
 #### Notes
 1. Websocket reload on code change does not work in docker-compose (see
 https://github.com/angular/angular-cli/issues/6349).
-1. Changes to `package.json` or `requirements.txt` require a rebuild with:
+1. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
```
rexwangcc: nit: `1.` -> `2.` although they are the same thing in rendered markdown file.

rsasch: thanks for pointing that out -- fixed

calbach: FYI: This was intentional, it's to avoid having to shift every number in the list if you add a new bullet.

README.md
```diff
@@ -67,12 +67,12 @@ see [servers/cromwell](servers/cromwell/README.md#Development).
...
 ### Run Locally
 1. Run `docker-compose up` from the root of the repository:
-1. Navigate to http://localhost:4200.
+2. Navigate to http://localhost:4200.
 
 #### Notes
 1. Websocket reload on code change does not work in docker-compose (see
 https://github.com/angular/angular-cli/issues/6349).
-1. Changes to `package.json` or `requirements.txt` require a rebuild with:
+2. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
```
calbach: This actually shouldn't be true to the best of my knowledge; can you confirm that you needed to do this? What happened when you didn't?

README.md
rsasch: I was reacting to a comment by @rexwangcc -- I have no strong opinion either way.

README.md
rsasch: I did – without regenerating the API (I was only testing my dev environment while pointed at a Cromwell instance), I would get an error when I tried to stand up my dev env. Once it was suggested that I regenerated the API (and I did so), the issue was resolved, so I wanted the documentation to reflect that.
I think the error I saw was `Failed to find application: 'jobs`', but I will confirm later.

README.md
rexwangcc: OK, so I didn't know it was an intentional cookie :). I personally prefer to outline the ordered lists use ordered numbers here. Although all of them will eventually be rendered to "<li></li>" on web pages, having ordered numbers helps me manage the content in vim/sublime. 

I won't be unhappy if we switch back to use all `1.`s though(we need to make sure the first item uses `1.` based on Markdown's spec)

README.md
calbach: I see - so you recently added post-codegen modifications to __init__.py; this change makes sense then (also probably was not obvious this file was generated).

This is fine. However, we don't do this in the dsub server. I'm curious about the reason for deviating from what we did there (how we invoke gunicorn):
https://github.com/DataBiosphere/job-manager/blob/master/servers/dsub/Dockerfile#L19

RE: Needing to rebuild the servers: after code regeneration, worst case you should need to ctrl+c, then re-up the servers. All source code is mounted into docker via volumes, so any code change in our codebase should not require rebuilding the image. If that changed at some point for the Cromwell shim (I don't think it has), we should fix it.

calbach:
Also, thanks for adding documentation in response to the issues you hit, that's great. Just trying to understand the issues you were hitting here to see if there's anything that can be improved/simplified more generally.

README.md
rexwangcc: The reason for deviating from `ENTRYPOINT ["/env/bin/gunicorn", "jobs.__main__:app"]` to `ENTRYPOINT ["gunicorn", "jobs:run()"]` is that, the formar approach will tell the gunicorn that `app` is in the` __main__.py`, so gunicorn can always grab that "app", even if we don't want it to start if HTTPerror happens(https://github.com/DataBiosphere/job-manager/blob/37ff27fceb125f4097681c29b0bbfa1594d988c0/servers/cromwell/jobs/__main__.py#L93). Separating the app from`__main__.py` to `__init__.py` can enable the pre-start check and prevent gunicorn from starting if HTTPErrors happen. At least that is the best approach we came up with at that time (https://github.com/DataBiosphere/job-manager/pull/224#discussion_r170754300)

However, it would be really appreciated if we could find a better way for it :)

calbach: Ah, thanks for the link - makes sense. This seems like a reasonable solution to me.

I'd guess you could avoid this __init__ import via `jobs.__main__:run()` with the same effect, but it doesn't really matter.

rexwangcc: `jobs.__main__:run()` seems better to me. I can't remember if I have tried this before, but will definitely try to switch to it in next PR if it works. Thanks!

PR #257, opened 2018-04-17
Fix table width sizing and long field truncation

bfcrampton:
Make the job list table horizontally scrollable within the page. Fixes the bug where the query builder is cut off when there are too many columns in the table (overflow list tables). Also fixes the bug where the checkbox and job name appeared stacked on top of each other.

This is a precursor to pin selected-jobs controls to window bottom which I will follow up with in a separate PR.

### DIFF
--- a/ui/src/app/job-list/job-list.component.css
+++ b/ui/src/app/job-list/job-list.component.css
@@ -2,7 +2,6 @@
   position: relative;
   overflow: scroll;
   width: 100%;
-  height: 100%;
 }
 
 .spinner-container {
--- a/ui/src/app/job-list/job-list.component.css
+++ b/ui/src/app/job-list/job-list.component.css
@@ -1,5 +1,8 @@
 .table-container {
   position: relative;
+  overflow: scroll;
+  width: 100%;
+  height: 100%;
 }
 
 .spinner-container {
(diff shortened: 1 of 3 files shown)


ui/src/app/job-list/job-list.component.css
```diff
@@ -1,5 +1,8 @@
 .table-container {
   position: relative;
+  overflow: scroll;
+  width: 100%;
+  height: 100%;
```
calbach: Does this change add vertical scrolling to the job list as well?

bfcrampton: Nope no-op change, removed.

rexwangcc:
Thanks for this fix! A quick question:
> Also fixes the bug where the checkbox and job name appeared stacked on top of each other.
What does it mean?

bfcrampton:
This issue which used to occur when the window is too narrow.

screen shot 2018-04-16 at 6 10 14 pm

rexwangcc:
@bfcrampton Oh, I see! I tried out the latest master, but it seems this "stack problem" is still there? 
screen shot 2018-04-16 at 9 12 42 pm

bfcrampton:
Hmm okay seems like it didn't fix the issue for longer names for some reason, I'll look into it more. Thanks for letting me know!

PR #260, opened 2018-04-17
[UI] Fix nested scroll bars

bfcrampton:
Didn't notice this on my mac but on chrome on linux there were two vertical scroll bars displayed on the page.


bfcrampton:
Rolled this fix into https://github.com/DataBiosphere/job-manager/pull/259

PR #263, opened 2018-04-26
[Bug] Fix the issue that filtering on labels doesn't work for Cromwell.

rexwangcc:
This PR fixes the bug that filtering on labels is not working for Cromwell. 

The reason is that we didn't implement that functionality in Cromwell's controllers before.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -256,6 +256,10 @@ def cromwell_query_params(query, page, page_size):
     if query.statuses:
         statuses = [{'status': s} for s in set(query.statuses)]
         query_params.extend(statuses)
+    if query.labels:
+        labels = [{'label': k + ':' + v} for k, v in query.labels.items()]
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -256,6 +256,10 @@ def cromwell_query_params(query, page, page_size):
     if query.statuses:
         statuses = [{'status': s} for s in set(query.statuses)]
         query_params.extend(statuses)
+    if query.labels:
+        labels = [{'label': k + ':' + v} for k, v in query.labels.items()]
```
rexwangcc: I'm not sure if it will change in the future, but currently, Cromwell doesn't allow non-string key/value for labels, so I didn't add `str(k)` and `str(v)` here.

rsasch:
This looks much better.  I am seeing rows returned that aren't a match (it seems to be searching all labels, not just the one specified), but I'm also seeing that in the Cromwell swagger.

PR #264, opened 2018-04-27
Prevent search on discard of empty chip

rsasch:
see https://elastc.com/c/46LPfbYI/291-don-t-trigger-search-on-discard-of-empty-chip

### DIFF
--- a/ui/src/app/shared/header/header.component.ts
+++ b/ui/src/app/shared/header/header.component.ts
@@ -174,8 +174,13 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
   }
 
   removeChip(chipKey: string): void {
-    this.chips.delete(chipKey);
-    this.search();
+    if (this.chips.get(chipKey).valueOf()) {
+      this.chips.delete(chipKey);
+      this.search();
+    }
+    else {
+      this.chips.delete(chipKey);
+    }
   }
 
   updateValue(chipKey: string, chipValue: string): void {
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -256,6 +256,10 @@ def cromwell_query_params(query, page, page_size):
     if query.statuses:
         statuses = [{'status': s} for s in set(query.statuses)]
         query_params.extend(statuses)
+    if query.labels:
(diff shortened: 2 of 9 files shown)


ui/src/app/shared/header/header.component.ts
```diff
@@ -174,8 +174,13 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
...
 
   removeChip(chipKey: string): void {
-    this.chips.delete(chipKey);
-    this.search();
+    if (this.chips.get(chipKey).valueOf()) {
+      this.chips.delete(chipKey);
+      this.search();
+    }
+    else {
+      this.chips.delete(chipKey);
```
bfcrampton: nit: move out of the else:
```
this.chips.delete(chipKey);
if (this.chips.get(chipKey).valueOf()) {
  this.search();
}
```

rsasch: If the chip has been deleted, `this.chips.get(chipKey)` returns an error; I can assign it to a variable and then use that in the if statement, if you think that would be cleaner.

ui/src/app/shared/header/header.component.ts
```diff
@@ -174,8 +174,13 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
   }
 
   removeChip(chipKey: string): void {
-    this.chips.delete(chipKey);
-    this.search();
+    if (this.chips.get(chipKey).valueOf()) {
+      this.chips.delete(chipKey);
+      this.search();
+    }
```
calbach: `} else {`

The rest of the codebase uses this convention.

ui/src/app/shared/header/header.component.ts
```diff
@@ -174,8 +174,13 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
   }
 
   removeChip(chipKey: string): void {
-    this.chips.delete(chipKey);
-    this.search();
+    if (this.chips.get(chipKey).valueOf()) {
+      this.chips.delete(chipKey);
+      this.search();
```
calbach: Per the other ticket about "?q=", this special-casing shouldn't be necessary if we omit q= on an unfiltered query. I would just fix that other issue and leave this part of the code unchanged.

PR #265, opened 2018-04-27
Job list: drop the "q" param entirely when there are 0 filters

rsasch:
check for query and query string before searching
see https://elastc.com/c/jPo5zc1X/287-job-list-drop-the-q-param-entirely-when-there-are-0-filters

### DIFF
--- a/ui/src/app/shared/header/header.component.ts
+++ b/ui/src/app/shared/header/header.component.ts
@@ -192,10 +192,16 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
       }
     });
     let query: string = URLSearchParamsUtils.encodeURLSearchParamsFromMap(paramMap);
-    this.router.navigate(
-      ['jobs'],
-      {queryParams: { q: query}}
-    );
+    if (query) {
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/header/header.component.ts
```diff
@@ -192,10 +192,16 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
...
-    this.router.navigate(
-      ['jobs'],
-      {queryParams: { q: query}}
-    );
+    if (query) {
+      this.router.navigate(
+        ['jobs'],
+        {queryParams: { q: query}}
+      );
+    } else if (this.route.snapshot.queryParams['q']) {
```
calbach: nit: I might just depend on the router to figure this out, i.e. always navigate but exclude "q" if unfiltered. There may be more edge cases which wouldn't be caught here, e.g.

- Navigate to laggy filtered page from unfiltered
- Change my mind and remove the filter while it's loading

In that situation, the router would probably cancel the first navigate.

rsasch: done

PR #269, opened 2018-05-03
Implement log icons update

rsasch:
- replaced status icons with clarity icons and tooltips
- combined log columns into one and replaced links with clarity icons and tooltips

### DIFF
--- a/ui/src/app/job-details/tasks/tasks.component.css
+++ b/ui/src/app/job-details/tasks/tasks.component.css
@@ -55,9 +55,21 @@
 
 clr-tooltip-content {
   background-color: rgba(0, 0, 0, 0.65);
-  padding: 0.25rem;
+  padding: 0.25rem 0.5rem;
   font-size: 0.65rem;
   color: #fff;
   display: block;
   margin: -0.6rem 0.5rem 0 0;
+}
+
+clr-icon {
+  margin-right: 1rem;
+}
+
+::ng-deep clr-icon.has-alert.log-error svg.can-alert path.clr-i-alert.clr-i-outline--alerted {
+  fill: #e62700;
+}
+
+::ng-deep clr-icon.has-alert.log-error.disabled svg.can-alert path.clr-i-alert.clr-i-outline--alerted {
+  fill: currentColor;
 }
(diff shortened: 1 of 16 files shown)


ui/src/app/job-details/tasks/tasks.component.css
```diff
@@ -52,3 +52,24 @@
...
+  color: #fff;
+  display: block;
+  margin: -0.6rem 0.5rem 0 0;
+}
+
+clr-icon {
+  margin-right: 1rem;
+}
+
+::ng-deep clr-icon.has-alert.log-error svg.can-alert path.clr-i-alert.clr-i-outline--alerted {
```
calbach: `::ng-deep` I think should be avoided if possible. This CSS selector is also pretty complicated and I could imagine it breaking if we upgraded our clarity icons version, for example. Have you looked into whether that is any other way to apply this style?

ui/src/app/job-details/tasks/tasks.component.html
```diff
@@ -49,20 +51,27 @@
...
-        </mat-cell>
-      </ng-container>
-      <ng-container matColumnDef="stderr">
-        <mat-header-cell *matHeaderCellDef> Stderr </mat-header-cell>
-        <mat-cell *matCellDef="let t" class="column">
-          <a class="log-item" href="{{ getResourceUrl(t.stderr) }}">
-            {{ getResourceFileName(t.stderr) }}
-          </a>
+          <clr-tooltip *ngIf="getResourceUrl(t.stdout)">
+            <a class="log-item" href="{{ getResourceUrl(t.stdout) }}">
```
calbach: So the link goes into the tooltip? Should the icon also be a link? Could you provide a link to the mocks for this? Couldn't find this interaction.

ui/src/app/job-details/tasks/tasks.component.css
rsasch: I wasn't able to get this to work without using ::ng-deep, but I'm definitely open to not using it.  I can definitely shorten the list of selectors.

ui/src/app/job-details/tasks/tasks.component.html
calbach: Per offline discussion, apparently this does put the link onto the icon and not in the tooltip. I find the structure of this HTML surprising, but it appears to be the intended usage.

calbach:
Not sure where the new Circle failure came from, might be unrelated to this.

I created a change to pin the angular-cli which should hopefully fix this.

PR #273, opened 2018-05-10
[UI] Truncate log files from GCS at 1Mb

bfcrampton:


calbach:
Sorry, was trying to get my local dev environment working so I could try this out. I wanted to check whether this HTTP header works instead: https://cloud.google.com/storage/docs/json_api/v1/parameters#contentrange

Not sure about the behavior if you request past the end of the file. I will have to circle back to my dev environment later - but feel free to test that out in the meantime. Would be nicer than making metadata requests and would let you show the head of the file instead.

bfcrampton:
`content-range` is defined as:
>  A header used in some upload requests and download responses.

However the [`range`](https://cloud.google.com/storage/docs/json_api/v1/parameters#range) is what we want:
> A header used in some upload responses and download requests

Verified that setting `range: 'bytes=0-999999'` gives us exactly what we want, allowing users to see the head of the file and no metadata requests.

ui/src/app/core/gcs.service.ts
```diff
@@ -7,30 +7,21 @@ declare const gapi: any;
...
 
-  private static readonly apiPrefix = "https://www.googleapis.com/storage/v1"
+  private static readonly apiPrefix = "https://www.googleapis.com/storage/v1";
+  private static readonly maximumBytes = "1000000";
 
   constructor(private readonly authService: AuthService) {}
 
   readObject(bucket: string, object: string): Promise<string> {
+    return this.isAuthenticated()
+      .then( () => this.getObjectData(bucket, object))
```
calbach: nit: rm space before `()`

ui/src/app/core/gcs.service.ts
```diff
@@ -7,30 +7,21 @@ declare const gapi: any;
...
-            return "";
-          }
-          return Promise.reject({
-            status: response.status,
-            title: "Could not read file",
-            message: response.body,
-          });
-        })
+        return Promise.resolve();
       } else {
```
calbach: nit: drop the else to reduce indentation

ui/src/app/core/gcs.service.ts
```diff
@@ -40,4 +31,31 @@ export class GcsService {
...
+  getObjectData(bucket: string, object: string): Promise<string> {
+    return gapi.client.request({
+      path: `${GcsService.apiPrefix}/b/${bucket}/o/${object}`,
+      params: {alt: 'media'},
+      headers: {range: 'bytes=0-999999'} // Limit file size to 1Mb
+    })
+    .then(response => {
+      if (response.headers["content-length"] == GcsService.maximumBytes) {
+        return response.body + "\n\nTruncated download at 1Mb...";
+      } else {
```
calbach: nit: can drop else here too

ui/src/app/core/gcs.service.ts
```diff
@@ -40,4 +31,31 @@ export class GcsService {
       }
     });
   }
+
+  getObjectData(bucket: string, object: string): Promise<string> {
+    return gapi.client.request({
+      path: `${GcsService.apiPrefix}/b/${bucket}/o/${object}`,
+      params: {alt: 'media'},
+      headers: {range: 'bytes=0-999999'} // Limit file size to 1Mb
```
calbach: `s/Mb/MB/` here and below; lowercase  is bits

ui/src/app/core/gcs.service.ts
```diff
@@ -7,30 +7,21 @@ declare const gapi: any;
 @Injectable()
 export class GcsService {
 
-  private static readonly apiPrefix = "https://www.googleapis.com/storage/v1"
+  private static readonly apiPrefix = "https://www.googleapis.com/storage/v1";
+  private static readonly maximumBytes = "1000000";
```
calbach: opt: This feels a bit wrong as a string.. since you're using `==` below it would still actually work if this were a number.

Alternatively, switch to number then `contentLength === String(maximumBytes)` below

ui/src/app/job-details/resources/resources.component.ts
```diff
@@ -111,7 +111,7 @@ export class JobResourcesComponent implements OnInit {
   private readResourceFile(file: string): Promise<[string, string]> {
     let bucket = ResourceUtils.getResourceBucket(this.job.extensions.logs[file]);
     let object = ResourceUtils.getResourceObject(this.job.extensions.logs[file]);
-    return this.gcsService.readObject(bucket, object)
+    return this.gcsService.readObject('bryancrampton-testing', 'eligibility_very_large.csv')
```
calbach: revert

bfcrampton: Oops sorry testing.

PR #274, opened 2018-05-10
added ui and api restrictions on label keys & values

rsasch:
see https://elastc.com/c/jsfz&&H7/304-label-user-input-validation


servers/cromwell/jobs/__main__.py
```diff
@@ -67,6 +68,10 @@
 try:
     with open(capabilities_path) as f:
         capabilities_config = json.load(f)
+    for settings in capabilities_config['displayFields']:
+        if settings['field'].startswith('labels.'):
```
samanehsan: It looks like you can specify `minLength`, `maxLength` and a regex `pattern` for fields in the swagger definition, so you could do that for the labels instead of having to write out these checks: https://swagger.io/docs/specification/data-models/data-types/

bfcrampton:
Generally when we make changes to the API and autogenerated models we have a PR for just for that change sent to someone at the Broad and Verily. This helps keep the auto-generation clutter out of the implementation code, and allows both sides to provide input on API changes. Would you mind separating out the changes to `DisplayField` out into a separate PR?

servers/cromwell/jobs/__main__.py
```diff
@@ -67,6 +68,10 @@
...
+            if len(settings['field']) == 7 or len(settings['field']) > 262:
```
rexwangcc: nit: I would define static variables
```python
CROMWELL_LABEL_MIN_LENGTH = 1
CROMWELL_LABEL_MAX_LENGTH = 256
LABELS_PREFIX_LENGTH = len('.labels')
```
at the top of this file, and refactor this line to:
`if not CROMWELL_LABEL_MIN_LENGTH < len(settings['field']) - LABELS_PREFIX_LENGTH < CROMWELL_LABEL_MAX_LENGTH:`

But this is totally optional.

rsasch: apologies -- this PR is premature, as it relies on label editing being fully implemented, which it isn't

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -68,13 +68,11 @@ def update_job_labels(id, body, **kwargs):
 
     # Follow API spec
     result = response.json()
-    all_labels = get_job(id).labels
-    if not all_labels:
-        all_labels = {}
+    if not result.get('labels'):
+        result.get('labels').update({})
```
rexwangcc: Could you clarify when this situation(the returned result under `labels` field is `None`) happens? And why is it necessary to make it as an empty dict`{}` here?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -68,13 +68,11 @@ def update_job_labels(id, body, **kwargs):
...
 
     # Redundantly update all_labels with updated labels to provide consistency guarantees
```
rexwangcc: Should we remove this comment, since the latest Cromwell returns all the labels for the `PATCH` now?

PR #275, opened 2018-05-10
backend changes for label editing

rsasch:
part of https://elastc.com/c/0xVkDwyc/38-change-labels-on-jobs


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -68,13 +68,11 @@ def update_job_labels(id, body, **kwargs):
 
     # Follow API spec
     result = response.json()
-    all_labels = get_job(id).labels
-    if not all_labels:
-        all_labels = {}
+    if not result.get('labels'):
+        result.get('labels').update({})
 
     # Redundantly update all_labels with updated labels to provide consistency guarantees
```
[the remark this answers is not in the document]
rsasch: Good point – removed.

api/jobs.yaml
```diff
@@ -149,6 +149,19 @@ definitions:
         type: string
       display:
         type: string
+      editable:
+        type: boolean
+        default: false
+      fieldType:
+        type: string
+        default: text
+        enum:
```
bfcrampton: nit: I might separate this out into its own enum definition like we do for [Status](https://github.com/DataBiosphere/job-manager/blob/master/api/jobs.yaml#L350). This would allow `FileType` to be re-used if we wanted to also add it to say `CommonLabels`.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -68,13 +68,9 @@ def update_job_labels(id, body, **kwargs):
...
     result = response.json()
-    all_labels = get_job(id).labels
-    if not all_labels:
-        all_labels = {}
-
-    # Redundantly update all_labels with updated labels to provide consistency guarantees
-    all_labels.update(result.get('labels'))
-    return UpdateJobLabelsResponse(labels=all_labels)
+    if not result.get('labels'):
+        result.get('labels').update({})
```
bfcrampton: Is this still dependent on the cromwell update? If so you can probably leave this off for now.

api/jobs.yaml
```diff
@@ -149,6 +149,19 @@ definitions:
...
+          - text
+          - list
+      fieldValues:
```
bfcrampton: nit: Could you add a description to this field to make it really clear what it means? I also might name it `possibleValues` or `possibleFieldValues`, fieldValues feels a little vague, but up to you.

servers/cromwell/README.md
```diff
@@ -38,15 +38,33 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     {
       "field": "labels.cromwell-workflow-id",
       "display": "Workflow ID"
+    },
```
bfcrampton: Might not add this to the README quite yet since it won't make any visible changes to UI.

PR #277, opened 2018-05-10
Fix local development for Google workstations

calbach:
"api" has a DNS mapping now internally for us. Apparently our workstation DNS mapping supersedes the internal docker network.

I couldn't figure out how to reconfigure that - switching to an internal only docker network makes the proxy inaccessible locally. So I took the easy way out and just used a different service name.

### DIFF
--- a/cromwell-compose-staging.yml
+++ b/cromwell-compose-staging.yml
@@ -1,10 +1,10 @@
 version: '2.1'
 services:
-  ui:
+  jmui:
     extends:
       file: common-compose.yml
-      service: ui
-    command: ["npm", "run-script", "ng", "--", "serve", "--host", "ui", "--env=cromwell_staging"]
+      service: jmui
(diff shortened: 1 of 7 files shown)


cromwell-compose-staging.yml
```diff
@@ -1,10 +1,10 @@
 version: '2.1'
 services:
-  ui:
+  jmui:
```
calbach: @samanehsan @rexwangcc  IIRC one or more of these compose files are unused. Which ones can I delete?

rexwangcc: Sorry for this late reply, I think it is ok to remove the `cromwell-compose-staging.yml`, `environment.cromwell.staging.ts` and its registry in `.angular-cli.json`, since they are identical to the dev ones, except the names.

calbach: Cool thanks, I will remove that in a follow-up PR.

PR #279, opened 2018-05-17
added handler for shift-clicking on job checkbox to select multiple jobs

rsasch:
see https://elastc.com/c/UhPkLrGx/327-shift-select-many

### DIFF
--- a/ui/src/app/job-list/table/table.component.spec.ts
+++ b/ui/src/app/job-list/table/table.component.spec.ts
@@ -279,6 +279,31 @@ describe('JobsTableComponent', () => {
     });
   }))
 
+  it('should select multiple jobs on shift-click', async( () => {
+    fixture.detectChanges();
+    let de: DebugElement = fixture.debugElement;
+
+    const checkboxes = de.queryAll(By.css('.mat-checkbox-input'));
+
+    checkboxes[2].nativeElement.click();
+    fixture.detectChanges();
+    let shiftClick = new MouseEvent('click', {
+      'button' : 0,
+      'buttons' : 0,
+      'bubbles': true,
+      'cancelable' : false,
+      'shiftKey': true
+    });
+    testComponent.jobsTableComponent.updateCheckBoxSelection(jobs[3], shiftClick);
+    fixture.detectChanges();
+
+    expect(checkboxes[1].nativeElement.checked).toEqual(false);
+    expect(checkboxes[2].nativeElement.checked).toEqual(true);
+    expect(checkboxes[3].nativeElement.checked).toEqual(true);
+    expect(checkboxes[4].nativeElement.checked).toEqual(true);
+    expect(checkboxes[5].nativeElement.checked).toEqual(false);
+  }))
+
   @Component({
     selector: 'jm-test-table-component',
     template:
--- a/ui/src/app/job-list/table/table.component.spec.ts
+++ b/ui/src/app/job-list/table/table.component.spec.ts
@@ -51,6 +51,20 @@ describe('JobsTableComponent', () => {
       ]
     };
 
+  const shiftClick = new MouseEvent('click', {
+    'button' : 0,
+    'buttons' : 0,
+    'bubbles': true,
+    'cancelable' : false,
+    'shiftKey': true
+  });
+
+  function getJobCheckboxes(): DebugElement[] {
+    let jobCheckboxes = fixture.debugElement.queryAll(By.css('.mat-checkbox-input'));
+    jobCheckboxes.shift();
+    return jobCheckboxes;
+  }
+
   function testJobs(): QueryJobsResult[] {
     return [{
       id: 'JOB1',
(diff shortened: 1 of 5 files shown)


ui/src/app/job-list/table/table.component.ts
```diff
@@ -200,4 +200,26 @@ export class JobsTableComponent implements OnInit {
       this.jobs.forEach(j => this.selection.select(j));
     }
   }
+
+  updateCheckBoxSelection(clickedJob: QueryJobsResult, event: MouseEvent): void {
+    // if the user has shift-clicked on a job, find the last selected job and
+    // select all jobs between the two
+    if (event.shiftKey && this.selection.selected.length) {
```
calbach: nit: `.length` sort of redundant, `[]` is falsey in JS

alternatively: `this.selection.hasValue()`

ui/src/app/job-list/table/table.component.ts
```diff
@@ -200,4 +200,26 @@ export class JobsTableComponent implements OnInit {
...
+      const lastJobClicked = this.selection.selected[this.selection.selected.length - 1];
+      this.selection.select(clickedJob);
+      let indexes = [this.jobs.findIndex(j => j.id === lastJobClicked.id),
+        this.jobs.findIndex(j => j.id === clickedJob.id)];
+      indexes.sort();
+      var startSelecting = false;
+      for (let job of this.jobs) {
```
calbach: I think it would be clearer to use index iteration here, then drop all of the ID comparisons and job ID matching inside the loop:

```
for (let i = indexes[0] + 1; i < indexes[1]-1; i++) {
  this.selection.select(this.jobs[i]);
}
```

ui/src/app/job-list/table/table.component.ts
calbach: opt_nit: could invert this condition and return early to avoid indenting everything

ui/src/app/job-list/table/table.component.ts
```diff
@@ -200,4 +200,26 @@ export class JobsTableComponent implements OnInit {
...
     }
   }
+
+  updateCheckBoxSelection(clickedJob: QueryJobsResult, event: MouseEvent): void {
+    // if the user has shift-clicked on a job, find the last selected job and
+    // select all jobs between the two
+    if (event.shiftKey && this.selection.selected.length) {
+      const lastJobClicked = this.selection.selected[this.selection.selected.length - 1];
+      this.selection.select(clickedJob);
+      let indexes = [this.jobs.findIndex(j => j.id === lastJobClicked.id),
```
calbach: very_optional: another (IMO) pretty cool way to write this, but perhaps more complicated:

```
const [from, to] = [
  this.jobs.findIndex(j => j.id === lastJobClicked.id),
  this.jobs.findIndex(j => j.id === clickedJob.id)
].sort();
```

rsasch: actually, I think that's clearer -- thanks!

ui/src/app/job-list/table/table.component.ts
rsasch: good point – done

ui/src/app/job-list/table/table.component.ts
```diff
@@ -200,4 +200,19 @@ export class JobsTableComponent implements OnInit {
...
+  updateCheckBoxSelection(clickedJob: QueryJobsResult, event: MouseEvent): void {
+    // if the user has shift-clicked on a job, find the last selected job and
+    // select all jobs between the two
+    if (event.shiftKey && this.selection.selected) {
+      const lastJobClicked = this.selection.selected[this.selection.selected.length - 1];
+      const [from, to] = [
+        this.jobs.findIndex(j => j.id === lastJobClicked.id),
+        this.jobs.findIndex(j => j.id === clickedJob.id)
+      ].sort();
+      for (let i = from; i <= to; i++) {
```
calbach: I thought that you didn't want to toggle the clicked checkbox? Would this double-toggle the checkbox you clicked?

ui/src/app/job-list/table/table.component.ts
rsasch: it would avoid the indenting, but I think the logic is clearer when you're reading the code this way

rsasch: agreed – done

ui/src/app/job-list/table/table.component.ts
```diff
@@ -200,4 +200,19 @@ export class JobsTableComponent implements OnInit {
...
     }
   }
+
+  updateCheckBoxSelection(clickedJob: QueryJobsResult, event: MouseEvent): void {
+    // if the user has shift-clicked on a job, find the last selected job and
+    // select all jobs between the two
+    if (event.shiftKey && this.selection.selected) {
+      const lastJobClicked = this.selection.selected[this.selection.selected.length - 1];
+      const [from, to] = [
+        this.jobs.findIndex(j => j.id === lastJobClicked.id),
```
calbach: nit: this is probably more accurately the `lastJobSelected`. We may have clicked another one to deselect it.

ui/src/app/job-list/table/table.component.ts
rsasch: It's not toggling, it's marking them as selected; if they are already selected, there will be no change.

calbach: But isn't `change` still running on the target checkbox? Is this depending on the `change` callback happening before `click`?

rsasch: when I tested it, `change` was not triggered when handled by the `click` handler

calbach: Interesting.. that's not what I observe when I test out the mat-checkbox in stackblitz. Are you sure it's not just calling `change`, then calling `click` (overriding the toggle effect)? Can you add logging locally? I'm not sure I understand why normal checkbox interactions would still be working if change is not being triggered.

Anyways - unit tests may reveal more.

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +279,31 @@ describe('JobsTableComponent', () => {
...
 
+  it('should select multiple jobs on shift-click', async( () => {
+    fixture.detectChanges();
+    let de: DebugElement = fixture.debugElement;
+
+    const checkboxes = de.queryAll(By.css('.mat-checkbox-input'));
+
+    checkboxes[2].nativeElement.click();
+    fixture.detectChanges();
+    let shiftClick = new MouseEvent('click', {
```
calbach: nit: const

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +279,31 @@ describe('JobsTableComponent', () => {
     });
   }))
 
+  it('should select multiple jobs on shift-click', async( () => {
```
calbach: Thanks for adding this, would be good to have a few more test cases to make the behavior more explicit, which should be pretty easy copy-paste-modifies of what you have e.g.:

- Shift click on an already-checked job
- Reverse order selection
- Shift click a range with an internal already-checked job in the range
- Verify behavior if you check/uncheck a job (should range select to the previous selection)

rsasch: Will do – can you clarify the last one?

rsasch: Thanks – done

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +279,103 @@ describe('JobsTableComponent', () => {
...
+      'buttons' : 0,
+      'bubbles': true,
+      'cancelable' : false,
+      'shiftKey': true
+    });
+    testComponent.jobsTableComponent.updateCheckBoxSelection(jobs[2], shiftClick);
+    fixture.detectChanges();
+
+    expect(checkboxes[1].nativeElement.checked).toEqual(false);
+    expect(checkboxes[2].nativeElement.checked).toEqual(false);
```
calbach: The 1-based index is confusing here. Perhaps you could do a `checkboxes.shift()` to get rid of the "select all" checkbox so these indices match up with `jobs`. Or else modify your selector for checkbox inputs

calbach: opt: if you do the above comment, you could probably clean up these expectations a bit:

`expect(checkboxes.map(c => c.nativeElement.checked)).toEqual([false, false, true, true, true]);`

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +279,103 @@ describe('JobsTableComponent', () => {
...
+
+  it('should select multiple jobs on shift-click (with an internal already-checked job in the range)', async( () => {
+    fixture.detectChanges();
+    let de: DebugElement = fixture.debugElement;
+    const checkboxes = de.queryAll(By.css('.mat-checkbox-input'));
+
+    checkboxes[2].nativeElement.click();
+    checkboxes[1].nativeElement.click();
+    fixture.detectChanges();
+    const shiftClick = new MouseEvent('click', {
```
calbach: nit: probably extract this to a private helper

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +279,103 @@ describe('JobsTableComponent', () => {
...
+      'bubbles': true,
+      'cancelable' : false,
+      'shiftKey': true
+    });
+    testComponent.jobsTableComponent.updateCheckBoxSelection(jobs[4], shiftClick);
+    fixture.detectChanges();
+
+    expect(testComponent.jobsTableComponent.allSelected()).toEqual(true);
+  }))
+
```
calbach: Only missing case I was curious about was the boundary being already checked. What should happen there?

calbach: For example, job3 is already selected. Select job1, then shift-click job3.

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -279,6 +293,66 @@ describe('JobsTableComponent', () => {
...
+    const jobCheckboxes = getJobCheckboxes();
+    jobCheckboxes[1].nativeElement.click();
+    testComponent.jobsTableComponent.updateCheckBoxSelection(jobs[3], shiftClick);
+    fixture.detectChanges();
+
+    expect(jobCheckboxes.map(c => c.nativeElement.checked))
+      .toEqual([false, true, true, true, false]);
+  }))
+
+  it('should select multiple jobs on shift-click (reverse order)', async( () => {
```
calbach: nit: One last thing I just noticed - please drop the space after async( `async( ()` -> `async(()`

rsasch: well spotted – done

PR #280, opened 2018-05-18
Update README.md

kbergin:
Updated a bit of the roadmap section. May revisit again later.


cjllanwarne:
Did something odd happen with the git history here? 

As is, it looks like an uncontroversial one-liner to me - can we merge it?

cjllanwarne:
@kbergin can we merge this now or is there more to come?

kbergin:
@cjllanwarne we can merge it, but I think there are needs for updates to the readme in the future.

PR #281, opened 2018-05-22
[UI] Cache navigation JobDetail -> JobList

bfcrampton:


bfcrampton:
yapf failures are unrelated to this change, due to a new update in yapf. Will rebase to fix once https://github.com/DataBiosphere/job-manager/pull/282 is merged.

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  shouldDetach(route: ActivatedRouteSnapshot): boolean {
+    // Currently we only cache the job list page, we may want to also do
+    // details pages in the future.
+    return !!route.routeConfig && route.routeConfig.path == "jobs";
+  }
+
+  store(route: ActivatedRouteSnapshot, handle: DetachedRouteHandle): void {
+    let storedRoute = {
```
calbach: nit: I'd probably just inline this and avoid the local var

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  retrieve(route: ActivatedRouteSnapshot): DetachedRouteHandle {
+    if (!route.routeConfig || !RouteReuse.cached[RouteReuse.routeKey(route)]) {
+      return null;
+    }
+    return RouteReuse.cached[RouteReuse.routeKey(route)].handle;
+  }
+
+  shouldReuseRoute(future: ActivatedRouteSnapshot, curr: ActivatedRouteSnapshot): boolean {
```
calbach: The docs aren't very informative on this one, but I'm slightly surprised by the implementation of `return false;`. Is this just meant to answer whether we can reuse a route for a particular transition (curr->future)?

I'm thinking perhaps you should copy the default impl here, does that cause issues?:
https://github.com/angular/angular/blob/9141424ac69f8ff9702485e7f00e58fa503d8c42/packages/router/src/route_reuse_strategy.ts#L70

ui/src/app/job-list/job-list-resolver.service.ts
```diff
@@ -21,11 +22,19 @@ export class JobListResolver implements Resolve<JobStream> {
 
   resolve(route: ActivatedRouteSnapshot,
```
calbach: Man - the resolver runs again when you reattach to a component? That's pretty surprising.. I couldn't find any documentation one way or the other, so we'll have to base this on your local experimentation. Should double-check this again if you wind up changing reuseRoute per my below comment.

ui/src/app/job-list/job-list-resolver.service.ts
```diff
@@ -21,11 +22,19 @@ export class JobListResolver implements Resolve<JobStream> {
...
-        .loadAtLeast(initialBackendPageSize)
-        .then(resp => jobStream);
+    const routeKey = RouteReuse.routeKey(route);
+    // If this route has been cached do not wait for the load to display the
+    // component. Instead, just mark the JobStream as needing a refresh.
+    if (!!RouteReuse.cached[routeKey]) {
+      let jobStream = RouteReuse.cached[routeKey].snapshot.data.stream;
+      jobStream.setNeedsRefresh();
+      return Promise.resolve(jobStream);
+    } else {
```
calbach: opt_nit: drop the else

ui/src/app/job-list/job-list-resolver.service.ts
```diff
@@ -21,11 +22,19 @@ export class JobListResolver implements Resolve<JobStream> {
...
+      let jobStream = new JobStream(this.jobManagerService,
```
calbach: nit const

ui/src/app/job-list/job-list.component.ts
```diff
@@ -66,15 +67,19 @@ export class JobListComponent implements OnInit {
...
         this.handleError(event.error);
       }
+      if (event instanceof NavigationEnd && this.jobStream.value.needsRefresh) {
+        this.reloadJobs(this.route.snapshot.queryParams['q'], false);
+      }
     });
   }
 
-  reloadJobs(query: string) {
+  reloadJobs(query: string, displayLoading: boolean): void {
```
calbach: opt: could use a default param value here `displayLoading = true`

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,52 @@
...
+@Injectable()
+export class RouteReuse implements RouteReuseStrategy {
+
+  public static cached: { [key: string]: any } = {};
+
+  public static routeKey(route: ActivatedRouteSnapshot): string {
+    // We currently only cache the jobs-list route so we can store them based
+    // on just the query parameter. If we add job-detail caching we should
+    // make this key more specific.
+    return route.queryParams.q
```
calbach: ;

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  DetachedRouteHandle,
+  RouteReuseStrategy,
+  Routes
+} from '@angular/router';
+
+// Configure reuse of particular routes so components do not have to re-load
+@Injectable()
+export class RouteReuse implements RouteReuseStrategy {
+
+  public static cached: { [key: string]: any } = {};
```
calbach: I'm wondering whether this cache would make more sense to put into the `JobListResolver`. Thoughts?

calbach: A timed cache expiration would be nice here. A simple `setTimeout()` might work well actually.

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  ActivatedRouteSnapshot,
+  DetachedRouteHandle,
+  RouteReuseStrategy,
+  Routes
+} from '@angular/router';
+
+// Configure reuse of particular routes so components do not have to re-load
+@Injectable()
+export class RouteReuse implements RouteReuseStrategy {
```
calbach: Is this testable or does the test routing module not cover this stuff? (unit tests for this class by itself are not worth it)

ui/src/app/job-list/job-list-resolver.service.ts
```diff
@@ -21,11 +22,19 @@ export class JobListResolver implements Resolve<JobStream> {
...
-
-    let jobStream = new JobStream(this.jobManagerService,
-                                  URLSearchParamsUtils.unpackURLSearchParams(route.queryParams['q']));
-    return jobStream
-        .loadAtLeast(initialBackendPageSize)
-        .then(resp => jobStream);
+    const routeKey = RouteReuse.routeKey(route);
+    // If this route has been cached do not wait for the load to display the
+    // component. Instead, just mark the JobStream as needing a refresh.
+    if (!!RouteReuse.cached[routeKey]) {
```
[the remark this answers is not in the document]
bfcrampton: Done.

ui/src/app/job-list/job-list-resolver.service.ts
bfcrampton: Done.

ui/src/app/job-list/job-list-resolver.service.ts
bfcrampton: Done.

ui/src/app/job-list/job-list.component.ts
bfcrampton: Done.

ui/src/app/route-reuse.service.ts
bfcrampton: I kind of like it here since this is the class which fetches and stores from the cache. Instead I added a public interface to this class for determining if a route is cached and getting the cached data. Let me know if you think that's better.

bfcrampton: Do you mean do this instead of the `needsRefresh` on `JobListView` so that after some amount of time going back to this component totally recreates it?

I think I kind of prefer the current strategy of never blowing away the cache so that you see something on the page immediately always (except first load). Instead maybe we could change `needsRefresh` to a timestamp and after some amount of time trigger a refresh of the page (with the loading indicator over the list). What do you think?

ui/src/app/route-reuse.service.ts
bfcrampton: Done.

ui/src/app/route-reuse.service.ts
bfcrampton: Done.

ui/src/app/route-reuse.service.ts
bfcrampton: Yeah I wasn't 100% sure what this is actually used for but I think its if the next component being navigated to is the same as the current one it allows it to be reused instead of recreating. I think this is kind of separate from the caching we're doing. But you're right I can use the default implementation for this. Although it doesn't fix the resolver issue unfortunately.

ui/src/app/route-reuse.service.ts
bfcrampton: I tried testing it in the app.component.spec but unfortunately with no luck, it didn't seem to get triggered for some reason.

ui/src/app/job-list/job-list-resolver.service.ts
bfcrampton: Yup I was also surprised to find this...pretty annoying. Unfortunately `shouldReuseRoute` and all the other callbacks on `RouteReuseStrategy` are called after the resolver is triggered so we have to check the cache here.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -66,15 +67,19 @@ export class JobListComponent implements OnInit {
...
         this.handleError(event.error);
       }
+      if (event instanceof NavigationEnd && this.jobStream.value.needsRefresh) {
+        this.reloadJobs(this.route.snapshot.queryParams['q'], false);
+      }
     });
   }
 
-  reloadJobs(query: string) {
+  reloadJobs(query: string, displayLoading = true): void {
```
calbach: opt: it would be nice if we had the less invasive "snackbar" loading indicator in the false case here i.e. give the user some indication that something is happening so that they aren't surprised when the UI shuffles around

ui/src/app/route-reuse.service.ts
calbach: OK - that makes sense. Could you make this not static and inject it into the JobListResolver instead then?

ui/src/app/job-list/job-list.component.ts
```diff
@@ -33,7 +33,8 @@ export class JobListComponent implements OnInit {
   // whenever we change query parameters, via a subscription.
   jobs = new BehaviorSubject<JobListView>({
     results: [],
-    exhaustive: false
+    exhaustive: false,
+    needsRefresh: false
```
calbach: So - problem here I think - `reloadJobs()` works by simply discarding the old job stream and creating a new one. The stream in the cache would be unmodified and we would not readd this to the cache. So the cached stream is just the first job stream for the given route.

calbach: nit: on naming - I think I might prefer "stale" rather than "needsRefresh".

ui/src/app/route-reuse.service.ts
calbach: My second concern here was around this cache growing indefinitely. So at first I was going to suggest an LRU cache + expiry similar to what you're describing. Then I thought maybe kill two birds with one stone with a timed expiry, since there's a limit on how quickly a person can generate queries (I'm not sure whether this would be a real problem even for very long-lived tabs). Simply for expiry handling, agreed that the approach you describe can work.

ui/src/app/route-reuse.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    RouteReuse.cached[RouteReuse.routeKey(route)] = {
+        snapshot: route,
+        handle: handle
+    };
+  }
+
+  shouldAttach(route: ActivatedRouteSnapshot): boolean {
+    return !!route.routeConfig
+      && route.routeConfig.path == "jobs"
+      && !!RouteReuse.cached[RouteReuse.routeKey(route)];
```
calbach: `isCached`

ui/src/app/shared/job-stream.ts
```diff
@@ -38,14 +40,23 @@ export class JobStream extends BehaviorSubject<JobListView> {
...
+          exhaustive: !resp.nextPageToken,
+          stale: false
         });
         return resp;
       });
     });
     return this.queryPromise;
   }
 
+  public setNeedsRefresh(): void {
```
calbach: `setStale()` to match

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -98,7 +99,8 @@ describe('JobListComponent', () => {
       providers: [
         {provide: JobManagerService, useValue: fakeJobService},
         {provide: CapabilitiesService, useValue: new FakeCapabilitiesService(capabilities)},
-        JobListResolver
+        JobListResolver,
+        RouteReuse
```
calbach: Can you add a test case that exercises the "stale" setting?

bfcrampton: Done.

ui/src/app/shared/job-stream.ts
bfcrampton: Done.

PR #285, opened 2018-05-29
individual label editing

rsasch:
see https://elastc.com/c/0xVkDwyc/38-change-labels-on-jobs


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -68,13 +68,9 @@ def update_job_labels(id, body, **kwargs):
...
     result = response.json()
-    all_labels = get_job(id).labels
-    if not all_labels:
-        all_labels = {}
-
-    # Redundantly update all_labels with updated labels to provide consistency guarantees
-    all_labels.update(result.get('labels'))
-    return UpdateJobLabelsResponse(labels=all_labels)
+    if not result.get('labels'):
+        result.get('labels').update({})
```
calbach: I don't think this makes sense anymore. `update()` is a method on a dict, and this if statement was attempting to catch when `labels` is null (`None` in python). Therefore this block will either never execute, or if it does, fail with method not defined. Remove or `result['labels'] = {}`

ui/src/app/core/job-manager.service.ts
```diff
@@ -103,6 +105,15 @@ export class JobManagerService {
       .catch((e) => this.handleError(e));
   }
 
+  updateJobLabels(id: string, req: UpdateJobLabelsRequest): Promise<UpdateJobLabelsResponse> {
+    return this.http.post(`${environment.apiUrl}/jobs/${id}/updateLabels`,
+      req,
+      new RequestOptions({headers: this.getHttpHeaders()}))
+      .toPromise()
+      .then((response) => response.status == 200)
```
calbach: When you call then/catch on a promise, it returns a new promise with a potentially different resolved value. So this method returns a `Promise<boolean>`, which is probably not what you were aiming to do. You can just drop the `then`.

ui/src/app/job-list/table/table.component.html
```diff
@@ -67,7 +67,43 @@
...
+                </button>
+              </div>
+              <mat-menu #menu="matMenu" class="wide-menu" xPosition="before" yPosition="below" [overlapTrigger]="false">
+                <mat-form-field>
+                  <input matInput
+                         #labelField
+                         value="{{ getFieldValue(j, df) }}"
+                         maxlength="255"
+                         (click)="$event.stopPropagation()">
+                  <mat-hint align="end">{{labelField.value?.length || 0}}/255</mat-hint>
```
calbach: Where's the 255 coming from? Comment? Or pull from a constant and comment there?

ui/src/app/job-list/table/table.component.html
```diff
@@ -67,7 +67,43 @@
...
+                </mat-form-field>
+                <button class="apply-link" (click)="setFieldValue(j, df.field, labelField.value)">Apply</button>
+              </mat-menu>
+            </div>
+            <mat-form-field *ngSwitchCase="'list'">
+              <mat-select #labelSelect
+                          [value]="getFieldValue(j, df) ? getFieldValue(j, df) : null"
```
calbach: Is the ternary necessary here? What falsey value are you guarding against that you want to convert to null?

ui/src/app/job-list/table/table.component.html
```diff
@@ -67,7 +67,43 @@
...
+                          (change)="setFieldValue(j, df.field, labelSelect.value)">
+                <mat-option value="">(none)</mat-option>
+                <mat-option *ngFor="let fieldOption of getFieldOptions(df)"
+                            [value]="fieldOption"
+                            [attr.selected]="(getFieldValue(j, df) == fieldOption) ? 'selected' : null"
```
calbach: nit: parens unneeded, ternary is generally very low precedence

ui/src/app/job-list/table/table.component.ts
```diff
@@ -90,6 +91,23 @@ export class JobsTableComponent implements OnInit {
     return job.status == JobStatus.Submitted || job.status == JobStatus.Running;
   }
 
+  canEdit(df: DisplayField): boolean {
+    return df.editable;
+  }
+
+  setFieldValue(job: QueryJobsResult, displayField: string, value: string) {
+    const labelParts = displayField.split('.');
+    const label = (labelParts[1] != null) ? labelParts[1] : value;
```
calbach: nit: would drop parens here as well

ui/src/app/job-list/table/table.component.html
rsasch: this is so that, if the field is empty, it won't select and display the "(none)" option, which muddies up the table

ui/src/app/job-list/table/table.component.ts
```diff
@@ -90,6 +91,23 @@ export class JobsTableComponent implements OnInit {
...
+    let req: UpdateJobLabelsRequest = {};
+    req.labels = JSON.parse("{\"" + label + "\":\"" + value + "\"}");
+    this.jobManagerService.updateJobLabels(job.id, req)
+      .then(() => {
+        job.labels[label] = value;
```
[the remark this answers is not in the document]
rsasch: I found out that the Cromwell API will respond with the new value(s) even if the update is not made (ran into this when there was a bug), so getting it from the response will only add more code.

servers/cromwell/jobs/__main__.py
```diff
@@ -67,6 +67,11 @@
 try:
     with open(capabilities_path) as f:
         capabilities_config = json.load(f)
+    for settings in capabilities_config['displayFields']:
+        if settings['field'].startswith('labels.'):
+            if len(settings['field']) == 7 or len(settings['field']) > 262:
```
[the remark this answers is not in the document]
rsasch: added, but with one small change (LABELS_PREFIX vs LABELS_PREFIX_LENGTH)

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: makes sense – removed

ui/src/app/job-list/table/table.component.html
calbach: I see, `|| null` would be slightly more readable in that case

ui/src/app/job-list/table/table.component.ts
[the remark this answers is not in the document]
rsasch: When I take out the fallback, the requests still succeed, but the snackbar pops up empty (with just a "dismiss" button) when the list of jobs reloads.  Any idea why?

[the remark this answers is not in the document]
calbach: Most likely this means that a non-API error hit the `handleError` method, e.g. a Javascript runtime error.

I don't really know what's going on here specifically. In cases like these, I would recommend putting a `debugger;` statement above this line, and tracing the code in Chrome Inspector to see what is happening.

ui/src/app/job-list/table/table.component.html
calbach: I don't feel too strongly on this, but this is a relatively common pattern and I don't recall seeing this style used before. Alternatively, extract this into a helper method, since ternaries as a whole are generally not considered very good for readability.

ui/src/app/job-list/table/table.component.html
calbach: It could eventually go into capabilities, but that's probably overkill right now. I think a constant with clear documentation of the origin and current status of this limit is good for now.

ui/src/app/job-list/table/table.component.ts
```diff
@@ -90,6 +91,23 @@ export class JobsTableComponent implements OnInit {
...
 
+  canEdit(df: DisplayField): boolean {
+    return df.editable;
+  }
+
+  setFieldValue(job: QueryJobsResult, displayField: string, value: string) {
+    const labelParts = displayField.split('.');
+    const label = (labelParts[1] != null) ? labelParts[1] : value;
+    let req: UpdateJobLabelsRequest = {};
+    req.labels = JSON.parse("{\"" + label + "\":\"" + value + "\"}");
```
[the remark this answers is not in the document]
calbach: My concern is with the usage of JSON.parse. Why would you need to use that here rather than constructing a Javascript object normally?

ui/src/app/job-list/table/table.component.html
rsasch: removed the `()`

ui/src/app/job-list/table/table.component.html
rsasch: good point – done

ui/src/app/job-list/table/table.component.ts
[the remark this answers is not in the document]
calbach: I don't understand the comment about the Cromwell API. My understanding is that it should be returning all labels on the job (including updated values). Is that not what you're seeing?

ui/src/app/job-list/table/table.component.html
rsasch: added – please let me know if it's not clear

ui/src/app/job-list/table/table.component.ts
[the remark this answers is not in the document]
rsasch: The Cromwell API will return all the labels on the job, including any changes that have been requested.  The return of these values does not mean the changes have actually been applied to the job; if something goes wrong with updating the job (whether it's a Cromwell bug or something else), the return values will show the requested changes, regardless.  Hence, the returned data is not any more useful than the data that is sent in the request.

ui/src/app/job-list/table/table.component.html
```diff
@@ -67,7 +67,43 @@
...
+              </mat-menu>
+            </div>
+            <mat-form-field *ngSwitchCase="'list'">
+              <mat-select #labelSelect
+                          [value]="getFieldValue(j, df) || null"
+                          (change)="setFieldValue(j, df.field, labelSelect.value)">
+                <mat-option value="">(none)</mat-option>
+                <mat-option *ngFor="let fieldOption of getFieldOptions(df)"
+                            [value]="fieldOption"
+                            [attr.selected]="getFieldValue(j, df) == fieldOption || null"
```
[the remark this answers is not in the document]
rsasch: It should, but the way it behaves in some (possibly most) browsers is any value is interpreted as `true` (e.g. `selected="false"`).

servers/cromwell/jobs/__main__.py
```diff
@@ -67,6 +71,13 @@
 try:
     with open(capabilities_path) as f:
         capabilities_config = json.load(f)
+    for settings in capabilities_config['displayFields']:
+        if settings['field'].startswith(LABELS_PREFIX):
+            if len(settings['field']) == len(LABELS_PREFIX) or len(
```
samanehsan: Is there any way we could define the `minLength` and `maxLength` in the swagger definition of the model so that we don't need to do the validation here?

rsasch: The restrictions can be outlined in swagger, but that means that the problem won't be visible until a user tries to change something relating to that label (the request will fail and user sees an error in the snackbar "Invalid label: ...") via the UI. I think, when it's a configuration problem, it's best to fail when the application is being started up, since the person doing so is more likely to know how to fix that sort of issue (as opposed to a user).  See the discussion here: https://elastc.com/c/jsfz&&H7/304-label-user-input-validation.

samanehsan: Oh that makes sense!

ui/src/app/job-list/table/table.component.ts
```diff
@@ -90,6 +96,26 @@ export class JobsTableComponent implements OnInit {
...
+  canEdit(df: DisplayField): boolean {
+    return df.editable;
+  }
+
+  setFieldValue(job: QueryJobsResult, displayField: string, value: string) {
+    const label = displayField.replace('labels.', '');
+    const req: UpdateJobLabelsRequest = { labels : {} };
+    req.labels[label] = value;
+    this.jobManagerService.updateJobLabels(job.id, req)
+    /* NOTE: currently, response does not reflect whether or not the requested changes to job have
```
calbach: (for Cromwell...)

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -204,10 +208,24 @@ describe('JobsTableComponent', () => {
...
+    let de: DebugElement = fixture.debugElement;
+    expect(de.queryAll(By.css('.cdk-column-labels-status-detail .edit-field')).length)
+      .toEqual(0);
+  }));
+
+  it('should display editable field for job label if config has explicitly said it is editable', async(() => {
+    fixture.detectChanges();
+    let de: DebugElement = fixture.debugElement;
+    expect(de.queryAll(By.css('.cdk-column-labels-comment .edit-field')).length)
+      .toEqual(de.queryAll(By.css('.cdk-column-labels-comment')).length - 1);
```
calbach: I don't think I follow this logic. Why is this not equal to 1?

rsasch: There's one comment edit field per job.

ui/src/app/job-list/table/table.component.ts
rsasch: noted – made comment more specific

ui/src/app/job-list/table/table.component.spec.ts
calbach: Ah thanks - comment?

PR #286, opened 2018-05-30
[RFC] Stock Deployment for Job Manager

rexwangcc:
This version of implementation strictly follows the "Job Manager Deployment Discussion" proposal. This PR:
1. Implements an environment loader.
2. Cleans up all environment specific ts files.
3. Updates the compose file, angular config, and documentation.
4. Adds environment JSON files to assets folder.
5. Standardizes Cromwell environments to 2 categories: `instance` and `caas`

Future works for this PR:
- [x] Try to bring back the `--prod` flag for `ng build`.
- [ ] Attempts to implement the caching mechanism suggested by @bfcrampton .
- [ ] Attempts to inline the config load into the initial page load, so that we're fetching the initial JS bundle and environment.json simultaneously, as suggested by @calbach 
- [ ] Performs another round of benchmarking based on the new implementation.
- [ ] Adds test cases once the implementation becomes stable.


common-compose.yml
```diff
@@ -36,7 +36,7 @@ services:
       - PATH_PREFIX=/api/v1
       - CROMWELL_CREDENTIALS=/app/jobs/config.json
       - CAPABILITIES_CONFIG=/app/jobs/capabilities_config.json
-      - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1}
```
rexwangcc: Since there are more engineers work on Job Manager now, it doesn't make sense to keep this default value here (mint team-specific) anymore.

ui/.angular-cli.json
```diff
@@ -29,11 +29,7 @@
       "environmentSource": "environments/environment.ts",
       "environments": {
         "dev": "environments/environment.ts",
-        "prod": "environments/environment.prod.ts",
-        "dsub_google": "environments/environment.dsub.google.ts",
-        "dsub_local": "environments/environment.dsub.local.ts",
-        "cromwell_dev": "environments/environment.cromwell.dev.ts",
-        "cromwell_caas": "environments/environment.cromwell.caas.ts"
+        "prod": "environments/environment.prod.ts"
       }
```
rexwangcc: Only leave the 2 angular default ts files here.

servers/cromwell/README.md
```diff
@@ -4,11 +4,7 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
 
 ## Development
 
-1. Set the `CROMWELL_URL` environment variable to specify which cromwell instance to use. Job Manager pulls workflow data from `https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1` by default:
-
-    ```
-    export CROMWELL_URL=https://cromwell.test-cromwell.broadinstitute.org/api/workflows/v1
-    ```
+1. Set the `CROMWELL_URL` environment variable to specify which cromwell instance to use.
```
samanehsan: I think it's still useful to have an example url here

ui/src/environments/environmentLoader.ts
```diff
@@ -0,0 +1,20 @@
+import { environment as defaultEnvironment } from './environment';
+
+export const environmentLoader = new Promise<any>( (resolve) => {
```
samanehsan: nit: you can remove the parenthesis around `resolve`

ui/assets/environments/environment.cromwell.caas.json
```diff
@@ -0,0 +1,5 @@
+{
+  "production": false,
+  "apiUrl": "/api/v1",
+  "clientId": "619310558212-u9uso7pdgh7s6r6gqfrs1tm58abdqtf2.apps.googleusercontent.com"
```
samanehsan: Let's use the clientId from the caas-dev google project `652421484530-nnetfth6psch0h0rjt1chc4hru9r0j0o.apps.googleusercontent.com`

servers/cromwell/README.md
```diff
@@ -48,9 +44,42 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
 }
 ```
 
-4. Symbolically link the cromwell docker compose file depending on your `CROMWELL_URL`. For `https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1`, use `cromwell-compose-caas.yaml` otherwise use `cromwell-compose-dev.yaml`, e.g:
+   **Note:** If you want to use use Job Manager against Cromwell-as-a-Service, which is using SAM/Google OAuth for authN and authZ, the `capabilities_config.json` must also include some extra fields, as well as proper scopes, which are shown as below:
```
calbach: nit:  Google Oauth / SAM

SAM -> authz, Google Oauth -> authn  (this has them listed in reverse order)

ui/nginx.prod.conf
```diff
@@ -1,6 +1,9 @@
 # nginx server for static assets. For directive documentation, see
 # http://nginx.org/en/docs/dirindex.html
 
+# Note: this nginx configuration is designed for Google App Engine, to deploy Job Manager on other type of servers,
```
calbach: Thanks for raising this. It's news to me that you would need to override this file. Can you add more detail about what the issue was using this for Kubernetes? Also, if there's another nginx conf file you're using we should consider adding it here or reconciling it with this one.

ui/src/environments/environment.prod.ts
```diff
@@ -1,4 +1,8 @@
 // Placeholder file. Specific production deployments should overwrite this file
```
calbach: Please remove or update comment

ui/src/environments/environmentLoader.ts
calbach: nit: rm space before (resolve)

ui/src/environments/environmentLoader.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  let xmlhttp = new XMLHttpRequest(),
+    method = 'GET',
+    url = './assets/environments/environment.json';
+
+  xmlhttp.open(method, url, true);
+
+  xmlhttp.onload = function() {
```
calbach: Should use an ES6 arrow function as usual

calbach: This should handle `onerror`

ui/src/environments/environmentLoader.ts
```diff
@@ -0,0 +1,20 @@
...
+    if (xmlhttp.status === 200) {
+      resolve(JSON.parse(xmlhttp.responseText));
+    } else {
+      resolve(defaultEnvironment);
```
calbach: This will lead to very strange behaviors. We should be showing an error in this case.

This could perhaps be a follow-up fix, but there should at least be a TODO here. I would prefer that we never attempt to continue running the application if this JSON load fails, as the behavior will be undefined.

ui/nginx.prod.conf
rexwangcc: We are currently using this nginx config: https://github.com/HumanCellAtlas/secondary-analysis/blob/master/deploy/job-manager/nginx.conf.ctmpl All we have to do is creating a configMap object and mount it to the path `/etc/nginx/nginx.conf`

There are not really many problems prevent us from us from using the nginx file here, except:
- k8s cannot find the path for these log dirs e.g. `/var/log/app_engine/app.log;`
- In order to make the GKE work, we have to configure the readinessProbe, so we added a hacky `/health` endpoint/location in nginx config.
- In some cases, we want to enable a HTTPBasicAuth proxy, so we have to make the switch for auth basic optional in nginx config.

ui/assets/environments/environment.cromwell.caas.json
rexwangcc: Thanks for the catch, will update it!

servers/cromwell/README.md
rexwangcc: Oops, flipped order. Thanks for the catch, will fix the doc.

servers/cromwell/README.md
rexwangcc: Sure, will add it back.

ui/src/environments/environment.prod.ts
rexwangcc: Will do, thanks!

ui/src/environments/environment.prod.ts
```diff
@@ -1,4 +1,8 @@
...
 // when building the UI server as it contains deployment-specific details.
 
-export const environment = {}
+export const environment = {
+  apiUrl: '',
+  clientId: '',
```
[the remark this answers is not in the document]
rexwangcc: Unfortunately, if I remove the `apiUrl` and `clientId` here, all of the promises in the core will complain about missing fields during the `ng build --prod` process, which is the main reason that was preventing me from using `--prod` previously.

For the `dev` one you described here, I though the default `environment.ts` already satisfied?

ui/src/environments/environmentLoader.ts
```diff
@@ -0,0 +1,20 @@
+import { environment as defaultEnvironment } from './environment';
```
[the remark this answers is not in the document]
rexwangcc: Thanks for the suggestion, I like `ConfigurationLoader` here. Will refactor it.

ui/src/main.ts
```diff
@@ -3,10 +3,16 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
...
-  enableProdMode();
-}
 
-platformBrowserDynamic().bootstrapModule(AppModule)
-  .catch(err => console.log(err));
+environmentLoaderPromise.then(env => {
+  if (env.production) {
+    enableProdMode();
+  }
+  environment.apiUrl = env.apiUrl;
```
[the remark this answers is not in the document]
rexwangcc: Going to replace these assignments with a loop.

ui/src/main.ts
```diff
@@ -3,10 +3,16 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
 
 import {AppModule} from './app/app.module';
 import {environment} from './environments/environment';
+import { environmentLoader as environmentLoaderPromise } from './environments/environmentLoader';
 
-if (environment.production) {
```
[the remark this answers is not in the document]
rexwangcc: Agree on leaving this as it is. My question here is do we have any uses case that requires us to change the mode according to JSON configs? if so, we might want to change the order of `enableProdMode()` and the `configurationLoaderPromise.then()`

ui/src/environments/environmentLoader.ts
rexwangcc: Makes sense to me, I will try to add error handling for both the `xmlhttp.open()` and this loading process.

ui/src/environments/environmentLoader.ts
```diff
@@ -0,0 +1,20 @@
...
+export const environmentLoader = new Promise<any>( (resolve) => {
+
+  let xmlhttp = new XMLHttpRequest(),
+    method = 'GET',
+    url = './assets/environments/environment.json';
+
+  xmlhttp.open(method, url, true);
+
+  xmlhttp.onload = function() {
+    if (xmlhttp.status === 200) {
```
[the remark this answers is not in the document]
rexwangcc: By saying "Cache it in this file", do you mean saving it into an object not as a promise, or caching the whole JSON file? For the latter case, currently most browsers(like Chrome) already cache the JSON file for us, although I don't like to depend on browser behaviors. For the former case, could you elaborate a little bit about the advantages to going that way?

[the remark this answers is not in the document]
bfcrampton: I don't think we should depend on Chrome caching the loaded JSON file for us, this ensures we don't  have to read the file again after initial app load.

ui/src/environments/configurationLoader.ts
```diff
@@ -0,0 +1,28 @@
...
+    } else {
+      reject("err_msg") // log for debugging TODO: show err msg: fail to load the file
+    }
+
+  };
+
+  xmlhttp.onerror = event => {
+    reject("err_msg"); // log for debugging TODO: show err msg: fail to load the file
+  };
+
```
rexwangcc: Not really familiar with the error handling for xmlhttp, is this what you meant previously? @calbach

ui/src/main.ts
```diff
@@ -3,10 +3,30 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
...
+  })
+  .then((env) => {
+    Object.entries(env).forEach(([key, value]) => {
+      environment[key] = value;
+    });
+
+    platformBrowserDynamic().bootstrapModule(AppModule)
+      .catch(err => console.log(err));
+  });
+
```
rexwangcc: And here above @calbach

ui/src/main.ts
```diff
@@ -3,10 +3,30 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
...
+    //TODO: Stop the app from moving forward here and show error about retrieving the file (snackBar(err))
+  })
+  .then((envfile) => {
+    return JSON.parse(envfile);
+  })
+  .catch((SyntaxError) => {
+    console.log(SyntaxError); // TODO: should stop the app here and show error about parsing the content snackBar(SyntaxError)
+  })
+  .then((env) => {
+    Object.entries(env).forEach(([key, value]) => {
```
calbach: I would probably put this logic one level down in ConfigurationLoader. I don't think it should be returning  raw response text.

servers/cromwell/README.md
```diff
@@ -4,10 +4,10 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
```
kbergin: Will the information in this readme be sufficient for anyone to use the stock deployment to deploy their own job manager for use (not for development)?

rexwangcc: I think it is tricky for users to get a clear understanding of how to use/deploy JMUI based on this readme (combine with the UI readme), so it is still important to me to add a `deploy` folder to provide template configs and deployment facing readme. I could add a minimal doc along with this PR, but it makes sense to separate dsub and Cromwell deployment instructions into several PRs.

kbergin: Excellent, I'd love if that folder could exist with some basic information about what you've implemented here and placeholders for what we need to make. That'd be very helpful!

rexwangcc:
Note: This RFC PR is kinda premature, I have received a lot of useful comments and reviews on this PR.  Since the both the ACs and the main implementation design has been changed now, and moving forward on this PR requires not only massive refactoring but rebasing, I'm going to close this PR and open a new PR(https://github.com/DataBiosphere/job-manager/pull/299) for this topic.

rexwangcc:
Note: This RFC PR is kinda premature, I have received a lot of useful comments and reviews on this PR.  Since the both the ACs and the main implementation design has been changed now, and moving forward on this PR requires not only massive refactoring but rebasing, I'm going to close this PR and open a new PR(https://github.com/DataBiosphere/job-manager/pull/299) for this topic.

PR #287, opened 2018-05-30
Bulk change labels

rsasch:
see https://elastc.com/c/h*z2HQp4/163-bulk-change-labels


ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,58 @@ export class JobsTableComponent implements OnInit {
...
+          this.bulkLabelFields[f]['default'] = jobLabelValue;
+        } else if (this.bulkLabelFields[f]['default'] !== false &&
+          this.bulkLabelFields[f]['default'] != jobLabelValue) {
+          this.bulkLabelFields[f]['default'] = false;
+        }
+      }
+    }
+    let dialogRef = this.bulkChangeDialog.open(JobsBulkChangeComponent, {
+      width: '350px',
+      height: (this.bulkLabelFields.length * 150).toString() + 'px',
```
calbach: Do you need an explicit height? Can it not just fit to the contents?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,58 @@ export class JobsTableComponent implements OnInit {
...
+        if (this.bulkLabelFields[f]['default'] == null) {
+          this.bulkLabelFields[f]['default'] = jobLabelValue;
+        } else if (this.bulkLabelFields[f]['default'] !== false &&
+          this.bulkLabelFields[f]['default'] != jobLabelValue) {
+          this.bulkLabelFields[f]['default'] = false;
+        }
+      }
+    }
+    let dialogRef = this.bulkChangeDialog.open(JobsBulkChangeComponent, {
+      width: '350px',
```
calbach: Probably belongs in a style sheet(?)

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,58 @@ export class JobsTableComponent implements OnInit {
       }
     }
   }
+
+  openBulkChangeDialog(): void {
+    this.resetBulkLabelFieldDefaults();
+    // figure out default values for bulk fields; if all jobs have the same label value,
+    // set that to the defualt; otherwise, set default to boolean false
+    for (let j = 0; j < this.selection.selected.length; j++) {
+      for (let f = 0; f < this.bulkLabelFields.length; f++) {
```
calbach: Unless you explicitly need the index, I would iterate with "let ... of", i.e. `let f of this.bulkLabelFields`

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,58 @@ export class JobsTableComponent implements OnInit {
...
+        const label = this.bulkLabelFields[f].field.replace('labels.', '');
+        const jobLabelValue = this.selection.selected[j].labels[label] || '';
+        if (this.bulkLabelFields[f]['default'] == null) {
```
calbach: Probably can get some code reduction here:

```
let default = null;
let v = this.selection.selected[0].labels[label];
if (this.selection.selected.every(s => v === s.labels[label])) {
  default = v || '';
}
```

calbach: I'd recommend using `null` to indicate no default, then use a string for everything else (i.e. `""` even if null).

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,58 @@ export class JobsTableComponent implements OnInit {
...
+      }
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
+      if (result) {
+        for (let i = 0; i < result.jobs.length; i++) {
+          const req: UpdateJobLabelsRequest = { labels : {} };
+          Object.keys(result.fields).forEach((displayField) => {
+            const label = displayField.replace('labels.', '');
+            req.labels[label] = result.fields[displayField];
```
calbach: Should match the single edit code

ui/src/app/job-list/table/table.component.ts
rsasch: `null` means the default hasn't been set to anything yet, `false` means there's multiple values, so no default, `""` means the default has so far been set to an empty string

ui/src/app/job-list/job-list.module.ts
```diff
@@ -50,8 +53,12 @@ import {SharedModule} from '../shared/shared.module';
     RouterModule,
     SharedModule,
   ],
+  entryComponents: [
+    JobsBulkChangeComponent
```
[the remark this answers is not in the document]
rsasch: changed

ui/src/app/job-list/table/bulk-change/bulk-change.component.html
```diff
@@ -0,0 +1,43 @@
+<h2 mat-dialog-title>Set Metadata</h2>
+<mat-dialog-content>
+  <ng-container *ngFor="let df of data.bulkLabelFields">
+    <div [ngSwitch]="getFieldType(df)">
+      <div *ngSwitchCase="'text'">
+        <label for="{{ df.field }}">{{df.display}}</label>
+        <mat-form-field>
+                  <input matInput
```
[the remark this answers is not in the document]
rsasch: done

ui/src/app/job-list/table/bulk-change/bulk-change.component.html
```diff
@@ -0,0 +1,43 @@
...
+            <clr-icon shape="times-circle"></clr-icon>
+          </button>
+          <mat-hint align="end">{{bulkLabelField.value?.length || 0}}/255</mat-hint>
+        </mat-form-field>
+      </div>
+      <div *ngSwitchCase="'list'">
+        <label for="{{ df.field }}">{{ df.display }}</label>
+        <mat-form-field>
+          <mat-select #bulkLabelSelect
+                      [placeholder]="(df.default !== false) ? df.default : '(multiple)'"
```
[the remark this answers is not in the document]
rsasch: re-structured

ui/src/app/job-list/table/table.component.ts
rsasch: done

ui/src/app/job-list/table/bulk-edit/bulk-edit.component.html
```diff
@@ -0,0 +1,43 @@
...
+    <div [ngSwitch]="getFieldType(dfItem.displayField)">
+      <div *ngSwitchCase="'text'">
+        <label for="{{ dfItem.displayField.field }}">{{dfItem.displayField.display}}</label>
+        <mat-form-field>
+          <input matInput
+                 type="text"
+                 [value]="(dfItem.default !== false) ? dfItem.default : '(multiple)'"
+                 [id]="dfItem.displayField.field"
+                 #bulkLabelField
+                 maxlength="255"
```
samanehsan: How about replacing this number with a variable defined in the component since it's used in a couple of places in this template?

servers/cromwell/README.md
```diff
@@ -53,7 +54,8 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
      "field": "labels.label",
      "display": "Label",
      "fieldType": "text",
-     "editable": true
+     "editable": true,
+     "bulkEditable": true
```
samanehsan: The last item in `displayFields` here is missing a closing `}`

rsasch: thanks catching that – fixed

servers/cromwell/README.md
```diff
@@ -40,26 +40,29 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
-     "field": "labels.comment",
-     "display": "Comment",
-     "fieldType": "text",
-     "editable": true
+      "field": "labels.flag",
+      "display": "Flag",
+      "editable": true,
+      "bulkEditable": true,
+      "fieldType": "list",
+      "fieldValues": [
```
samanehsan: Please change this to `validFieldValues`

rsasch: done – thanks for catching that!

ui/src/app/job-list/table/bulk-edit/bulk-edit.component.html
```diff
@@ -0,0 +1,43 @@
+<h2 mat-dialog-title>Set Metadata</h2>
+<mat-dialog-content>
+  <ng-container *ngFor="let dfItem of data.bulkLabelFields">
+    <div [ngSwitch]="getFieldType(dfItem.displayField)">
+      <div *ngSwitchCase="'text'">
+        <label for="{{ dfItem.displayField.field }}">{{dfItem.displayField.display}}</label>
+        <mat-form-field>
+          <input matInput
+                 type="text"
+                 [value]="(dfItem.default !== false) ? dfItem.default : '(multiple)'"
```
samanehsan: You could extract this logic into a function in the component, since it gets used in a couple of places here.

samanehsan:
This looks great @rsasch! I left a few comments on your PR ☀️

ui/src/app/job-list/table/table.component.ts
calbach: My issue is that you have a value of type `boolean|string|null`, which is pretty confusing in my opinion. When you're rendering the dialog, the only thing you care about is "default value" or "(multi)". It seems more natural to me to express (multi) as null here.

calbach: (also: see my other comment suggesting how to rework this code block - I don't think using null to track the value is necessary)

ui/src/app/job-list/table/table.component.ts
```diff
@@ -41,6 +42,7 @@ export class JobsTableComponent implements OnInit {
   private mouseoverJob: QueryJobsResult;
 
   public displayFields: DisplayField[];
+  public bulkLabelFields;
```
calbach: Please use a type:

```
...: Array<{
  default: string|null,
  field: DisplayField
}>;
```

After this change, please fix all accesses to use normal property name access instead of using strings. You should only need to use strings to access properties if the property name is actually dynamic (e.g. for labels).

ui/src/app/job-list/table/bulk-edit/bulk-edit.component.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  getFieldType(df: DisplayField): string {
+    return df.fieldType.toString();
+  }
+
+  getFieldOptions(df: DisplayField): string[] {
+    return df.validFieldValues;
+  }
+
+  getFieldDefaultValue(displayFieldItem: {'default'}): string {
```
calbach: I don't understand this type signature, what is this saying?

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,59 @@ export class JobsTableComponent implements OnInit {
...
+        const jobLabelValue = job.labels[label] || '';
+        if (bulkFieldItem.default === null) {
+          bulkFieldItem.default = jobLabelValue;
+        } else if (bulkFieldItem.default !== false && bulkFieldItem.default !== jobLabelValue) {
+          bulkFieldItem.default = false;
+        }
+      }
+    }
+    let dialogRef = this.bulkEditDialog.open(JobsBulkEditComponent, {
+      disableClose: true,
```
calbach: Is there another way to cancel the bulk edit? I don't see this in the dialog

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +261,59 @@ export class JobsTableComponent implements OnInit {
...
+      data: {
+        'bulkLabelFields' : this.bulkLabelFields,
+        'selectedJobs' :  this.selection.selected,
+        'newValues': []
+      }
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
+      if (result) {
```
calbach: nit: invert and return early to avoid indenting the whole function

ui/src/app/job-list/table/table.component.ts
rsasch: I understand your issue with `false` being a possible value for default value. Your suggested code is definitely much more compact, but I also find it a little hard to read through, so I'm going to poke at it for a bit.

ui/src/app/job-list/table/table.component.ts
calbach: What happens if the user cancels or doesn't change any values?

ui/src/app/job-list/table/table.component.ts
rsasch: There is a 'Cancel' and an 'Apply' button

ui/src/app/job-list/table/table.component.ts
rsasch: if the user cancels, then `result` will be empty because `saveChanges()` will not have been called. If there are no changes, then `result.fields` will be empty, so I'll add in a check for that.

ui/src/app/job-list/table/table.component.ts
calbach: That is one possible bug that could result here (false being a default value), but in general I would recommend against using union types like this anywhere. It makes for more confusing code IMO.

No need to use my exact code, but I don't think using null as a third state is doing anything for you currently. You could just as easily directly compare against the label value of the first job. Alternatively initialize defaultValue to the first label value, iterate through, and if something disagrees set defaultValue to null and break.

ui/src/app/job-list/table/bulk-edit/bulk-edit.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  getFieldType(df: DisplayField): string {
+    return df.fieldType.toString();
+  }
+
+  getFieldOptions(df: DisplayField): string[] {
+    return df.validFieldValues;
+  }
+
+  getFieldDefaultValue(displayFieldItem: {
```
calbach: Since you use this type in multiple places, probably makes sense to define this as an interface

ui/src/app/job-list/table/bulk-edit/bulk-edit.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    default: string|null,
+    displayField: DisplayField }): string {
+    if (displayFieldItem.default !== null) {
+      return displayFieldItem.default
+    } else {
```
calbach: opt_nit: I would just drop the "else" to reduce indentation

ui/src/app/job-list/table/table.component.spec.ts
```diff
@@ -265,6 +292,21 @@ describe('JobsTableComponent', () => {
     expect(isGroupAbortIsEnabled()).toBeTruthy();
   }))
 
+  it('disables the bulk edit button when there is a bulkEditable field but less than two jobs selected', async(() => {
```
calbach: grammar nit: fewer

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +264,56 @@ export class JobsTableComponent implements OnInit {
...
+      for (let job of result.jobs) {
+        this.jobManagerService.updateJobLabels(job.id,
+            this.prepareUpdateJobLabelsRequest(result.fields))
+          .then((response: UpdateJobLabelsResponse) => {
+            job.labels = response.labels;
+            lastJobUpdated = job;
+          })
+          .catch((error) => this.handleError(error));
+      }
+      this.onJobsChanged.emit(lastJobUpdated);
```
calbach: onJobsChanged is already supposed to emit an array of jobs, rather than a single job (see the type). This should just emit the entire array of jobs.

calbach: At this point in the code, we just sent a bunch of async XHRs to update job labels, but they haven't returned yet. We should only be emitting a change event once the calls actually complete.

Accumulating the Promises via `Promise.all()` may be what you're looking for.

ui/src/app/job-list/table/table.component.spec.ts
calbach: I don't think I agree with this >=2 behavior. I don't see any advantage to disabling bulk edits for a single job. Agreed, they could do it more quickly by just selecting the specific job directly, but maybe they haven't discovered that particular interaction or they don't want to think about how many they selected.

ui/src/app/job-list/table/table.component.ts
```diff
@@ -249,4 +262,58 @@ export class JobsTableComponent implements OnInit {
...
+      }
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
+      if (!result || !this.doesResultHaveLabelChanges(result)) {
+        return;
+      }
+      const jobUpdates: Promise<void>[] = [];
+      for (let job of result.jobs) {
+        jobUpdates.push(this.jobManagerService.updateJobLabels(job.id,
```
calbach: opt: Another way of writing this:

```
const jobUpdates = result.jobs.map((job) => {
  updateJobLabels(job)...
}
```

api/jobs.yaml
```diff
@@ -483,6 +486,15 @@ definitions:
     enum:
       - text
       - list
+  BulkLabelField:
+    description: a DisplayField that can be bulk-editable via the UI
+    required:
+      - field
```
samanehsan: Should this be `displayField`?

rsasch: yes...thanks – good catch!

api/jobs.yaml
```diff
@@ -483,6 +486,15 @@ definitions:
     enum:
       - text
       - list
+  BulkLabelField:
+    description: a DisplayField that can be bulk-editable via the UI
+    required:
+      - displayField
```
samanehsan: It looks like you'll need to re-generate the models as well!

rsasch: done

PR #293, opened 2018-06-06
made rules for including 'fieldType' in label config more accurate

rsasch:


servers/cromwell/README.md
```diff
@@ -72,7 +72,7 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
   "queryExtensions": []
 }
 ```
-The default "fieldType" is `text` and both "editable" and "bulkEditable" will treated as `false` unless explicity set to `true`.
+Both "editable" and "bulkEditable" will treated as `false` unless explicity set to `true`; if the field is editable, then "fieldType" is required.
```
rexwangcc: will *be* treated?

rsasch: yup – thanks for the catch – fixed

PR #294, opened 2018-06-07
fix Cromwell bug where "of many" shows up when there's only one page

rsasch:
- aligned return from query_jobs in cromwell to query_jobs in dsub when it comes to page_token

see: https://elastc.com/c/C46fjeNX/320-bug-when-there-s-only-one-page-it-still-says-of-many

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -206,7 +206,7 @@ def query_jobs(body, **kwargs):
         offset = offset + page_size
         page = page_from_offset(offset, page_size)
 
-    if query_page_size > len(results):
+    if query_page_size >= len(results):
         return QueryJobsResponse(results=results)
 
     next_page_token = page_tokens.encode_offset(offset)
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -206,6 +206,9 @@ def query_jobs(body, **kwargs):
         offset = offset + page_size
         page = page_from_offset(offset, page_size)
 
+    if query_page_size > len(results):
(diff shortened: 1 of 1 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -206,6 +206,9 @@ def query_jobs(body, **kwargs):
         offset = offset + page_size
         page = page_from_offset(offset, page_size)
 
+    if query_page_size > len(results):
```
calbach: This still mishandles one edge case which is where the last page is exactly the length of the page size. You don't want to return a next_page_token in that case.

I'm not familiar with this particular code, but typically you want to try and collect query_page_size+1 elements, then truncate to page_size on return, populating next_page_token only if you see the n+1th element.

calbach: The above loop breaks at query_page_size, so now there's a new bug which is that you might not return a page token when there are still more jobs. See the second part of my above comment.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -206,6 +206,9 @@ def query_jobs(body, **kwargs):
         offset = offset + page_size
         page = page_from_offset(offset, page_size)
 
+    if query_page_size >= total_results:
```
calbach: From looking above, total_results is the total number of all workflows matching the given query (unpaginated). If so, you don't want that.

Is there a test for this code that you can iterate against?

rsasch:
removing this PR because this fix is blocked, waiting for update to Cromwell

calbach:
OK, not sure what Cromwell update you're referring to (maybe subworkflows?), but FWIW there are clearly bugs here specific to the Cromwell shim implementation here.

PR #295, opened 2018-06-07
Empty dashboard wired into routing

zacharyhsia:
Add dashboard page to routing and set a flag in environment files.

### DIFF
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -38,6 +38,7 @@ const routes: Routes = [
   {
     path: 'dashboard',
     component: DashboardComponent,
+    //TODO: (zach) dashboard need a query param of project id before it can be activated
     canActivate: [CapabilitiesActivator]
   },
   {
--- a/ui/src/app/app-routing.module.ts
+++ b/ui/src/app/app-routing.module.ts
@@ -15,6 +15,7 @@ import {ProjectsComponent} from './projects/projects.component'
 import {RouteReuse} from './route-reuse.service';
 
 import {environment} from '../environments/environment';
+import {DashboardComponent} from "./dashboard/dashboard.component";
 
 // Based on the URL mapping in "routes" below, the RouterModule attaches
 // UI Components to the <router-outlet> element in the main AppComponent.
@@ -32,6 +33,11 @@ const routes: Routes = [
   {
     path: 'projects',
     component: ProjectsComponent,
+    canActivate: [CapabilitiesActivator]
+  },
+  {
+    path: 'dashboard',
+    component: DashboardComponent,
     canActivate: [CapabilitiesActivator]
   },
   {
--- a/ui/src/app/app.module.ts
+++ b/ui/src/app/app.module.ts
@@ -11,7 +11,6 @@ import {JobDetailsModule} from './job-details/job-details.module';
 import {JobListModule} from './job-list/job-list.module';
 import {SignInModule} from './sign-in/sign-in.module';
 import {ProjectsModule} from './projects/projects.module';
-import { DashboardComponent } from './dashboard/dashboard.component';
 import {DashboardModule} from "./dashboard/dashboard.module";
 
 @NgModule({
--- a/ui/src/app/app.module.ts
+++ b/ui/src/app/app.module.ts
@@ -11,6 +11,8 @@ import {JobDetailsModule} from './job-details/job-details.module';
 import {JobListModule} from './job-list/job-list.module';
 import {SignInModule} from './sign-in/sign-in.module';
 import {ProjectsModule} from './projects/projects.module';
+import { DashboardComponent } from './dashboard/dashboard.component';
+import {DashboardModule} from "./dashboard/dashboard.module";
 
 @NgModule({
   imports: [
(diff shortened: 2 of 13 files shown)


ui/src/app/app.module.ts
```diff
@@ -11,6 +11,8 @@ import {JobDetailsModule} from './job-details/job-details.module';
 import {JobListModule} from './job-list/job-list.module';
 import {SignInModule} from './sign-in/sign-in.module';
 import {ProjectsModule} from './projects/projects.module';
+import { DashboardComponent } from './dashboard/dashboard.component';
```
calbach: nit: drop the surrounding spaces to match the rest of the style

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe('DashboardComponent', () => {
+  let component: DashboardComponent;
+  let fixture: ComponentFixture<DashboardComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ DashboardComponent ]
+    })
+    .compileComponents();
```
calbach: Generally should indent for continuations like this. Alternatively in this case I would probably put it on the previous line

zacharyhsia:
Please review the PR. Thanks!

ui/src/app/projects/projects.component.ts
```diff
@@ -94,6 +95,10 @@ export class ProjectsComponent implements OnInit {
 
   navigateJobs() {
     let extras = {queryParams: {q: URLSearchParamsUtils.encodeURLSearchParams({extensions: {projectId: this.projectsControl.value}})}};
-    this.router.navigate(['jobs'], extras);
+    if (environment.dashboardEnabled) {
+      this.router.navigate(['dashboard'], extras);
```
calbach: Have you thought about the URL scheme? I'm thinking we may just want `projectId=<project>`

ui/src/app/app-routing.module.ts
```diff
@@ -34,6 +35,11 @@ const routes: Routes = [
     component: ProjectsComponent,
     canActivate: [CapabilitiesActivator]
   },
+  {
+    path: 'dashboard',
+    component: DashboardComponent,
+    canActivate: [CapabilitiesActivator]
```
calbach: One thing we'll need to handle in the CapabilitiesActivator: the dsub dashboard page will require a project (probably via query parameter). If it's not there, we'd want to redirect to the projects page (similar to how we're redirecting from "jobs" -> "projects" if there is no projectId in the query).

Feel free to handle this in this PR, or else just put a TODO in the CapabilitiesActivator

ui/src/app/projects/projects.component.ts
zacharyhsia: I thought we can parse that 'extra' thing when handling the navigation. It is not that big.

calbach: I think the only filter we would be supporting here is projectId (and possibly a query param for the time range), so I wouldn't support the more generic "?q=<encoded_search_query>" that we have on the job list page.

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -6,7 +6,7 @@ import {JobStatus} from "../model/JobStatus";
 import {FieldDataType, queryDataTypes, queryExtensionsDataTypes} from "../common";
 
 /** Utilities for working with URLSearchParams*/
-export class URLSearchParamsUtils {
+  export class URLSearchParamsUtils {
```
calbach: Revert

ui/src/app/projects/projects.component.ts
```diff
@@ -93,7 +94,13 @@ export class ProjectsComponent implements OnInit {
   }
 
   navigateJobs() {
-    let extras = {queryParams: {q: URLSearchParamsUtils.encodeURLSearchParams({extensions: {projectId: this.projectsControl.value}})}};
-    this.router.navigate(['jobs'], extras);
+    let jobsExtras = {queryParams: {q: URLSearchParamsUtils.encodeURLSearchParams({extensions: {projectId: this.projectsControl.value}})}};
+    let dashboardExtras = {queryParams: {projectId: this.projectsControl.value}};
```
calbach: nit: I would prefer inlining these below or else defining the variables right before they are used.

calbach: opt: existing issue, but these should really be const's

zacharyhsia:
I added another extra with only projectId as param. 
How can I add multiple param? 
Here is what I found on angular doc,
```javascript
interface NavigationExtras {
  relativeTo?: ActivatedRoute | null
  queryParams?: Params | null
  fragment?: string
  preserveQueryParams?: boolean
  queryParamsHandling?: QueryParamsHandling | null
  preserveFragment?: boolean
  skipLocationChange?: boolean
  replaceUrl?: boolean
}
```

```javascript
type Params = {
    [key: string]: any;
};
```

I am confused by `[key: string]: any;`, I thought the following code would fit in but it dose not.
```javascript
let dashboardExtras = {queryParams: [{projectId: this.projectsControl.value}, {anotherField: 'value'}]};
```

How to interpret `[key: string]: any;`

rexwangcc:
@calbach Thanks for the information here!

PR #297, opened 2018-06-13
API modified based on Bryan's design

zacharyhsia:
Hi! I modified the API based on [Bryan's design](https://docs.google.com/document/d/1zDCUiZvIYp7R03-ZMG_xJWIsLXTmRV76jH576UIS4QE/edit). 

As discussed with CH, for this model:
```
StatusCounts {
	// A mapping of JobStatus to number of jobs with that status. If a 
	// status is not present, its job count is assumed 0.
	counts: {JobStatus : int}
}
```
I am not sure how we can use JobStatus as the key and I left it as an **Object**. See `job.yaml` line 311.

Please help improve the change. Thanks.


api/jobs.yaml
```diff
@@ -130,6 +130,37 @@ paths:
...
         - Jobs
+  '/aggregations':
+    get:
+      operationId: GetJobAggregation
+      summary: Query for aggregated jobs in the given time frame
+      parameters:
+        - name: timeFrame
+          description: Time Frame
+          required: true
+          type: string
```
bfcrampton: This should reference the `TimeFrame` definition below:
```
- name: timeFrame
  description: Time frame to aggregate over
  required: true
  in: query
  schema:
    $ref: '#/definitions/TimeFrame'
```

api/jobs.yaml
```diff
@@ -187,6 +218,11 @@ definitions:
         description: Fields on ExtendedQueryFields which are queryable
         items:
           type: string
+      aggregationTimeFrames:
+        type: array
+        description: Time frames for jobs information agggregation
```
bfcrampton: nit: Time frames supported by the /aggregations endpoint.

api/jobs.yaml
```diff
@@ -244,6 +280,46 @@ definitions:
       labels:
         type: object
         description: All job labels, after the update has been applied.
+  AggregationResponse:
+    description: >
+      Returns aggregated jobs with specified time range or label.
```
bfcrampton: nit: Something along the lines of:
"Jobs aggregated by label and status within the specified time frame"

api/jobs.yaml
```diff
@@ -244,6 +280,46 @@ definitions:
...
+        type: string
+      values:
+        properties:
+          key:
+            type: string
+          value:
+              $ref: '#/definitions/StatusCounts'
+  StatusCounts:
+    description: >
+      A mapping of JobStatus to number of jobs with that status. If a
```
bfcrampton: tiny nit: A mapping of JobStatuses to the number of jobs with those statuses

api/jobs.yaml
```diff
@@ -244,6 +280,46 @@ definitions:
...
+          $ref: '#/definitions/Aggregation'
+  Aggregation:
+    description: Aggregated jobs specified by the keys.
+    properties:
+      name:
+        type: string
+      key:
+        type: string
+      values:
+        properties:
```
bfcrampton: This implicitly generates another definition called `AggregationValues`. I'd prefer this to be expressed explicitly so its clearer from looking at the swagger definition what the relevant models are. I would also take the same approach to `StatusCounts` below. It would look something like:

```
  Aggregation:
    description: |
      Aggregation of jobs by label values and statuses, within the specified
      time frame.
    properties:
      name:
        type: string
      key:
        type: string
      entries:
        type: array
        items:
          $ref: '#/definitions/AggregationEntry'
  AggregationEntry:
    description: |
      Aggregation of jobs by statuses for a given label, within a specified time
      frame.
    properties:
      label:
        type: string
        description: The label value for this aggregation entry.
      status_counts:
        $ref: '#/definitions/StatusCounts'
  StatusCounts:
    description: >
      A mapping of JobStatuses to number of jobs with those statuses. If a
      status is not present, its job count is assumed 0.
    properties:
      counts:
        type: array
        items:
          $ref: '#/definitions/StatusCount'
  StatusCount:
    description: A mapping of JobStatus to number of jobs with that status.
    properties:
      status:
        $ref: '#/definitions/JobStatus'
      count:
        type: integer
```

api/jobs.yaml
zacharyhsia: Seems that if we use the TimeFrame schema, we have to put the parameter to request body, is that what we want?

api/jobs.yaml
```diff
@@ -187,6 +218,11 @@ definitions:
         description: Fields on ExtendedQueryFields which are queryable
         items:
           type: string
+      aggregationTimeFrames:
```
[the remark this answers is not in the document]
bfcrampton: It wasn't in the initial design but I think that sounds like a good idea, yeah.

api/jobs.yaml
calbach: I think it may be possible, see "Reusable Enums" here: https://swagger.io/docs/specification/2-0/enums

api/jobs.yaml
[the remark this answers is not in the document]
calbach: OK, probably not necessary for now - perhaps we can add this later if needed. (specific use case I had in mind is if we wanted to show the page before we fully loaded the data, then we could at least prepopulate the card headers and know the number of cards).

api/jobs.yaml
```diff
@@ -8,6 +8,14 @@ info:
   version: '0.0.1'
 produces:
   - application/json
+x-types:
```
calbach: What is this? It doesn't seem to actually be in use

zacharyhsia:
Ready for code review. Thanks!

servers/cromwell/jobs/controllers/aggregation_controller.py
```diff
@@ -0,0 +1,21 @@
...
+    Query for aggregated jobs in the given time frame.
+    
+    :param timeFrame: Time Frame to aggregate over.
+    :type timeFrame: str
+    :param projectId: The ID of the project to query.
+    :type projectId: str
+
+    :rtype: AggregationResponse
+    """
+    # return 'do some magic!'
```
calbach: nit: remove this comment, here and in the other controller

servers/cromwell/jobs/controllers/aggregation_controller.py
```diff
@@ -0,0 +1,21 @@
...
+from jobs.models.aggregation_response import AggregationResponse
+from datetime import date, datetime
+from typing import List, Dict
+from six import iteritems
+from ..util import deserialize_date, deserialize_datetime
+
+
+def get_job_aggregations(timeFrame, projectId=None):
+    """
+    Query for aggregated jobs in the given time frame.
```
calbach: nit: We use a different comment syntax - please update: https://github.com/DataBiosphere/job-manager/blob/master/servers/dsub/jobs/controllers/jobs_controller.py#L76

servers/cromwell/jobs/controllers/aggregation_controller.py
```diff
@@ -0,0 +1,21 @@
...
+    
+    :param timeFrame: Time Frame to aggregate over.
+    :type timeFrame: str
+    :param projectId: The ID of the project to query.
+    :type projectId: str
+
+    :rtype: AggregationResponse
+    """
+    # return 'do some magic!'
+    raise NotImplemented('function not implemented yet.')
```
calbach: This exception type isn't imported:
See for example this import https://github.com/DataBiosphere/job-manager/blob/master/servers/dsub/jobs/controllers/jobs_controller.py#L12

api/jobs.yaml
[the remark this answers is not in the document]
zacharyhsia: No additional field added to `/capabilities` right now.

servers/cromwell/jobs/controllers/aggregation_controller.py
zacharyhsia: Style fixed.

PR #299, opened 2018-06-14
Stock Deployment for Job Manager 

rexwangcc:
This is derived from https://github.com/DataBiosphere/job-manager/pull/286 which was premature.

This PR:
1. Implements an environment loader that loads the config file on the UI side. This service also caches the result and use it throughout the application, which is suggested by @bfcrampton 
2. Replaces all of the `environment.xxx` with `this.environmentConfigurationLoaderService.getEnvironmentConfigSynchronous()['xxx']`
3. Cleans up all environment specific ts files.
4. Updates the compose file, angular config, and documentation.
5. Adds environment JSON files to assets folder.
6. Standardizes Cromwell environments into 2 categories: instance and caas

There will be a followup PR after this that get merged adds `deploy` folder to the repo with documentation about various deployment choices.

Future works for this PR:
- [ ] Performs another round of benchmarking based on the new implementation.
- [x] Adds test cases once the implementation becomes stable.
<del>- [ ] Use snackBar to show the error message.<del>
- [x] Rebase onto the latest master branch.


ui/src/app/core/capabilities.service.ts
```diff
@@ -12,7 +12,8 @@ export class CapabilitiesService {
   private capabilitiesResponse: CapabilitiesResponse;
   private capabilitiesResponsePromise: Promise<CapabilitiesResponse>;
 
-  constructor(private http: Http) {}
+  constructor(private http: Http,
+              private readonly environmentConfigurationLoaderService:EnvironmentConfigurationLoaderService) {}
```
bfcrampton: nit: private?

ui/src/app/core/job-manager.service.ts
```diff
@@ -18,7 +18,8 @@ export class JobManagerService {
   private static readonly defaultErrorDetail = "An unknown error has ocurred. Please try again later.";
   private static readonly defaultErrorTitle = "Unknown";
 
-  constructor(private readonly authService: AuthService, private http: Http) {}
+  constructor(private readonly authService: AuthService, private http: Http,
+              private readonly environmentConfigurationLoaderService:EnvironmentConfigurationLoaderService) {}
```
bfcrampton: nit: private?

ui/src/app/core/job-manager.service.ts
```diff
@@ -97,7 +98,7 @@ export class JobManagerService {
   }
 
   abortJob(id: string): Promise<void> {
-    return this.http.post(`${environment.apiUrl}/jobs/${id}/abort`,
+    return this.http.post(`${this.environmentConfigurationLoaderService.getEnvironmentConfigSynchronous()['apiUrl']}/jobs/${id}/abort`,
```
bfcrampton: nit: extract to a local variable

ui/src/app/core/job-manager.service.ts
```diff
@@ -106,7 +107,7 @@ export class JobManagerService {
   }
 
   updateJobLabels(id: string, req: UpdateJobLabelsRequest): Promise<UpdateJobLabelsResponse> {
-    return this.http.post(`${environment.apiUrl}/jobs/${id}/updateLabels`,
+    return this.http.post(`${this.environmentConfigurationLoaderService.getEnvironmentConfigSynchronous()['apiUrl']}/jobs/${id}/updateLabels`,
```
bfcrampton: same here and below local variable

ui/src/environments/environment-configuration-loader.service.ts
```diff
@@ -0,0 +1,59 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class EnvironmentConfigurationLoaderService {
```
bfcrampton: I would probably just call this the ConfigLoaderService, this name is super long and it indicates that this is completely separate from the environment file now.

ui/src/environments/environment-configuration-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+  private environmentConfig: object;
```
bfcrampton: nit: any to be consistent the Promise value

ui/src/environments/environment-configuration-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          if (xmr.status === 200) {
+            EnvironmentConfigurationLoaderService.loadJSONAsync(xmr.responseText)
+              .catch(SyntaxError => console.log(SyntaxError))
+              .then(parsedConfig => resolve(parsedConfig));
+          } else {
+            reject({
+              message: "Cannot find the environment configuration file",
+              status: xmr.status,
+              statusText: xmr.statusText
+            }) //Rex TODO: show error msg with snackBar
```
bfcrampton: tiny nit: newline and we usually have the format `TODO(name):  ...`

ui/src/main.ts
```diff
@@ -1,12 +1,27 @@
...
+import {EnvironmentConfigurationLoaderService} from "./environments/environment-configuration-loader.service";
 
 if (environment.production) {
   enableProdMode();
 }
 
-platformBrowserDynamic().bootstrapModule(AppModule)
-  .catch(err => console.log(err));
+let environmentConfigurationLoader = ReflectiveInjector
+  .resolveAndCreate([EnvironmentConfigurationLoaderService])
```
bfcrampton: Do you need to use a ReflectiveInjector here? Why can't you just do `new EnvironmentConfigurationLoaderService()`?

ui/src/environments/environment-configuration-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          this.environmentConfig = parsedConfig;
+          return this.environmentConfig;
+        });
+    }
+    return this.environmentConfigPromise;
+  }
+
+  getEnvironmentConfigSynchronous(): object {
+    if (this.environmentConfig) {
+      console.log("Yahoo, hit the cache!!");
```
rexwangcc: I intentionally left the debug messages there for benchmarking, will remove them later.

ui/src/main.ts
```diff
@@ -1,12 +1,27 @@
...
+  .get(EnvironmentConfigurationLoaderService)
+
+environmentConfigurationLoader.getEnvironmentConfig()
+  .then(environmentConfig => {
+    console.log("The loaded env config is: ");
+    console.log(environmentConfig);
+    console.log("The env config stored in the singleton is: ");
+    console.log(environmentConfigurationLoader.getEnvironmentConfigSynchronous());
```
rexwangcc: Intentionally left them there.

ui/src/environments/environment-configuration-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+@Injectable()
+export class EnvironmentConfigurationLoaderService {
+  private environmentConfig: object;
+  private environmentConfigPromise: Promise<any>;
+
+  constructor() { }
+
+  private static loadJSONAsync(rawText:string): Promise<object> {
+    return Promise.resolve(JSON.parse(rawText));
+  }
```
rexwangcc: I'm not sure if I abused using Promises in this file, new to TS.

ui/src/environments/environment-configuration-loader.service.ts
rexwangcc: Makes sense to me, will do!

ui/src/main.ts
rexwangcc: From the Angular docs, it suggests to avoid using `new` for services as much as possible. Although `ReflectiveInjector` does provide advanced features in this case, it resolves dependencies between services when it's getting more complicated, based on my reading on the docs. So I personally prefer to avoid `new` here. 

However, as I did more research on this now, I should have used `StaticInjector` here because `ReflectiveInjector` is deprecated! Thanks for cathing this.

ui/src/app/core/job-manager.service.ts
rexwangcc: What do you think about declaring an instance variable e.g. `apiUrl`and assign it with `this.environmentConfigurationLoaderService.getEnvironmentConfigSynchronous()['apiUrl']` so that all of the methods can share the value?

ui/src/main.ts
```diff
@@ -3,10 +3,20 @@ import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
...
-platformBrowserDynamic().bootstrapModule(AppModule)
-  .catch(err => console.log(err));
+const configLoader = new ConfigLoaderService();
+
+configLoader.getEnvironmentConfig()
+  //TODO(Rex): Show err with snackBar(optional) and stop app peacefully
+  .catch(err => console.log(err))
+  .then(() => {
+    platformBrowserDynamic([{
+      provide: ConfigLoaderService, useValue: configLoader
```
samanehsan: What is the difference between including `ConfigLoaderService` as a provider here instead of in app.module.ts?

servers/cromwell/README.md
```diff
@@ -4,10 +4,10 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
 
 ## Development
 
-1. Set the `CROMWELL_URL` environment variable to specify which cromwell instance to use. Job Manager pulls workflow data from `https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1` by default:
-
+1. Set the `CROMWELL_URL` environment variable to specify which cromwell instance to use, for example:
+    
     ```
-    export CROMWELL_URL=https://cromwell.test-cromwell.broadinstitute.org/api/workflows/v1
+    export CROMWELL_URL=https://exapmle-cromwell.broadinstitute.org/api/workflows/v1
```
samanehsan: nit: small typo in "example"

servers/cromwell/README.md
```diff
@@ -38,45 +38,52 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
   "commonLabels": [
     "cromwell-workflow-id",
-    "workflow-name",
-    "flag"
+    "workflow-name"
   ],
   "queryExtensions": []
 }
 ```
-Both "editable" and "bulkEditable" will be treated as `false` unless explicity set to `true`; if the field is editable, then "fieldType" is required.
```
samanehsan: I think we should still include an example of editable/bulkEditable fields somewhere in the readme

servers/cromwell/README.md
```diff
@@ -23,7 +23,7 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     }
     ```
 
-3. (Optional) If you want to change the view of UI to some extent, i.e. display more columns in the job list view, such as labels of the jobs, you can add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations. The `capabilities_config.json` should **strictly** follow the following structure:
+3. (Optional) If you want to change the either the predefined view or the default behavior of UI to some extent, e.g. display more columns in the job list view such as labels of the jobs, or make more columns to be popped up by the query builder, you can add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations. The `capabilities_config.json` should **strictly** follow the following structure:
```
samanehsan: nit: please remove the extra "the" in front of "either the predefined view"

ui/src/main.ts
rexwangcc: Good question! Normally we put the service class name in the `app.module.ts` so that the provider will be responsible for creating singleton/instantiating the service. But in this case, we want the LoaderService's promise to be executed before all other services' initialization, which means we have to instantiate the service singleton by ourselves, make it store the environment variables which are previously stored in envrionments.ts files, and pass it to the module providers. So providers will try to use this specific singleton (by `useValue`) instead of creating an empty new one. I tried also include the class in `app.module.ts` at the same time, but it turned out that will shadow this one. What do you think?

servers/cromwell/README.md
rexwangcc: Oops, thanks for the catch!

servers/cromwell/README.md
rexwangcc: Oops, something wrong during the rebasing I think, good catch!

ui/src/environments/config-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor() { }
+
+  private static loadJSONAsync(rawText:string): Promise<object> {
+    return Promise.resolve(JSON.parse(rawText));
+  }
+
+  getEnvironmentConfig(): Promise<any> {
+    if (!this.environmentConfigPromise) {
+      this.environmentConfigPromise = new Promise<any>((resolve, reject) => {
```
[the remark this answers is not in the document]
rexwangcc: Sure, the earlier return will save a lot of time in that case, will do!

ui/src/main.ts
bfcrampton: I think that makes sense and this is the right place to put it 👍

samanehsan: Thank you for the explanation! I think the way you have it now makes sense.

ui/src/environments/config-loader.service.ts
```diff
@@ -0,0 +1,59 @@
...
+        const xmr = new XMLHttpRequest(),
+          method = 'GET',
+          url = '../../../assets/environments/environment.json';
+
+        xmr.open(method, url, true);
+
+        xmr.onload = () => {
+          if (xmr.status === 200) {
+            ConfigLoaderService.loadJSONAsync(xmr.responseText)
+              .catch(SyntaxError => console.log(SyntaxError))
```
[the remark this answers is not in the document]
rexwangcc: It's a surprise that when I was trying to take it out, the outer promise cannot catch the SyntaxError raised by the `JSON.parse()`. The current way here is not working either. As it's the latest commit, I have to use `return Promise.resolve(rawText).then(JSON.parse)` and keep this inner `catch` here to catch the SyntaxError properly. 

This observation also makes me think if I have abused using Promise, and should have used the callback styled simple `try..catch(SyntaxError)` for `JSON.parse()`here. What do you think? @bfcrampton

servers/cromwell/README.md
```diff
@@ -63,21 +89,47 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
+       "display": "Comment",
+       "fieldType": "text",
+       "editable": true
      }
    ],
    "commonLabels": [
      "cromwell-workflow-id",
-     "workflow-name"
+     "workflow-name",
+     "flag"
```
samanehsan: Should this include `label` and `comment` as well?

servers/cromwell/README.md
```diff
@@ -38,15 +38,41 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
+      "field": "labels.comment",
+      "display": "Comment",
+      "fieldType": "text",
+      "editable": true
     }
   ],
   "commonLabels": [
     "cromwell-workflow-id",
-    "workflow-name"
+    "workflow-name",
```
samanehsan: Same as below, I think we would include `label` and `comment` here so that these fields appear in the query builder

ui/src/environments/config-loader.service.ts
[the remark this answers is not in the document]
bfcrampton: Ah okay. I think its just because of the complexity of the promises in this case which is required due to the weird interface on `XMLHttpRequest` (namely setting the `onload` and `onerror` callbacks manually inside of a promise which would be more naturally handled by a `Promise`.

After looking into it more, I think you want to use the [Angular Http Module](https://medium.com/google-developer-experts/angular-2-introduction-to-new-http-module-1278499db2a0), like we do in the `job-manager.service` and `capabilities.service`. I think it would look something like:
```
constructor(private http: Http) {}

getEnvironmentConfig(): Promise<any> {
  return this.http.get(`../../../assets/environments/environment.json`,
    {},
    new RequestOptions({headers: new Headers({'Content-Type': 'application/json'})}))
    .toPromise()
    .then(response => response.json())
    .catch((e) => this.handleError(e));
}
```

The http client also does the parsing into an object for you so you won't even have to parse raw text.

Let me know if something like that works.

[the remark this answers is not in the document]
rexwangcc: This approach looks elegant to me, just tried but ran into 2 issues:
1. According to https://angular.io/api/http/Http, the `Http` has been deprecated in Angular. 
2. The deprecation of the `Http` is OK since we can either use the new `HttpClient` to do the something or keep using it like we did in the other services, the tricky problem here is, since the `ConfigLoaderService` has to be instantiated in the main.ts:

If we want to use `Http` since there are no modules stand with main.ts, instead of simply passing `Http` or `HttpModule` in module.ts, we have to manually resolve and create the provider for the Http instance that will be passed into the constructor .e.g `new ConfigLoaderService(injector.get(Http));`, to achieve this, we need something like:
```
const injector = ReflectiveInjector.resolveAndCreate([HTTP_PROVIDERS]);
const configLoader = new ConfigLoaderService(injector.get(Http));
```
However, since `HTTP_PROVIDERS` is also deprecated and *removed* (https://github.com/angular/angular/pull/10864), I have to something like:
```
const HTTP_PROVIDERS = [
  {provide: Http, useFactory:
      (xhrBackend: XHRBackend, requestOptions: RequestOptions): Http =>
        new Http(xhrBackend, requestOptions),
    deps: [XHRBackend, RequestOptions]},
  BrowserXhr,
  {provide: RequestOptions, useClass: BaseRequestOptions},
  {provide: ResponseOptions, useClass: BaseResponseOptions},
  XHRBackend,
  {provide: XSRFStrategy, useFactory: () => new CookieXSRFStrategy()},
];
const injector = ReflectiveInjector.resolveAndCreate(HTTP_PROVIDERS);
const configLoader = new ConfigLoaderService(injector.get(Http));
```
which worked but really clunky.

Another way could be using the new [`HttpClient`](https://angular.io/guide/http), but using it will bring code style inconsistency between the ConfigloaderService and other services(which are still using Http), I just created a ticket regarding this in the board. Given the effort required fo refactoring this, would you mind I treat this as a follow-up PR/work? Or, did i miss anything here? @bfcrampton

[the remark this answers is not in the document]
bfcrampton: Ah okay I didn't think about all the dependency injection you'd need since it's in `main.ts`. Definitely fine leaving this as follow up work, and I agree we should move to `HttpClient` over the deprecated `Http` everywhere. 

Does that clunky version work fine and simplify the logic here? I think I'd prefer that if it's easy just because it is more consistent with the rest of the app. But if that also takes work and we're going to migrate over to `HttpClient` later I'm fine with using `XMLHttpRequest` directly.

rexwangcc:
@bfcrampton Thanks for taking another look on this! Now as I revisit it, the clucky version doesn't work without copying over some of the source code of Angular (e.g. https://angular.io/api/http/HttpModule) to the Job Manager, which is I'm a little bit hesitant to do. I'm going to merge this PR today.

bfcrampton:
Sounds good, merge as is 👍. 

Would you mind filing tickets for switching off of `XMLHttpRequest` here and off of the deprecated `Http` in our other services?

rexwangcc:
Thanks for all the comments on this PR and the closed PR during the implementation process. Going to merge in this PR, tickets no.346 and no.347 which cover follow up questions for this PR has been filed in the backlog.

PR #302, opened 2018-06-25
Added link to status counts on dashboard page.

zacharyhsia:
Please help improve the PR, thanks!

### DIFF
--- a/ui/src/app/dashboard/dashboard.component.spec.ts
+++ b/ui/src/app/dashboard/dashboard.component.spec.ts
@@ -8,7 +8,7 @@ import {RouterTestingModule} from '@angular/router/testing';
 
 import {JobManagerService} from '../core/job-manager.service';
 import {SharedModule} from '../shared/shared.module';
-import {Router} from "@angular/router";
+import {ActivatedRoute, Router} from "@angular/router";
 import 'rxjs/add/observable/of';
 import {RouteReuse} from '../route-reuse.service';
 import {DashboardResolver} from "./dashboard.resolver.service";
@@ -31,6 +31,7 @@ describe('DashboardComponent', () => {
         DashboardComponent,
         TotalSummaryComponent,
         GroupedSummaryComponent,
+        TestJobListComponent,
       ],
       imports: [
         CommonModule,
@@ -39,8 +40,9 @@ describe('DashboardComponent', () => {
         MatTableModule,
         RouterTestingModule.withRoutes([
           {path: 'dashboard', component: DashboardComponent, resolve: {aggregations: DashboardResolver}},
+          {path: 'jobs', component: TestJobListComponent}
         ]),
-        SharedModule
+        SharedModule,
       ],
       providers: [
         {provide: JobManagerService, useValue: fakeJobService},
@@ -67,9 +69,23 @@ describe('DashboardComponent', () => {
     expect(testComponent).toBeTruthy();
   }));
 
+  it('should navigate to job-list page (fake) when status counts are clicked', fakeAsync(() => {
+    let countAnchor = fixture.debugElement.query(By.css("jm-dashboard a")).nativeElement;
+    countAnchor.click();
+    fixture.detectChanges();
+    tick();
+    expect(fixture.debugElement.query(By.css("div")).nativeElement.textContent).toEqual("fake job-list page");
+  }));
+
   @Component({
     selector: 'jm-test-app',
     template: '<router-outlet></router-outlet>'
   })
   class AppComponent {}
+
+  @Component({
+    selector: 'jm-test-job-list-component',
+    template: '<div>fake job-list page</div>'
+  })
+  class TestJobListComponent {}
 });
--- a/ui/src/app/dashboard/dashboard.component.spec.ts
+++ b/ui/src/app/dashboard/dashboard.component.spec.ts
@@ -5,22 +5,23 @@ import {Component} from '@angular/core';
 import {MatCardModule, MatTableModule} from '@angular/material';
 import {MatDividerModule} from '@angular/material/divider';
 import {RouterTestingModule} from '@angular/router/testing';
+import {ActivatedRoute, Router} from "@angular/router";
+import 'rxjs/add/observable/of';
 
 import {JobManagerService} from '../core/job-manager.service';
 import {SharedModule} from '../shared/shared.module';
-import {ActivatedRoute, Router} from "@angular/router";
-import 'rxjs/add/observable/of';
 import {RouteReuse} from '../route-reuse.service';
 import {DashboardResolver} from "./dashboard.resolver.service";
 import {DashboardComponent} from "./dashboard.component";
 import {TotalSummaryComponent} from "./total-summary/total-summary.component";
 import {GroupedSummaryComponent} from "./grouped-summary/grouped-summary.component";
-import {FakeAggregationService} from "../testing/fake-aggregation.service";
+import {FakeAggregationService, CARD_NUM, TEST_AGGREGATION_RESPONSE} from "../testing/fake-aggregation.service";
 
 describe('DashboardComponent', () => {
   let testComponent: DashboardComponent;
   let fixture: ComponentFixture<AppComponent>;
   let fakeJobService: FakeAggregationService;
+  let testJobListComponent: TestJobListComponent;
 
   beforeEach(async(() => {
     fakeJobService = new FakeAggregationService();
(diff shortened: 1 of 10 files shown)


zacharyhsia:
Wanted to add tests that are more than creating sub-components but had not cue about how to test those **links**.

zacharyhsia:
I would like to improve my skills of writing technical documentations. Could you please send me some sample docs to learn? Thanks!

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +69,23 @@ describe('DashboardComponent', () => {
     expect(testComponent).toBeTruthy();
   }));
 
+  it('should navigate to job-list page (fake) when status counts are clicked', fakeAsync(() => {
+    let countAnchor = fixture.debugElement.query(By.css("jm-dashboard a")).nativeElement;
+    countAnchor.click();
+    fixture.detectChanges();
+    tick();
+    expect(fixture.debugElement.query(By.css("div")).nativeElement.textContent).toEqual("fake job-list page");
```
calbach: Could you make some assertions about the route, i.e. parsing the query parameter? Possibly by injecting the route into the `TestJobListComponent`, then getting a reference there? Alternatively, you cou[ld try to inject the ActiveRoute directly into the test case.

If you do this, you probably also want to select a more specific link on the page to click, e.g. in the "total summaries", find the link that says "failed", then click it.

If possible, another test case that clicks one of the aggregation cards would also be nice here.

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
+
+  it('should navigate to job-list page (fake) when status counts are clicked', fakeAsync(() => {
+    let countAnchor = fixture.debugElement.query(By.css("jm-dashboard a")).nativeElement;
+    countAnchor.click();
+    fixture.detectChanges();
+    tick();
+    expect(fixture.debugElement.query(By.css("div")).nativeElement.textContent).toEqual("fake job-list page");
+  }));
+
+  it('should has status as url param when totalSummaryComponent links are clicked', fakeAsync(() => {
```
calbach: grammar nit: "has" -> "have"

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
+    let totalSummaryAnchor = fixture.debugElement.query(By.css("jm-total-summary .count a")).nativeElement;
+    let status = fixture.debugElement.query(By.css("jm-total-summary .status")).nativeElement.textContent.trim();
+
```
calbach: I'd assert a specific value for `status` here

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
+    totalSummaryAnchor.click();
+    fixture.detectChanges();
+    tick();
+    testJobListComponent = fixture.debugElement.query(By.css("jm-test-job-list-component")).componentInstance;
+    expect(testJobListComponent.route.snapshot.queryParams['q'].indexOf(status)).toBeGreaterThan(-1);
```
calbach: You can do `expect(...).toContain(status);`, but also see next comment

calbach: I think it would probably be better to actually unpack `q` with the parameter utility here, then test against a specific value in that request (can also check project there)

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
+  it('should navigate to job-list page (fake) when status counts are clicked', fakeAsync(() => {
+    let countAnchor = fixture.debugElement.query(By.css("jm-dashboard a")).nativeElement;
+    countAnchor.click();
+    fixture.detectChanges();
+    tick();
+    expect(fixture.debugElement.query(By.css("div")).nativeElement.textContent).toEqual("fake job-list page");
+  }));
+
+  it('should has status as url param when totalSummaryComponent links are clicked', fakeAsync(() => {
+    let totalSummaryAnchor = fixture.debugElement.query(By.css("jm-total-summary .count a")).nativeElement;
```
calbach: optional: Alias debugElement to make this test a bit less verbose: `const de = fixture.debugElement;`

calbach: Use `const` for all final variables.

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -5,22 +5,23 @@ import {Component} from '@angular/core';
...
 import {TotalSummaryComponent} from "./total-summary/total-summary.component";
 import {GroupedSummaryComponent} from "./grouped-summary/grouped-summary.component";
-import {FakeAggregationService} from "../testing/fake-aggregation.service";
+import {FakeAggregationService, CARD_NUM, TEST_AGGREGATION_RESPONSE} from "../testing/fake-aggregation.service";
 
 describe('DashboardComponent', () => {
   let testComponent: DashboardComponent;
   let fixture: ComponentFixture<AppComponent>;
   let fakeJobService: FakeAggregationService;
+  let testJobListComponent: TestJobListComponent;
```
calbach: I don't think the variable makes sense here, since it is only populated within specific test cases. I would just remove this line and then put a `const` or `let` where you're defining it

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
   })
   class AppComponent {}
+
+  @Component({
+    selector: 'jm-test-job-list-component',
+    template: '<div>fake job-list page</div>'
+  })
+  class TestJobListComponent {
+    public route: ActivatedRoute;
+      constructor(activatedRoute : ActivatedRoute) {
```
calbach: In Typescript, there's syntactic sugar for this pattern, so you shouldn't need to make these Java-style constructors in general. The following achieves the same result as what you have currently:

```
class TestJobListComponent {
  constructor(public activatedRoute: ActivatedRoute) {}
}
```

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -67,9 +70,55 @@ describe('DashboardComponent', () => {
...
+    testJobListComponent = fixture.debugElement.query(By.css("jm-test-job-list-component")).componentInstance;
+    expect(testJobListComponent.route.snapshot.queryParams['q'].indexOf(status)).toBeGreaterThan(-1);
+  }));
+
+  it("should has status and label as url params when groupedSummaryComponent links are clicked", fakeAsync(() => {
+    let groupedSummaryAnchor = fixture.debugElement.query(By.css("jm-grouped-summary tr td.count a")).nativeElement;
+    groupedSummaryAnchor.click();
+    tick();
+    testJobListComponent = fixture.debugElement.query(By.css("jm-test-job-list-component")).componentInstance;
+    let q = testJobListComponent.route.snapshot.queryParams['q'];
```
calbach: Same as above, would parse the search request out instead, and check a specific field.

ui/src/app/testing/fake-aggregation.service.ts
```diff
@@ -89,9 +91,13 @@ const aggregationResponse: AggregationResponse =
    }
   };
 
+// This is the number of mat-card that would display on dashboard page,
+// remember to change the number if you change the aggregationResponse above
+export const CARD_NUM = 4;
```
calbach: I think it would probably be slightly clearer to get rid of this property and instead do this from the test to avoid the risk of these going out of sync:

`TEST_AGGREGATION_RESPONSE.aggregations.length`

Also nitpick, but w.r.t. separation of concerns this service stub shouldn't know about "cards" since it's a presentation concern. The dashboard component renders specific card components in the UI, but the AggregationService should only know about the details of the API. This service could potentially be used elsewhere in the app in the future where there's no card UX, or for example in the future we might only display the first N cards.

ui/src/app/dashboard/dashboard.component.spec.ts
zacharyhsia: Adapted, thanks!

ui/src/app/dashboard/total-summary/total-summary.component.ts
```diff
@@ -9,5 +12,14 @@ import {StatusCounts} from "../../shared/model/StatusCounts";
 export class TotalSummaryComponent {
   @Input() summary: StatusCounts;
 
-  constructor() {}
+  constructor(private readonly activatedRoute: ActivatedRoute) {}
+
+  getUrlParams(statusCount: StatusCount) {
+    let map = new Map<string, string[]>();
+    map.set('projectId', [this.activatedRoute.snapshot.queryParams['projectId']]);
```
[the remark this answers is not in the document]
zacharyhsia: I made it an array to match the specified type of the map (`Map<string, string[]>`), which is required for the `encodeURLSearchParamsFromMap()` method. Is there a way to avoid doing this?

ui/src/app/testing/fake-aggregation.service.ts
```diff
@@ -1,13 +1,14 @@
 import {AggregationResponse, JobStatus, TimeFrame} from "../shared/model/models";
 
-const aggregationResponse: AggregationResponse =
+export const TEST_AGGREGATION_RESPONSE: AggregationResponse =
```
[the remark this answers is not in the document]
zacharyhsia: Adapted, thanks for the suggestion!

ui/src/app/testing/fake-aggregation.service.ts
zacharyhsia: Adapted!

ui/src/app/dashboard/total-summary/total-summary.component.ts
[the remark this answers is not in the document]
bfcrampton: Ah no you're correct, sorry!

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -5,22 +5,23 @@ import {Component} from '@angular/core';
 import {MatCardModule, MatTableModule} from '@angular/material';
 import {MatDividerModule} from '@angular/material/divider';
 import {RouterTestingModule} from '@angular/router/testing';
+import {ActivatedRoute, Router} from "@angular/router";
+import 'rxjs/add/observable/of';
```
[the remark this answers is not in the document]
bfcrampton: Yup just meant the `rxjs/add/observable/of` part, thanks

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -172,45 +172,45 @@ describe('DashboardComponent', () => {
...
     const queryJobRequest = URLSearchParamsUtils.unpackURLSearchParams(testJobListComponent.activatedRoute.snapshot.queryParams['q']);
     expect(queryJobRequest.statuses).toContain(status);
     expect(queryJobRequest.extensions['projectId']).toEqual(TEST_PROJECT);
   }));
 
-  it("should have status and label as url params when groupedSummaryComponent links are clicked", fakeAsync(() => {
-    const groupedSummaryAnchor = de.query(By.css("jm-grouped-summary tr td.count a")).nativeElement;
+  it('should have status and label as url params when groupedSummaryComponent links are clicked', fakeAsync(() => {
+    const groupedSummaryAnchor = de.query(By.css('jm-grouped-summary tr td.count a')).nativeElement;
     const status: JobStatus= JobStatus.Succeeded;
```
calbach: Same issue with `=` here

PR #304, opened 2018-06-29
Added header on the dashboard page.

zacharyhsia:
I agree with Bryan that there is no explicit reason we use a different URL encoding scheme on dashboard page and I changed it.

Also add **header component** on the dashboard page.

### DIFF
--- a/ui/src/app/dashboard/dashboard.component.spec.ts
+++ b/ui/src/app/dashboard/dashboard.component.spec.ts
@@ -18,6 +18,9 @@ import {FakeAggregationService} from "../testing/fake-aggregation.service";
 import {URLSearchParamsUtils} from "../shared/utils/url-search-params.utils";
 import {JobStatus} from "../shared/model/JobStatus";
 import {AggregationResponse} from "../shared/model/AggregationResponse";
+import {CapabilitiesResponse} from "../shared/model/CapabilitiesResponse";
+import {CapabilitiesService} from "../core/capabilities.service";
+import {FakeCapabilitiesService} from "../testing/fake-capabilities.service";
 
 const TEST_AGGREGATION_RESPONSE: AggregationResponse =
   {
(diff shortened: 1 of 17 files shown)


zacharyhsia:
Hi Bryan, ready to be reviewed! Thanks!

bfcrampton:
Taking a look now!

ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -155,17 +167,17 @@ describe('DashboardComponent', () => {
...
     });
 
     tick();
     fixture.detectChanges();
     tick();
-
   }));
 
   it('should create dashboard', fakeAsync(() => {
+    // debugger;
```
bfcrampton: rm

ui/src/app/dashboard/grouped-summary/grouped-summary.component.ts
```diff
@@ -37,7 +37,10 @@ export class GroupedSummaryComponent {
 
   getCommonUrlParamsMap(entry: AggregationEntry) {
     let map = new Map<string, string[]>();
-    map.set('projectId', [this.activatedRoute.snapshot.queryParams['projectId']]);
+
+    const projectId = URLSearchParamsUtils.unpackURLSearchParams(this.activatedRoute.snapshot.queryParams['q']).extensions['projectId'];
```
bfcrampton: Can you wrap this into two lines its pretty long

ui/src/app/projects/projects.component.ts
```diff
@@ -95,21 +95,20 @@ export class ProjectsComponent implements OnInit {
   }
 
   navigateJobs() {
+    const extras = {
```
bfcrampton: nit: name it something like projectQueryParams

zacharyhsia: Thanks! This value is of `[NavigationExtra](https://angular.io/api/router/NavigationExtras)` class defined by Angular. Only the `queryParams` is about query. Plus, it is originally called `extra`, which might make more sense here.

ui/src/app/dashboard/grouped-summary/grouped-summary.component.ts
zacharyhsia: Fixed!

ui/src/app/dashboard/dashboard.component.spec.ts
zacharyhsia: :)

PR #305, opened 2018-07-02
Make Gunicorn parameters configurable

rexwangcc:
This is an experiment PR, trying to tweak the Gunicorn parameters for docker-compose based on http://docs.gunicorn.org/en/stable/settings.html and http://docs.gunicorn.org/en/stable/design.html 

It's not necessary to merge in this PR if this doesn't make sense to others.


common-compose.yml
```diff
@@ -37,6 +37,7 @@ services:
       - CROMWELL_CREDENTIALS=/app/jobs/config.json
       - CAPABILITIES_CONFIG=/app/jobs/capabilities_config.json
       - CROMWELL_URL=${CROMWELL_URL}
+      - GUNICORN_CMD_ARGS=${GUNICORN_CMD_ARGS:-"--workers=5"}
```
samanehsan: How are the gunicorn workers currently configured for the shim layer for cromwell and dsub?

rexwangcc: We don't have any special settings for the gunicorn, which means currently it is using the default value: 1 synchronous worker. This might not be sufficient in some cases. Besides, parameterizing them here can improve the deployment, since different users may have different requirements.

servers/cromwell/README.md
```diff
@@ -15,15 +15,22 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     export CROMWELL_URL=http://192.168.0.106:8000/api/workflows/v1
     ```
 
-2. Add a file named `config.json` to `job-manager/servers/cromwell/jobs` that contains the username and password for the specified cromwell instance (You may want to skip this step if the Cromwell instance does not have HTTP Basic Authentication):
+2. (Optional) By default, the shim layer uses **5** `sync` [Gunicorn workers](http://docs.gunicorn.org/en/stable/settings.html#worker-class), the number of workers is **5**. 
+    Job Manager also comes with `gevent` workers, you can override the default worker type and use dynamic number of workers(based on the number of CPU cores) by:
+    
+    ```
+    export GUNICORN_CMD_ARGS="--workers=$((2 * $(getconf _NPROCESSORS_ONLN 2>/dev/null || getconf NPROCESSORS_ONLN 2>/dev/null || echo 2) + 1)) --worker-class gevent"
```
bfcrampton: Is `_NPROCESSORS_ONLN` available on all platforms? Mac OS X, Linux, etc.?

servers/cromwell/README.md
```diff
@@ -15,15 +15,22 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     export CROMWELL_URL=http://192.168.0.106:8000/api/workflows/v1
     ```
 
-2. Add a file named `config.json` to `job-manager/servers/cromwell/jobs` that contains the username and password for the specified cromwell instance (You may want to skip this step if the Cromwell instance does not have HTTP Basic Authentication):
+2. (Optional) By default, the shim layer uses **5** `sync` [Gunicorn workers](http://docs.gunicorn.org/en/stable/settings.html#worker-class), the number of workers is **5**.
```
bfcrampton: Why 5 workers, is that just arbitrary?

servers/cromwell/README.md
bfcrampton: Is our flask code compatible with gevent workers? Could you add some informatino explaining the functional difference between `sync` and `gevent`?

Also, I may separate these two concepts out. You can separately adjust the amount of workers and the worker-class so I might have these as two separate examples for clarity.

servers/cromwell/README.md
```diff
@@ -15,15 +15,22 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
     ```
     {
       "cromwell_user" : "USERNAME",
       "cromwell_password" : "PASSWORD"
     }
     ```
 
-3. (Optional) If you want to change either the predefined view or the default behavior of UI to some extent, e.g. display more columns in the job list view such as labels of the jobs, or make more columns to be popped up by the query builder, you can add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations. The `capabilities_config.json` should **strictly** follow the following structure:
+4. (Optional) If you want to change either the predefined view or the default behavior of UI to some extent, e.g. display more columns in the job list view such as labels of the jobs, or make more columns to be popped up by the query builder, you can add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations. The `capabilities_config.json` should **strictly** follow the following structure:
```
bfcrampton: minor nit: Does this not display correctly if we use `1.` for all of these? (I know sometimes it doesn't work but forget the circumstances)

common-compose.yml
samanehsan: I like the idea of making this a configurable option! How does this parameter get passed to gunicorn?

rexwangcc: Thanks! According to the documentation(http://docs.gunicorn.org/en/stable/settings.html#settings), gunicorn can read the parameters from the environment variable called `GUNICORN_CMD_ARGS `

servers/cromwell/README.md
rexwangcc: Oh, so according to the documentation(http://docs.gunicorn.org/en/stable/settings.html#workers), Gunicorn suggests a range of `2-4 x $(NUM_CORES)` for the number of workers to get the best load. People always use `2N+1` as a practice. The default number 5 here assumes you are using at least 2-core CPU (since even 4-core so so common nowadays), so `2 x 2 + 1 = 5`

servers/cromwell/README.md
rexwangcc: @bfcrampton 

1. `_NPROCESSORS_ONLN` is available for most of the Linux systems (including MAC OS), `NPROCESSORS_ONLN` is the equivalence in BSD systems. If neither of them are found, the default value `2 x 2 + 1 = 5` will be used here. This command in the readme should be portable enough for users (who deploy JMUI) to use. What do you think?

2. According to the documentation (http://docs.gunicorn.org/en/stable/design.html#choosing-a-worker-type) and my understanding, I think our flask code is compatible with `gevent`(or `Eventlet`) workers. In fact, I think we should even avoid using `sync` workers because `sync` workers assume that:
    >  your application shouldn’t do anything that takes an undefined amount of time

    Among all of the async workers Gunicorn provides(`gevent/Eventlet`, `tornado`, `asyncIO`, `gthread`), I believe `gevent` should be the most compatible worker type for us to use.

3. Separating the 2 concepts out is a good idea, I will change it! Thanks!

servers/cromwell/README.md
rexwangcc: I think all `1.`s will definitely work, as long as we have proper indentations. I changed here before due to my OCD on raw-text :( But now since we have more numbered items, I think it is a good idea to change back to use all `1.` here. Will do it.

servers/cromwell/README.md
rexwangcc: Interestingly, not sure if it's an illusion, I noticed a faster loading-speed when using `gevent` workers for JMUI. We should perform a benchmarking for this as well .

common-compose.yml
samanehsan: Oh, okay! Thanks for the explanation here

rexwangcc:
Thanks for all the comments on this PR! It seems there are no objections for this PR, so I'm going to merge this in.

servers/cromwell/README.md
bfcrampton: Sorry I never responded to this, all sounds good. As a side note: if `gevent` seems faster maybe we should see if that can be made the default.

PR #307, opened 2018-07-03
A few minor changes to the Cromwell getting started process docs

cjllanwarne:
Changes in-place: https://github.com/DataBiosphere/job-manager/blob/cjl_getting_started/README.md

### DIFF
--- a/servers/cromwell/README.md
+++ b/servers/cromwell/README.md
@@ -56,6 +56,8 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     - Make more columns to be popped up by the query builder
   - Add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations.
   - The `capabilities_config.json` should **strictly** follow the following structure:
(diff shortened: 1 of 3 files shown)


servers/cromwell/README.md
```diff
@@ -4,168 +4,181 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
     {
       "cromwell_user" : "USERNAME",
       "cromwell_password" : "PASSWORD"
     }
     ```
 
-1. (Optional) If you want to change either the predefined view or the default behavior of UI to some extent, e.g. display more columns in the job list view such as labels of the jobs, or make more columns to be popped up by the query builder, you can add a `capabilities_config.json` file to `job-manager/servers/cromwell/jobs` to override the pre-defined configurations. The `capabilities_config.json` should **strictly** follow the following structure:
-    ```
+- (Optional for local Cromwell) Configure fields to display
+  - **Note**: CaaS-specific details follow in the next section.
```
rsasch: I found this confusing; why refer to one section within a different section, when that first section is clearly labeled as "Required for CaaS"?

PR #312, opened 2018-07-05
Fix pagination for main job list for Cromwell

rsasch:
Now with Cromwell version 33, we can query for only parent workflows and do pagination properly.
- added `includeSubworkflows` set to `false` in query to Cromwell backend
- re-worked the way the shim layer for Cromwell uses page tokens
- added additional UI tests for pagination

**NB**: this depends on Cromwell version 34 or later

See https://elastc.com/c/C46fjeNX/320-bug-when-there-s-only-one-page-it-still-says-of-many


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -171,21 +171,18 @@ def query_jobs(body, **kwargs):
...
     total_results = get_total_results(query, auth, headers)
-
     results = []
     offset = page_tokens.decode_offset(query.page_token) or 0
-    page = page_from_offset(offset, page_size)
-    last_page = get_last_page(total_results, page_size)
+    page = page_from_offset(offset, query_page_size)
+    last_page = get_last_page(total_results, query_page_size)
 
     while len(results) < query_page_size and page <= last_page:
```
samanehsan: We can get rid of this while loop now that we no longer have to filter out subworkflows and continually make additional requests to get a full page of results. Instead, we can make a request to cromwell for a specific page.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -259,6 +256,7 @@ def cromwell_query_params(query, page, page_size):
     query_params.append({'page': str(page)})
     query_params.append({'additionalQueryResultFields': 'parentWorkflowId'})
     query_params.append({'additionalQueryResultFields': 'labels'})
+    query_params.append({'includeSubworkflows': 'true'})
```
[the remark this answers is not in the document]
rsasch: good catch

ui/src/app/shared/header/header.component.ts
```diff
@@ -246,39 +233,3 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
   }
 }
```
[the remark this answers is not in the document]
rsasch: I could be missing something, but it looked like there was customized pagination functionality to handle the fact that pagination wasn't working correctly on the backend for Cromwell. Since I'm hoping to fix this with this PR, I removed it.

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: thanks – will do

ui/src/app/shared/header/header.component.ts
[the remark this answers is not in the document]
calbach: That's not the purpose of this code. I will try to explain:
- QueryJobs is a paginated API.
- It is very common (at least for dsub), that when you come to the list page, your query will match way more results than you'd like to load (thousands, 10s of thousands), so we only load an initial hundred or so.
- (once you paginate, we will lazily load more pages as you near the end of what you have on the client)
- So a common state is that I've loaded ~100 / ~1k jobs that match my filter onto the client (in the client code, this is the "exhaustive" property).
- While we're in this state, we don't want to show "showing page 1/2", as this implies that your filter matches 100 jobs only.
- Instead, we have this custom code to override the default paginator text, and instead show "of many", since we don't have an accurate count of the total number of jobs.

Note: this behavior could actually be fixed in Cromwell's case by adding a "totalResults" field on the QueryJobsResponse. We don't know the answer to that on the dsub side though, so that would need to be an optional part of the API and we'd still need to maintain the existing behavior.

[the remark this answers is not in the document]
rsasch: ok, thanks for the info!

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -171,41 +171,30 @@ def query_jobs(body, **kwargs):
...
+        raise InternalServerError(response.json().get('message'))
+    response.raise_for_status()
+
+    now = datetime.utcnow()
+    jobs_list = [format_job(job, now) for job in response.json()['results']]
+    jobs_list.reverse()
 
+    if page == last_page:
+        return QueryJobsResponse(results=jobs_list)
     next_page_token = page_tokens.encode_offset(offset)
```
calbach: Are there tests for this? I can tell by eye this won't work since offset is never incremented.

ui/src/app/shared/header/header.component.spec.ts
```diff
@@ -166,7 +166,7 @@ describe('HeaderComponent', () => {
     });
   }));
 
-  it('should only show length for exhaustive job streams', async(() => {
+  it('should always show length of total job list', async(() => {
```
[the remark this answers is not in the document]
rsasch: can you explain?

ui/src/app/shared/header/header.component.ts
```diff
@@ -279,6 +279,6 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
     const endIndex = startIndex < length ?
         Math.min(startIndex + pageSize, length) :
         startIndex + pageSize;
-    return `${startIndex + 1} - ${endIndex} of many`;
+    return `${startIndex + 1} - ${endIndex} of ${length}`;
```
samanehsan: Can dsub get the total number of results here @calbach?

ui/src/app/shared/header/header.component.spec.ts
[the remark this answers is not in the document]
calbach: I didn't notice the header.component.ts change. That should also be reverted. This should be explained by my long comment, but let me know if something is still unclear about how this works.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -235,7 +225,10 @@ def get_last_page(total_results, page_size):
 
 
 def page_from_offset(offset, page_size):
-    return 1 + offset / page_size
+    if offset == 0:
+        return 1
+    else:
```
calbach: nit: can drop the "else" here to reduce indentation

ui/src/app/shared/job-stream.ts
```diff
@@ -36,7 +36,7 @@ export class JobStream extends BehaviorSubject<JobListView> {
         return prevResp;
       }
       let pageSize = Math.max(
-        JobStream.minBackendPageSize, this.value.results.length - atLeast);
+        JobStream.minBackendPageSize, Math.abs(this.value.results.length - atLeast));
```
calbach: Did you mean to leave this in? I can't rationalize why you would need this change.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -171,43 +171,33 @@ def query_jobs(body, **kwargs):
...
-            format_job(job, now) for job in response.json()['results']
-            if _is_parent_workflow(job)
-        ]
-        jobs_list.reverse()
-        results.extend(jobs_list)
-        offset = offset + page_size
-        page = page_from_offset(offset, page_size)
+    response = requests.post(
+        _get_base_url() + '/query',
+        json=cromwell_query_params(query, page, query_page_size),
```
samanehsan: This will return the results from oldest to newest (consistent with Cromwell's behavior), but that's not ideal for our use case here so the workaround for that was to fetch the pages in reverse order. I'm not sure how soon the Cromwell team is planning on implementing a way to sort results by newest first, but I don't think we should change the ordering in job manager.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -259,6 +252,7 @@ def cromwell_query_params(query, page, page_size):
     query_params.append({'page': str(page)})
     query_params.append({'additionalQueryResultFields': 'parentWorkflowId'})
     query_params.append({'additionalQueryResultFields': 'labels'})
+    query_params.append({'includeSubworkflows': 'false'})
```
rexwangcc: It is very likely we won't hit the issue, since Job Manager is using pagination to query Cromwell./ Just want to comment that there is an issue with the current version of Cromwell if you use `includeSubworkflows` to query for more than ~1000 workflows: https://github.com/broadinstitute/cromwell/issues/3873

ui/src/app/shared/job-stream.ts
rsasch: This is needed for the case where there needs to be a request for more jobs and `JobStream.minBackendPageSize` is smaller than `pageSize` (defined in `JobListComponent`).

calbach: OK, there is a bug here which is that the order was incorrect, should be `atLeast - ...results.length`

kbergin:
Merging this ticket with one I made

gemmalam:
@kbergin what is the ticket you made? any issue link in zenhub?

samanehsan:
Is there an update on if/when the Cromwell team can add in support for reversing the ordering of results?

gemmalam:
@samanehsan not yet, but @cjllanwarne will be talking with @geoffjentry to figure out a plan

cjllanwarne:
@samanehsan https://github.com/broadinstitute/cromwell/pull/3903

rsasch:
@samanehsan With Cromwell version 34, the list of jobs/workflows is returned with the latest submitted first. Would you mind taking another pass through this PR?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -172,43 +172,33 @@ def query_jobs(body, **kwargs):
     headers = kwargs.get('auth_headers')
     query = QueryJobsRequest.from_dict(body)
     query_page_size = query.page_size or _DEFAULT_PAGE_SIZE
-    # Request more than query.page_size from cromwell since subworkflows will get filtered out
-    page_size = query_page_size * 2
     total_results = get_total_results(query, auth, headers)
+    if query.page_token is None:
```
calbach: If the pages are in the proper order now, then you shouldn't need to get `total_results` first anymore; this was only done previously in order to find the "last page", then iterate backwards.

samanehsan:
Is there a standard for which versions of Cromwell the Job Manager should support? Or is it always just the latest release? It may be confusing to see this reversed order if you're using an older version.

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: `total_results` is also being used to figure out `last_page`, and that is used to figure out if a `next_page_token` should be returned

rsasch:
@samanehsan If I remember correctly, when we released a version of Job Manager that required (or worked best with) a minimum version of Cromwell, that was included in the the release notes. @kbergin can you confirm?

servers/cromwell/jobs/controllers/jobs_controller.py
calbach: But you should be able to get this value at line 183 without making this extra XHR, right?

rsasch: oh, right – because Cromwell will actually be returning the correct value 🙂

rsasch: fixed

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -236,7 +211,9 @@ def get_last_page(total_results, page_size):
 
 
 def page_from_offset(offset, page_size):
-    return 1 + offset / page_size
+    if offset == 0:
+        return 1
```
calbach: Are the pages actually 1-indexed...?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -172,59 +172,34 @@ def query_jobs(body, **kwargs):
...
     if response.status_code == BadRequest.code:
         raise BadRequest(response.json().get('message'))
     elif response.status_code == InternalServerError.code:
         raise InternalServerError(response.json().get('message'))
     response.raise_for_status()
 
-    return response.json()['totalResultsCount']
+    now = datetime.utcnow()
+    jobs_list = [format_job(job, now) for job in response.json()['results']]
+    if page == last_page:
```
calbach: nit: I'd go with >= to be on the safe side

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -172,59 +172,34 @@ def query_jobs(body, **kwargs):
...
-        ]
-        jobs_list.reverse()
-        results.extend(jobs_list)
-        offset = offset + page_size
-        page = page_from_offset(offset, page_size)
-
-    next_page_token = page_tokens.encode_offset(offset)
-    return QueryJobsResponse(results=results, next_page_token=next_page_token)
+    if query.page_token is None:
+        offset = 0
```
calbach: opt: personal preference probably but I do think something like this reads slightly better:

```
offset = 0
if query.page_token is not None:
  offset = ...
```

ui/src/app/shared/job-stream.ts
calbach: Please use the above fix instead.

rsasch: Are there really no circumstances where `this.value.results.length` could be more than `atLeast `?

calbach: At this point in the code, no. See line 34.

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: yes

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: changed

servers/cromwell/jobs/controllers/jobs_controller.py

ui/src/app/shared/job-stream.ts
rsasch: changed

PR #320, opened 2018-07-09
Replace the stub data (summary count) with real count 

zacharyhsia:


servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+
+    provider = providers.get_provider(generator.provider_type(), projectId,
+                                      generator.auth_token())
+    job_generator = generator.generate_jobs(provider, query)
+
+    total_count = 0
+
+    summary_counts = {}
+
+    # print(type(job_statuses.ApiStatus))
```
calbach: rm

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+    # for status in job_statuses.ApiStatus:
+    #     print(status)
+
+    for job in job_generator:
+        total_count += 1
+        _summary_count(job, summary_counts)
+
+    print("total: ", total_count, ", details: ", summary_counts)
```
calbach: rm

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+
+    return _get_stub_data(_to_summary_counts(summary_counts))
+
+
+def _summary_count(job, summary_counts):
+    if job.status == job_statuses.ApiStatus.SUBMITTED:
```
calbach: Do you need all these ifs? What about:

```
if job.status not in summary_counts:
  summary_counts[job.status] = 0
summary_counts[job.status] += 1
```

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+    elif job.status == job_statuses.ApiStatus.FAILED:
+        summary_counts[
+            job_statuses.ApiStatus.FAILED] = summary_counts.setdefault(
+                job_statuses.ApiStatus.FAILED, 0) + 1
+
+
+def _to_summary_counts(summary_counts):
+    counts = []
+
+    for status, count in summary_counts.items():
```
calbach: The usual way to do this in Python is to use a list comprehension (simple enough that I would probably inline it)  `[x for x in y]` (I forget the exact iterator you'll need for a dict).

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        counts.append(StatusCount(status, count))
+
+    return StatusCounts(counts)
+
+
+def _get_stub_data(summaryCounts):
     # temporary fake data for testing front-end.
-    statusCount1 = StatusCount(status='Succeeded', count=10)
+    # statusCount1 = StatusCount(status='Succeeded', count=10)
```
calbach: Please remove code rather than commenting it out.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -115,9 +111,12 @@ def query_jobs(body):
     Returns:
         QueryJobsResponse: Response containing results from query
     """
+    print(body)
```
calbach: Please remove all debugging statements from this PR

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,116 @@
...
+        last_create_time = job.submission
+
+    # If we hit the end of the dstat job generator, ensure to yield the jobs
+    # stored in the buffer before returning
+    for j in sorted(job_buffer, key=lambda j: j.id):
+        yield j
+
+
+def auth_token():
+    """(description)
```
calbach: Thanks for adding docs, but please follow the format we've used for the rest of our method documentation if you're going to add them, and also don't leave in this boilerplate line "(description)"

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,12 +25,79 @@ def get_job_aggregations(timeFrame, projectId=None):
     Returns:
         AggregationResponse: Response contains aggregation of jobs
     """
+    query = QueryJobsRequest.from_dict({
+        'extensions': {
+            'project_id': projectId
+        }
+    })
+
+    query.start = time_convert.time_frame_to_start_time(timeFrame)
```
calbach: We want the summaries to include all jobs that are still running or ended within the requested time period. This is not the same as filtering on the query start time.

For example, if I start a job 2 days ago and it's still running, then use TIME_FRAME = 1D, I should see the job in the summary.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -80,8 +74,10 @@ def get_job(id):
     Returns:
         JobMetadataResponse: Response containing relevant metadata
     """
-    proj_id, job_id, task_id = job_ids.api_to_dsub(id, _provider_type())
-    provider = providers.get_provider(_provider_type(), proj_id, _auth_token())
+    proj_id, job_id, task_id = job_ids.api_to_dsub(id,
+                                                   generator.provider_type())
+    provider = providers.get_provider(generator.provider_type(), proj_id,
```
calbach: Is this due to changes in yapf? Or your IDE? If it's the latter, please turn off the autoformatting or else make it match our yapf formatter

servers/dsub/jobs/controllers/utils/time_convert.py
```diff
@@ -0,0 +1,27 @@
+import datetime
+from dateutil.tz import tzlocal
+TIME_FRAME_MAP = {
```
calbach: newline above

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Being silly. Fixed.

servers/dsub/jobs/controllers/jobs_controller.py
zacharyhsia: It is due to yapf.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Thanks!

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/utils/time_convert.py
zacharyhsia: Fixed.

zacharyhsia: Fixed.

zacharyhsia:
Please review the PR. Thanks!

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+from jobs.controllers.utils import extensions, failures, job_ids, job_statuses, labels, providers, query_parameters
+from jobs.models.query_jobs_result import QueryJobsResult
+
+_MAX_PENDING_TIME_DAYS = 7
+
+
+def generate_jobs(provider, query, create_time_max=None, offset_id=None):
+    """Get the generator of jobs according to the specified filters.
+
+        Args:
```
calbach: nit: unindent (see other method docs in jobs_controller)

Please fix throughout this file

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+            provider (string): type of provider
+            query (QueryJobsRequest): query body
+            create_time_max (datetime): the latest job create time
+            offset_id (): paginator offset id
```
calbach: str

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+from jobs.models.query_jobs_result import QueryJobsResult
+
+_MAX_PENDING_TIME_DAYS = 7
+
+
+def generate_jobs(provider, query, create_time_max=None, offset_id=None):
+    """Get the generator of jobs according to the specified filters.
+
+        Args:
+            provider (string): type of provider
```
calbach: string->str. The "type" in python is "str".

servers/dsub/jobs/controllers/utils/time_convert.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Args:
+        time_frame_to_start_time (time_frame): 'HOURS_1', 'HOURS_8', 'HOURS_24',
+        'DAYS_7', 'DAYS_30' or 'ALL_TIME'
+
+    Returns:
+        start_time: a datetime object that is the start time of query and None if it is 'ALL_TIME'.
+    """
+
+    if (time_frame == 'ALL_TIME'):
```
calbach: In Python generally you don't put parens with if statements

servers/dsub/jobs/controllers/utils/time_convert.py
```diff
@@ -0,0 +1,28 @@
...
+    'DAYS_30': datetime.timedelta(days=30)
+}
+
+
+def time_frame_to_start_time(time_frame):
+    """
+    Map a time frame to start time.
+
+    Args:
+        time_frame_to_start_time (time_frame): 'HOURS_1', 'HOURS_8', 'HOURS_24',
```
calbach: The arg is named time_frame. The type is str

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+        if job.end and (job.end < window_min or job.end > window_max):
+            continue
+
+        jobs_buffer.append(job)
+
+    for job in jobs_buffer:
+        yield job
+
+
+def auth_token():
```
calbach: I would move auth_token and provider_type back into the controller, they aren't really related to the generator.

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+            job_buffer = []
+        last_create_time = job.submission
+
+    # If we hit the end of the dstat job generator, ensure to yield the jobs
+    # stored in the buffer before returning
+    for j in sorted(job_buffer, key=lambda j: j.id):
+        yield j
+
+
+def generate_jobs_count(provider,
```
calbach: It's not really returning counts, so I would call this `generate_jobs_by_window()`

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+                        project_id,
+                        window_min,
+                        window_max=datetime.datetime.now(tz=tzlocal())):
```
calbach: I think it would be better to default this to `None` in general. Also complex default parameters like this can have surprising behaviors; I would need to experiment and reread the Python docs to check that this does the right thing.

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+        statuses=None,
+        user_ids=None,
+        job_ids=None,
+        task_ids=None,
+        create_time_min=create_time_min,
+        create_time_max=window_max,
+        job_names=None,
+        labels=None))
+
+    jobs_buffer = []
```
calbach: I don't see any reason to buffer jobs here, just yield them immediately.

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,164 @@
...
+
+from flask import current_app, request
+from dsub.commands import ddel, dstat
+from dateutil.tz import tzlocal
+
+from jobs.common import execute_redirect_stdout
+from jobs.controllers.utils import extensions, failures, job_ids, job_statuses, labels, providers, query_parameters
+from jobs.models.query_jobs_result import QueryJobsResult
+
+_MAX_PENDING_TIME_DAYS = 7
```
calbach: At this point, it would probably make sense to name this more generally. e.g. MAX_RUNTIME_DAYS.

servers/dsub/jobs/controllers/utils/time_convert.py
```diff
@@ -0,0 +1,28 @@
+import datetime
+from dateutil.tz import tzlocal
+
+TIME_FRAME_MAP = {
+    'HOURS_1': datetime.timedelta(minutes=60),
+    'HOURS_8': datetime.timedelta(minutes=60 * 8),
```
calbach: hours=8?

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -1,21 +1,13 @@
...
-from typing import List, Dict
-from six import iteritems
-from ..util import deserialize_date, deserialize_datetime
-from werkzeug.exceptions import NotImplemented
 
 
 def get_job_aggregations(timeFrame, projectId=None):
-    """
-    Query for aggregated jobs in the given time frame.
+    """Query for aggregated jobs in the given time frame.
```
calbach: Have you looked at the jobs_controller test? I would recommend adding tests in this PR (or else, in a follow-up PR, but before you complete the full API implementation).

zacharyhsia: OK, will do it in the next PR.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Accepted.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Changed to None.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Accepted.

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Got it

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: The `aggregation_controller` can reuse these two functions. Either we put these two functions to other `utils` file where makes better sense or move the function calls of these two functions within the generator. I will probably go with the later one.

zacharyhsia: Both `aggragation_controller` and `job_controller` use these two functions so the second approach mentioned above is not viable. Do you have any suggestion where we should put them?

servers/dsub/jobs/controllers/utils/time_convert.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/utils/time_convert.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/utils/time_convert.py
zacharyhsia: Sure thing!

servers/dsub/jobs/controllers/utils/time_convert.py
```diff
@@ -0,0 +1,28 @@
+import datetime
```
bfcrampton: nit: I think `time_frame` or `time_frames` would be a better name for this file.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        _count_user_summary(job, user_summary)
+        _count_job_name_summary(job, job_name_summary)
+
+    return _get_stub_data(
+        _to_summary_counts(total_summary),
+        _to_user_aggregation_entry(user_summary),
+        _to_job_name_aggregation_entry(job_name_summary))
+
+
+def _count_total_summary(job, total_summary):
```
bfcrampton: nit: Group all of the `_count_` methods together

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+            countsList.append(StatusCount(status=status, count=count))
+
+        entries.append(
+            AggregationEntry(
+                label=user_id, status_counts=StatusCounts(countsList)))
+
+    return Aggregation(name='Job Name', key='name', entries=entries)
+
 
+def _get_stub_data(summaryCounts, user_aggregation, job_name_aggregation):
```
bfcrampton: nit: this is no longer stub data, rename

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        user_count[status] = 0
+    user_count[status] += 1
+
+    job_name_summary[job_name] = user_count
+
+
+def _to_job_name_aggregation_entry(job_name_summary):
+    entries = []
+
+    for user_id, countsDict in job_name_summary.items():
```
bfcrampton: isn't this a `job_name`?

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        user_count[status] = 0
+    user_count[status] += 1
+
+    user_summary[user_id] = user_count
+
+
+def _to_user_aggregation_entry(user_summary):
+    entries = []
+
+    for user_id, countsDict in user_summary.items():
```
bfcrampton: nit: snake_case

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+    user_count = job_name_summary.get(job_name, {})
+
+    if status not in user_count:
+        user_count[status] = 0
+    user_count[status] += 1
+
+    job_name_summary[job_name] = user_count
+
+
+def _to_job_name_aggregation_entry(job_name_summary):
```
bfcrampton: This function and `_to_user_aggregation_entry` can be condensed into one. This is also creating an `Aggregation`, not an `AggregationEntry` so should be named as such. Something like:
```
def _to_aggregation_entry(summary, name, key):
    entries = []
    for key, counts_dict in summary.items():
        counts = [StatusCount(s, c) for s,c in counts_dict.items()]
        entries.append(
            AggregationEntry(
                label=key, status_counts=StatusCounts(counts)))

    return Aggregation(name=name, key=key, entries=entries)
```

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        _to_job_name_aggregation_entry(job_name_summary))
+
+
+def _count_total_summary(job, total_summary):
+    if job.status not in total_summary:
+        total_summary[job.status] = 0
+    total_summary[job.status] += 1
+
+
+def _to_summary_counts(summary_counts):
```
bfcrampton: This is a trivial function I would just inline it above and shorten the vars:
`[StatusCount(s, c) for s, c in total_summary.items()]`

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    auth_header = request.headers.get('Authentication')
+    if auth_header:
+        components = auth_header.split(' ')
+        if len(components) == 2 and components[0] == 'Bearer':
+            return components[1]
+    return None
+
+
+def provider_type():
```
bfcrampton: Same here.

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,156 @@
...
+        job = _query_jobs_result(j, project_id)
+        # Filter jobs that do no end within the time window
+        if job.end and (job.end < window_min
+                        or window_max and job.end > window_max):
+            continue
+
+        yield job
+
+
+def auth_token():
```
bfcrampton: This is not used in this file, it should be moved back to `jobs_controller` (and added to `aggregation_controller` if necessary, as we did for the `capabilities_controller`)

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -0,0 +1,156 @@
+import datetime
```
bfcrampton: nit: generator is too general, I would name this file jobs_generator or something similar

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -73,28 +75,39 @@ def _to_user_aggregation_entry(user_summary):
...
-        status_counts=StatusCounts(counts=[
-            StatusCount(status='Succeeded', count=4),
-            StatusCount(status='Failed', count=6)
-        ]))
+    return Aggregation(name='Job Name', key='name', entries=entries)
 
-    ownerAggregation = Aggregation(
-        name='AnotherLabel', key='job-id', entries=[entry1, entry2])
 
+def _get_stub_data(summaryCounts, user_aggregation, job_name_aggregation):
```
calbach: This is no longer stub data, right? Also I would inline this method, it's simple enough to not justify using a helper function.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +16,98 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+    total_summary[job.status] += 1
+
+
+def _to_summary_counts(summary_counts):
+    return StatusCounts([
+        StatusCount(status, count) for status, count in summary_counts.items()
+    ])
+
+
+def _count_user_summary(job, user_summary):
```
bfcrampton: optional: You may want to think about making this function more generic by taking in a lambda for extracting the summary 'key' (the user_id in this case). Something like:
```
def count_for_key(job, summary, key_lambda):
    key = key_lambda(job)
    count = summary.get(key, {})

    if job.status not in user_count:
        user_count[status] = 0
    user_count[status] += 1

    summary[key] = count
```

And you can use it above like:
```
count_for_key(job, user_summary, lambda j: j.extensions.user_id)
count_for_key(job, job_name_summary, lambda j: j.name)
```

servers/dsub/jobs/controllers/utils/generator.py
```diff
@@ -84,23 +83,19 @@ def generate_jobs(provider, query, create_time_max=None, offset_id=None):
         yield j
 
 
-def generate_jobs_count(provider,
-                        project_id,
-                        window_min,
-                        window_max=datetime.datetime.now(tz=tzlocal())):
+def generate_jobs_by_window(provider, project_id, window_min, window_max=None):
     """Get the generator of jobs for aggregation.
```
calbach: Please specify that this is returning jobs that ended within the given window. And that it includes still running jobs if no window_max is specified.

servers/dsub/jobs/controllers/utils/generator.py
calbach: How about `utils/providers.py`?

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Nice! Accepted!

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Removed

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Renamed to jobs_generator

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: Accepted

servers/dsub/jobs/controllers/utils/time_convert.py
zacharyhsia: Renamed

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Removed

servers/dsub/jobs/controllers/aggregation_controller.py
bfcrampton: Sorry I forgot to rename the function, should be `_to_aggregation(...)`

servers/dsub/jobs/controllers/utils/generator.py
zacharyhsia: I planed to have running job aggregation shown on the dashboard since the mock includes one. Do we need it?

zacharyhsia:
Ready to for review

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -263,20 +186,24 @@ def _metadata_response(id, job):
         extensions=extensions.get_extensions(job))
 
 
-def _provider_type():
-    return current_app.config['PROVIDER_TYPE']
+def _auth_token():
+    """Get the authentication token from the request header.
 
+        Returns:
```
calbach: nit: fix indentation

ui/src/app/dashboard/dashboard.component.html
```diff
@@ -1,18 +1,18 @@
 <jm-header [showControls]="false"></jm-header>
 
 <!--TODO(zachx): improve the grid layout-->
-<mat-grid-list cols="6" rowHeight="200px">
-  <mat-grid-tile
-    [colspan]="3"
-    [rowspan]="2">
+<!--<mat-grid-list cols="6" rowHeight="200px">-->
```
calbach: What are these changes? Does this still render?

Please just remove code instead of commenting it out.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +19,110 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+
+
+def _count_total_summary(job, total_summary):
+    if job.status not in total_summary:
+        total_summary[job.status] = 0
+    total_summary[job.status] += 1
+
+
+def _count_for_key(job, summary, key_lambda):
+    key = key_lambda(job)
```
calbach: Why the lambda? Would it be easier to just pass in the key directly?

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +19,110 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        _count_for_key(job, user_summary, lambda j: j.extensions.user_id)
+        _count_for_key(job, job_name_summary, lambda j: j.name)
+        _count_top_labels(label_summaries, job, label_freq)
+
+    aggregations = [
+        _to_aggregation('User Id', 'userId', user_summary),
+        _to_aggregation('Job Name', 'name', job_name_summary)
+    ]
+
+    _to_top_labels_aggregations(label_freq, label_summaries, aggregations)
```
calbach: In general I would not recommend the pattern of passing in a value and mutating it if it is avoidable. On the above count_ methods, this is sort of acceptable since you are accumulating things as you iterate. On this line it would be simpler to have the function return a list of aggregations.

```
aggregations = [
 ...
] + _to_top_labels_aggregations(...)
```

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +19,110 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        entries.append(
+            AggregationEntry(
+                label=item, status_counts=StatusCounts(counts_list)))
+
+    return Aggregation(name=name, key=key, entries=entries)
+
+
+def _to_top_labels_aggregations(label_freq, label_summaries, aggregations):
+    num_label = min(len(label_freq), _NUM_TOP_LABEL)
+    for k, v in sorted(
```
calbach: v is unused. I forget if you can just drop it (else maybe _)

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +19,110 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+
+        entries.append(
+            AggregationEntry(
+                label=item, status_counts=StatusCounts(counts_list)))
+
+    return Aggregation(name=name, key=key, entries=entries)
+
+
+def _to_top_labels_aggregations(label_freq, label_summaries, aggregations):
+    num_label = min(len(label_freq), _NUM_TOP_LABEL)
```
calbach: Please leave a `# TODO` to somehow truncate long value lists. For example, if there is a unique label value per job, we wouldn't want to list the value for every job. Eventually it would probably make sense to factor this into which labels we're picking as well.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +19,110 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+            counts_list.append(StatusCount(status=status, count=count))
+
+        entries.append(
+            AggregationEntry(
+                label=item, status_counts=StatusCounts(counts_list)))
+
+    return Aggregation(name=name, key=key, entries=entries)
+
+
+def _to_top_labels_aggregations(label_freq, label_summaries, aggregations):
```
calbach: There should be a comment somewhere in this file that describes the approach we're using here and why we're doing it.

@bfcrampton FYI Zach and I came up with this as an initial solution: basically pick out the top 3 highest frequency labels and aggregate by them. Later we might make this configurable, or just defer to the user to pick better labels.

bfcrampton: Seems like a good first approach to me.

servers/dsub/jobs/controllers/jobs_controller.py
zacharyhsia: Doc removed for private method.

ui/src/app/dashboard/dashboard.component.html
zacharyhsia: Removed

zacharyhsia:
Please take a look. Thanks!

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +20,109 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+    for job in jobs:
+        _count_total_summary(job, total_summary)
+        _count_for_key(job, user_summary, job.extensions.user_id)
+        _count_for_key(job, job_name_summary, job.name)
+        _count_top_labels(job, label_summaries)
+
+    aggregations = [
+        _to_aggregation('User Id', 'userId', user_summary),
+        _to_aggregation('Job Name', 'name', job_name_summary)
+    ] + (_to_top_labels_aggregations(label_summaries))
```
calbach: nit: unnecessary parens

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +20,109 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+            total_count += sum(
+                v for _, v in counts.items() if v > _LABEL_VALID_MIN)
+        label_freq[label] = total_count
+
+    aggregations = []
+
+    # To avoid messy look of dashboard, instead of returning all the aggregations by label,
+    # we only return the top _NUM_TOP_LABEL result.
+    num_label = min(len(label_freq), _NUM_TOP_LABEL)
+    for k, _ in sorted(
```
calbach: Actually I just tried this locally, seems you can just do `for k in ...`

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -24,47 +20,109 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+
+def _to_top_labels_aggregations(label_summaries):
+    # Rank the label summaries by the sum of valid item,
+    # where valid means a label has jobs more than _LABEL_VALID_MIN.
+    label_freq = {}
+    for label, item in label_summaries.items():
+        total_count = 0
+        for label_value, counts in item.items():
+            total_count += sum(
+                v for _, v in counts.items() if v > _LABEL_VALID_MIN)
```
calbach: `v for v in counts.values()`

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -1,21 +1,17 @@
...
-from jobs.models.aggregation import Aggregation
-from jobs.models.aggregation_entry import AggregationEntry
 
-from datetime import date, datetime
-from typing import List, Dict
-from six import iteritems
-from ..util import deserialize_date, deserialize_datetime
-from werkzeug.exceptions import NotImplemented
+_NUM_TOP_LABEL = 3
+_LABEL_VALID_MIN = 10
```
calbach: nit: maybe _LABEL_MIN_COUNT_FOR_RANK ?

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: `k in sorted()` dose not work. `sorted(label_freq.items(), key=lambda x: x[1], reverse=True)` gives a list of tuple. If we let `k` to be the tuple, we need to use k[0], which is not as clear as `k, _` in my opinion. And we might not use `label_freq.values()` since we need the key.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Fixed.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Accepted.

servers/dsub/jobs/controllers/aggregation_controller.py
calbach: OK, sounds good. I missed that this was a list of tuples rather than a dict.

servers/dsub/jobs/controllers/utils/time_frame.py
```diff
@@ -0,0 +1,28 @@
...
+TIME_FRAME_MAP = {
+    'HOURS_1': datetime.timedelta(minutes=60),
+    'HOURS_8': datetime.timedelta(hours=8),
+    'HOURS_24': datetime.timedelta(days=1),
+    'DAYS_7': datetime.timedelta(days=7),
+    'DAYS_30': datetime.timedelta(days=30)
+}
+
+
+def time_frame_to_start_time(time_frame):
```
bfcrampton: tiny nit: I would just call this `start_time` or `to_start_time` since the caller is doing `timeframe.time_frame_to_start_time` right now the extract "time_frame" seems unnecessary

servers/dsub/jobs/controllers/utils/time_frame.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    Map a time frame to start time.
+
```
bfcrampton: nit: remove extra new line

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -263,20 +186,14 @@ def _metadata_response(id, job):
...
-        status=job_statuses.dsub_to_api(job),
-        # The LocalJobProvider returns create-time with milliescond granularity.
-        # For consistency with the GoogleJobProvider, truncate to second
-        # granularity.
-        submission=job['create-time'].replace(microsecond=0),
-        start=job.get('start-time'),
-        end=job['end-time'],
-        labels=labels.dsub_to_api(job),
-        extensions=extensions.get_extensions(job))
+def _provider_type():
```
bfcrampton: nit: I think CH had another comment suggesting to move this into `utils/providers.py`, seems like that would be the best place for this (might have to remove the _ if its used outside of providers.py for anything). 

Could also potentially move `_auth_token()` there since I think we only use the token to authenticate dsub, could also have a new `utils/auth.py`

PR #324, opened 2018-07-10
Fix config example JSON

cjllanwarne:
I noticed some errors in the sample JSON, possibly introduced by my last PR (#307)

Note that I made the JSON syntactically valid, I haven't checked that it's a JMUI configuration (could somebody verify?)

### DIFF
--- a/servers/cromwell/README.md
+++ b/servers/cromwell/README.md
@@ -101,7 +101,7 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
     "cromwell-workflow-id",
     "workflow-name",
     "label",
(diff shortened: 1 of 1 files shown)


servers/cromwell/README.md
```diff
@@ -155,18 +155,17 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
      "workflow-name",
      "flag",
      "label",
-     "comment
+     "comment"
    ],
-   "queryExtensions": []
- },
```
cjllanwarne: In particular, is this the right surplus `}` to remove?

PR #326, opened 2018-07-12
Add deployment docs

rexwangcc:
1. Add general deployment documentation and config files. Also add docker-compose and K8s templates for Cromwell.

2. Make a change to config-loader so it supports deployments better.

3. Fix typos in shim layers' readme files.

4. Add a missing gunicorn parameter to dsub compose section.

5. Add a build status badge to the readme file.

6. Update the gitignore file.

### Note:
- I left docs for GAE for future works since I don't have any experience working with GAE.
- For dsub, only docker-compose docs are provided in this PR since I'm not sure if it is proper to deploy JMUI with dsub on k8s cluster, potentially it can be a future work as well.


README.md
```diff
@@ -1,4 +1,7 @@
 # Job Manager
+
+[![CircleCI](https://circleci.com/gh/DataBiosphere/job-manager/tree/master.svg?style=svg)](https://circleci.com/gh/DataBiosphere/job-manager/tree/master)
```
bfcrampton: Nice! 😃

README.md
```diff
@@ -169,3 +172,17 @@ For `dsub` server documentation, see [servers/dsub](servers/dsub/README.md).
 
 ## Job Manager `cromwell` Server
 For `cromwell` server documentation, see [servers/cromwell](servers/cromwell/README.md).
+
+## Build docker images and releases
+From v0.2.0, Job Manager starts to release stock docker images on [DockerHub](https://hub.docker.com/u/databiosphere)
```
bfcrampton: Not really a comment, but are we planning on publishing every version release to DockerHub?

deploy/capabilities-config.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "field": "submission",
+      "display": "Submitted"
+    },
+    {
+      "field": "end",
+      "display": "End Time"
+    },
+    {
+      "field": "labels.cromwell-workflow-id",
```
bfcrampton: Should this file (and `api-config.json`) be within the cromwell folder since its specific to that deployment?

deploy/cromwell/docker-compose/README.md
```diff
@@ -0,0 +1,9 @@
+# Deploy Job Manager against Cromwell with docker-compose
+
+1. Prepare the config files following the instructions, based on your requirements.
+
+2. Make sure your network settings are correct, e.g. DNS, TLS certs.
+
+3. Start Job Manager with `docker-compose -f cromwell-compose-template.yml up`.
+
+4. Stop the server with `docker-compose -f cromwell-compose-template.yml down`.
```
bfcrampton: Optional: add a note that you can rebuild (for dependency changes and such) with a 
`docker-compose -f cromwell-compose-template.yml build (--no-cache)`

deploy/README.md
```diff
@@ -0,0 +1,221 @@
...
+- `fieldType`, `validFieldValues`, `editable` and `bulkEditable` are currently only supported by Cromwell backend. 
+Both `editable` and `bulkEditable` will be treated as false unless explicitly set to true, if the field is 
+`editable`, then `fieldType` is required.
+- `commonLabels` controls if the field will become a predefined search key in the query builder.
+- `queryExtensions` represents those ExtendedQueryFields which are queryable, depends on your requirements. Usually the 
+default value should be fine.
+- `authentication` is optional. You should set the `isRequired` to `true` if you want to enable the Google OAuth 
+Sign-In ability.
+
+For Cromwell, the default config looks like:
```
bfcrampton: nit: to make this README shorter you could just link these example files from the deploy directory directly (I think they're the same?) Up to you

deploy/cromwell/kubernetes/nginx.conf
```diff
@@ -0,0 +1,36 @@
...
+    # This is used for health check, which is required by Kubernetes
+    location /health {
+      return 200 "healthy!\n";
+      # because default content-type is application/octet-stream,
+      # browser will by default try to save the file
+      # below line allows you to see it in the web page
+      add_header Content-Type text/plain;
+    }
+  }
+}
```
bfcrampton: tiny nit: new line

README.md
cjllanwarne: I would move this line below the "how to build" instructions, and make it more of a "now upload the results to dockerhub (ask XYZ for access if you don't have it already)"

cjllanwarne: I'd also make "Add a github release pointing to the dockerhub images" a step

README.md
```diff
@@ -169,3 +172,17 @@ For `dsub` server documentation, see [servers/dsub](servers/dsub/README.md).
...
+
+- To build the `job-manager-ui` image with `$TAG` from the root of this Github repository:
+    ```
+    cd ui && docker build -t job-manager-ui:$TAG . -f Dockerfile
+    ```
+    
+- To build the `job-manager-api` image with `$TAG` from the root of this Github repository:
+    ```
+    docker build -t job-manager-api:v0.2.0 . -f servers/cromwell/Dockerfile
```
cjllanwarne: What is `api`? Is it *both* cromwell and dsub shims bundled together? Would it sensible or completely insane to split them into two different images?

cjllanwarne: Apologies for the naive question but would it make sense to push a "composed" docker image with UI and API together? 

Or is it just as simple to compose the two dockerhub images for an end-user at "I want to try this out" time?

deploy/README.md
cjllanwarne: +1 to linking out rather than copy/pasting

deploy/README.md
```diff
@@ -0,0 +1,221 @@
...
+  "commonLabels": [
+    "cromwell-workflow-id",
+    "label",
+    "flag"
+  ],
+  "queryExtensions": []
+}
+```
+
+For dsub, depends on the value of the environment variable `PROVIDER_TYPE`, the full version default config looks like:
```
cjllanwarne: Didn't you say earlier that "this isn't relevant for `dsub`?

deploy/cromwell/docker-compose/README.md
```diff
@@ -0,0 +1,9 @@
+# Deploy Job Manager against Cromwell with docker-compose
+
+1. Prepare the config files following the instructions, based on your requirements.
```
cjllanwarne: Following what instructions? Could we link to them?

deploy/cromwell/docker-compose/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+2. Make sure your network settings are correct, e.g. DNS, TLS certs.
+
+3. Start Job Manager with `docker-compose -f cromwell-compose-template.yml up`.
```
cjllanwarne: From where?

deploy/cromwell/kubernetes/nginx.conf
cjllanwarne: +1

cjllanwarne: (ps if you use intellij or pycharm to make these: https://stackoverflow.com/questions/16761227/how-to-make-intellij-idea-insert-a-new-line-at-every-end-of-file )

README.md
rexwangcc: Maybe @cjllanwarne here. @bfcrampton I don't think we have reached any team agreements on this, at least for Cromwell, this might be the plan(I'm not sure).

rexwangcc: That sounds better, will do.

README.md
rexwangcc: This is a good catch! Actually, it was my fault that I totally forgot the dsub image. In fact, dsub and Cromwell shim layers have their own separate Dockerfiles(and in turn separate images), so we **should** have 
- job-manager-api-cromwell:v0.2.0
- job-manager-api-dsub:v0.2.0

separately.  I need to make changes throughout this PR for this. The only concern of mine is that thus we will have 3 docker images/docker repos just for Job Manager, do you think this would be annoying to users? @cjllanwarne

rexwangcc: I prefer the latter way, since consolidating the 2 images into one is not a simple work to me, besides, I think the UI and shim were designed to be separate components (e.g. deploying them on GAE can get some benefits from this kind of isolation).

deploy/README.md
rexwangcc: Sounds better, will do!

deploy/README.md
rexwangcc: Sorry for the confusion here. Yeah, the `capabilities-config.json` is not relevant for `dsub`, since this part of in dsub is hard-coded in python code: https://github.com/DataBiosphere/job-manager/blob/1f549d20be568aa840e8553669e5bdfdab89c744/servers/dsub/jobs/controllers/capabilities_controller.py#L16 

I thought it would be nice to "translate" the python code as a default `capabilities.json` here so users could be aware of what values they will get as a default. Speaking of this, @bfcrampton I propose we make changes to the dsub controller so it can also accept JSON config files, for both flexibility and consistency reasons.

deploy/capabilities-config.json
rexwangcc: Good catch, will move. Besides, as I said above, do you think it would be a good idea to make changes to dsub shim so it can also accept JSON capabilities configs?

deploy/cromwell/docker-compose/README.md
rexwangcc: 👍 Will link to the other readmes.

README.md
cjllanwarne: Since as I understand it our main release artifact is a set of docker containers (and hopefully eventually a startup/shutdown script to orchestrate them together), it makes sense to me that every release would be pushed to a docker repo.

I'm not *especially* wed to Dockerhub in particular but I do think it should be cloud-agnostic and dockerhub has the benefit of being the hegemon right now. And Cromwell is on dockerhub too which also makes it a nice default choice.

deploy/README.md
rexwangcc: Actually I realized that the code snippets in the readme file serve purposes in the documentation from the file examples, will keep it there but add more description to clarify it.

README.md
rexwangcc: Fixed now.

rexwangcc:
Thanks for all of the comments, I just pushed new changes based on the suggestions.

@cjllanwarne 
> One question - I assume that all of these instructions currently rely on the repo being checked out, rather than pulling directly from the released images on dockerhub? 

Actually the docs in this PR is relied on pulling images from docker hub directly, users don't have check out this repo if they follow the instructions, which means right after merging this PR in, we need to cut off a new release and have the docker images ready on docker hub.

deploy/README.md
bfcrampton: Yup agreed, we should definitely do that for consistency.

rexwangcc: Cool, going to create an issue for it, and will make a PR in this week.

rexwangcc:
If there are no objections, I'm going to merge in this PR by today, and we can always have more iterations on top of this later to make the instruction more straight-forward and user-friendly.

PR #331, opened 2018-07-12
Add support for workflow "on hold" status 

rsasch:
See https://elastc.com/c/1u-Xut8k/329-adapt-to-new-cromwell-status or #332 

**NB**: I was not able to add the custom SVG icon in the mockups; used an out-of-the-box Clarity icon, instead.


ui/src/app/shared/common.ts
```diff
@@ -7,7 +7,8 @@ export enum JobStatusIcon {
   Aborting = 'exclamation-triangle',
   Failed = 'times',
   Succeeded = 'check',
-  Aborted = 'exclamation-triangle'
+  Aborted = 'exclamation-triangle',
+  'On Hold' = 'pause'
```
aednichols: An identifier with a space in it looks really weird to me, but perhaps it's practical and/or conventional in Javascript-land

ui/src/app/shared/header/header.component.ts
```diff
@@ -220,6 +221,10 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
     this.navigateWithStatus(this.failedStatuses.slice());
   }
 
+  showOnHoldJobs(): void {
+    this.navigateWithStatus(this.onHoldStatuses.slice());
```
aednichols: It might be nice to represent the statuses with algebraic datatypes instead of lists that we slice around in - i.e. an `Active` trait mixed into classes `Submitted`, `Running`, and`Aborting` but `OnHold` is just a standalone class since there's only one possibility

I don't actually know if this is possible in Typescript, but the "Type" part makes me think yes

geoffjentry: I think I'd suggest that if one is going to suggest using things which would require actual language constructs that one research the existence of those constructs prior to making the advice :)

aednichols: [Seems like yes](https://www.typescriptlang.org/docs/handbook/mixins.html).

They totally anticipated me coming over from Scala and immediately wanting this in Typescript ("You may be familiar with the idea of mixins or traits for languages like Scala")

servers/cromwell/jobs/swagger/swagger.yaml
```diff
@@ -475,6 +475,7 @@ definitions:
     - "Failed"
     - "Succeeded"
     - "Aborted"
+    - "On Hold"
```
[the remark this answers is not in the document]
calbach: jobs.yaml is the truth, the others are autogenerated (this is necessary) and checked in (not necessary, but we'd need to do work to fix that). There's a ticket somewhere to not generate them twice, and as part of that we probably don't want to be checking in the generated code.

ui/src/app/shared/header/header.component.html
```diff
@@ -49,6 +49,9 @@
         <button mat-raised-button class="status-button completed-button" (click)="showCompletedJobs()">
           Completed <span *ngIf="shouldDisplayStatusCounts()">({{ getCompletedCount() }})</span>
         </button>
+        <button mat-raised-button class="status-button onhold-button" (click)="showOnHoldJobs()">
+          On Hold <span *ngIf="shouldDisplayStatusCounts()">({{ getOnHoldCount() }})</span>
```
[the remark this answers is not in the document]
rsasch: `shouldDisplayStatusCounts` checks to see if all jobs are loaded client-side (I'm guessing because that means the counts can be accurate)

ui/src/app/shared/header/header.component.ts
rsasch: This was modeled the existing handling of other statuses for code consistency, but I'm open to other implementations.

aednichols: It's not obvious to me what a no-parameter invocation of `slice()` should do, hence my attention. But this is definitely a "thinking out loud" type of discussion rather than something @rsasch needs to change!

api/jobs.yaml
```diff
@@ -458,6 +458,7 @@ definitions:
       - Failed
       - Succeeded
       - Aborted
+      - On Hold
```
samanehsan: Can you include the regenerated `JobStatus` models for Cromwell and dsub?

ui/src/app/shared/model/JobStatus.ts
```diff
@@ -18,5 +18,6 @@ export enum JobStatus {
     Aborting = <any> 'Aborting',
     Failed = <any> 'Failed',
     Succeeded = <any> 'Succeeded',
-    Aborted = <any> 'Aborted'
+    Aborted = <any> 'Aborted',
+    OnHold = <any> 'On Hold'
```
samanehsan: Why does `OnHold` not have a space here like the model definition does?

api/jobs.yaml
rsasch: they weren't regenerated when I ran the swagger codgen, and it doesn't look like `dsub/jobs/models/job_status.py` or  `cromwell/jobs/models/job_status.py` specify possible values

ui/src/app/shared/model/JobStatus.ts
rsasch: I assume it's a swagger codgen thing.

samanehsan: ok!

api/jobs.yaml
calbach: I would definitely not recommend putting a space in an enum value. I'd just go with `OnHold`.

Separately, what does OnHold mean and how does it relate to the other statuses? Depending on what it means, I'd suggest moving this up into its logical chronological order for readability (guessing before or after running?)

ui/src/app/shared/header/header.component.ts
calbach: FYI: It's a fairly common pattern in JS to do a shallow copy of an array.

IIRC there was a bug with naviagateWithStatuses() mutating its parameter. This could probably be cleaned up, e.g. by not mutating the parameter, but I'm foggy on the details.

api/jobs.yaml
rsasch: The status in Cromwell has the space in it. If the `status` in JM is not exactly the same as the `status` submitted to Cromwell, where in the code is the best place to make that translation?

calbach: I would add a helper and invoke here and here:

https://github.com/DataBiosphere/job-manager/blob/master/servers/cromwell/jobs/controllers/jobs_controller.py#L118
https://github.com/DataBiosphere/job-manager/blob/master/servers/cromwell/jobs/controllers/jobs_controller.py#L279

rsasch: what about updating and using the below existing code?

https://github.com/DataBiosphere/job-manager/blob/master/servers/cromwell/jobs/controllers/utils/task_statuses.py
https://github.com/DataBiosphere/job-manager/blob/master/servers/dsub/jobs/controllers/utils/job_statuses.py

calbach: The first are task statuses. I still don't know what ON HOLD is, so I definitely don't know whether this is a task status, or just a workflow status (I couldn't find this status in the Cromwell docs either).

Good catch on the second. It's probably not going to matter since dsub doesn't have anything to map to ON_HOLD, but probably worth adding the reverse mapping ON_HOLD -> RUNNING just in case.

rexwangcc: @calbach "On Hold" is a pretty new workflow status, which indicates why the docs haven't been updated to include it: https://github.com/broadinstitute/cromwell/blob/develop/CHANGELOG.md#api

The first file is actually a mapping between JMUI status and Cromwell task execution statuses, @rsasch you might want to change it to `job_statuses.py` and add your own mapping to it.

ui/src/app/shared/common.ts
samanehsan: Should the space get removed here as well?

servers/cromwell/jobs/controllers/utils/job_statuses.py
```diff
@@ -0,0 +1,43 @@
...
+    ABORTING='Aborting',
+    FAILED='Failed',
+    SUCCEEDED='Succeeded')
+
+API_WORKFLOW_STATUS_MAP = {
+    ApiStatus.SUBMITTED: CromwellWorkflowStatus.SUBMITTED,
+    ApiStatus.ONHOLD: CromwellWorkflowStatus.ONHOLD,
+    ApiStatus.RUNNING: CromwellWorkflowStatus.RUNNING,
+    ApiStatus.ABORTING: CromwellWorkflowStatus.ABORTING,
+    ApiStatus.ABORTED: CromwellWorkflowStatus.FAILED,
```
samanehsan: Cromwell has an `Aborted` status, so I think this should map to that instead of `Failed`

servers/cromwell/jobs/controllers/utils/job_statuses.py
```diff
@@ -0,0 +1,43 @@
...
+    SUBMITTED='Submitted',
+    ONHOLD='OnHold',
+    RUNNING='Running',
+    ABORTING='Aborting',
+    ABORTED='Aborted',
+    FAILED='Failed',
+    SUCCEEDED='Succeeded')
+CromwellWorkflowStatus = enum(
+    SUBMITTED='Submitted',
+    ONHOLD='On Hold',
```
calbach: nit: more normal to use an underscore when you're doing ALL_CAPS casing.

ui/src/app/shared/header/header.component.ts
```diff
@@ -59,6 +59,7 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
   private readonly activeStatuses = [JobStatus.Submitted, JobStatus.Running, JobStatus.Aborting];
   private readonly completedStatuses = [JobStatus.Succeeded, JobStatus.Aborted];
```
cjllanwarne: I would put `Aborted` under `failedStatuses` rather than `completedStatuses`

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -249,7 +251,9 @@ def cromwell_query_params(query, page, page_size):
     if query.name:
         query_params.append({'name': query.name})
     if query.statuses:
-        statuses = [{'status': s} for s in set(query.statuses)]
+        statuses = [{
+            'status': job_statuses.api_workflow_status_to_cromwell(s)
+        } for s in set(query.statuses)]
```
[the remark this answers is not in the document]
rexwangcc: For something like: `[{'status': "On Hold"}, {'status': "Running"}]`, the result of the "single map" depends on how you flatten this list. Currently, we basically pass those 2 dictionaries(maps) to Cromwell's `/query` directly here.

servers/cromwell/jobs/controllers/utils/job_statuses.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def cromwell_workflow_status_to_api(cromwell_status):
+    """ Map an API workflow status to a Cromwell status.
+
+  :param str cromwell_status: 'Submitted', 'OnHold', 'Aborting', 'Failed', 'Succeeded'
+  :return:  API status 'Submitted', 'On Hold', 'Running', 'Aborting', 'Failed' or 'Succeeded'
+  """
+    if cromwell_status not in CROMWELL_STATUS_MAP:
+        raise ValueError('Unrecognized Cromwell execution status: {}'.format(
```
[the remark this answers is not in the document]
rexwangcc: 3 layers work together as the shim layer here: `controllers`(functions) -> `connexion` (API framework built on top of the Flask) -> `Gunicorn` Server.

So, if an exception in the controller has been thrown, e.g.:
```
cromwell_1    | ERROR:flask.app:Exception on /api/v1/jobs/query [POST]
cromwell_1    | Traceback (most recent call last):
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 2292, in wsgi_app
cromwell_1    |     response = self.full_dispatch_request()
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1815, in full_dispatch_request
cromwell_1    |     rv = self.handle_user_exception(e)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1718, in handle_user_exception
cromwell_1    |     reraise(exc_type, exc_value, tb)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1813, in full_dispatch_request
cromwell_1    |     rv = self.dispatch_request()
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1799, in dispatch_request
cromwell_1    |     return self.view_functions[rule.endpoint](**req.view_args)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/connexion/decorators/decorator.py", line 66, in wrapper
cromwell_1    |     response = function(request)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/connexion/decorators/validation.py", line 122, in wrapper
cromwell_1    |     response = function(request)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/connexion/decorators/validation.py", line 293, in wrapper
cromwell_1    |     return function(request)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/connexion/decorators/decorator.py", line 42, in wrapper
cromwell_1    |     response = function(request)
cromwell_1    |   File "/usr/local/lib/python2.7/dist-packages/connexion/decorators/parameter.py", line 195, in wrapper
cromwell_1    |     return function(**kwargs)
cromwell_1    |   File "/home/vmagent/app/jobs/controllers/utils/auth.py", line 14, in wrapper
cromwell_1    |     return fn(*args, **kwargs)
cromwell_1    |   File "/home/vmagent/app/jobs/controllers/jobs_controller.py", line 210, in query_jobs
cromwell_1    |     raise ValueError("I want an exception!")
cromwell_1    | ValueError: I want an exception!
```
The exception will be passed from controller to conexion and then to the low-level Flask framework, and Gunicorn will return a 500 error to the UI, which means the users will see `500: Internal Server Error` when a servre exception happens. 

I would say we need to use `try...except...finally...` as much as possible in controllers, to catch and mitigate those predictable exceptions at the 1st place. Furthermore, Flask/connexion supports adding exception handlers, which could be responsible for catching more exceptions. Finally, I think it would be better if UI could handle those 5xx errors more peacefully (like show a beautiful page or just snackBar) v.s. bare error page.

I'm not sure what is the equivalent in Python to the scala `scala.util.Try`, but try blocks + Exception handlers should be sufficient I think?

Issue #334, opened 2018-07-13
Update the mapping of task mappings for Cromwell 32+


rexwangcc:
According to [here](https://github.com/broadinstitute/cromwell/blob/32/core/src/main/scala/cromwell/core/ExecutionStatus.scala), from Cromwell 32(+), there is a new internal status called `WaitingForQueueSpace` as an execution status of tasks. Without updating the shim layer's mapping there will be a small chance the icon in the UI will break.

rexwangcc:
Addressed in #337

PR #337, opened 2018-07-13
Update the task status mappings.

rexwangcc:
This PR addresses the issue #334 .

### DIFF
--- a/servers/cromwell/jobs/controllers/utils/task_statuses.py
+++ b/servers/cromwell/jobs/controllers/utils/task_statuses.py
@@ -44,7 +44,7 @@ CROMWELL_EXECUTION_STATUS_MAP = {
 def cromwell_execution_to_api(execution_status):
     """ Map a Cromwell execution status to an API status.
 
-    :param str execution_status: 'NotStarted', 'QueuedInCromwell', 'Starting', 'Running', 'Aborting', 'Failed',
+    :param str execution_status: 'NotStarted', 'WaitingForQueueSpace', 'QueuedInCromwell', 'Starting', 'Running', 'Aborting', 'Failed',
      'RetryableFailure', 'Done', 'Bypassed', 'Aborted', or 'Unstartable'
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/utils/task_statuses.py
```diff
@@ -26,6 +27,7 @@
 
 CROMWELL_EXECUTION_STATUS_MAP = {
     CromwellExecutionStatus.NOTSTARTED: ApiStatus.SUBMITTED,
+    CromwellExecutionStatus.WAITINGFORQUEUESPACE: ApiStatus.SUBMITTED,
```
samanehsan: Please update the docstring below as well to include this status: https://github.com/DataBiosphere/job-manager/blob/e49012d6a068a9989e6b88b249f1cbdb1903e51d/servers/cromwell/jobs/controllers/utils/task_statuses.py#L47

rexwangcc: Good catch! will do

servers/cromwell/jobs/test/test_task_statuses.py
```diff
@@ -0,0 +1,29 @@
...
+from __future__ import absolute_import
+
+from . import BaseTestCase
+
+from jobs.controllers.utils import task_statuses
+
+
+class TestTaskStatuses(BaseTestCase):
+    # yapf: disable
+    def test_cromwell_execution_to_api_maps_all_execution_statuses_correctly(self):
```
samanehsan: This test seems unnecessary to me since it's just testing a dictionary look-up in `CROMWELL_EXECUTION_STATUS_MAP`

rexwangcc:
@samanehsan I still think it's not harmful to keep the test cases there, even they seem to be unnecessary for now. What do you think?

samanehsan:
Yeah I think that's fine, @rexwangcc!

PR #364, opened 2018-07-17
Add two tests for aggregation API

zacharyhsia:


zacharyhsia:
Not sure where should I put the global config parameter. Feel like there is a bunch of hard code in testing. Please help improve it, thanks!

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -94,22 +93,31 @@ def _to_aggregation(name, key, summary):
...
-        for label_value, counts in item.items():
+        for _, counts in item.items():
             total_count += sum(
-                v for v in counts.values() if v > _LABEL_MIN_COUNT_FOR_RANK)
+                v for v in counts.values() if v > _min_count_for_rank())
         label_freq[label] = total_count
 
     aggregations = []
 
+    # Jobs with aggregation-testing-unique has higher priority
```
calbach: I was suggesting your test would set something like:

```
app.app.config['AGGREGATION_FILTER_LABEL'] = "aggregation-testing-unique:<unique_test_value>"
```

Normally you would not set this config value, but if present, you pass it through to the job generator as a label filter. This option would likely never be used outside of testing, but is at least generic.

See test_token_label in the google test for a similar approach.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -64,6 +66,8 @@ def create_app(self):
             app = connexion.App(__name__, specification_dir='../swagger/')
             app.app.json_encoder = JSONEncoder
             app.add_api('swagger.yaml')
+            app.app.config[
+                'LABEL_MIN_COUNT_FOR_RANK'] = _LABEL_MIN_COUNT_FOR_RANK
```
calbach: 1 line?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +453,66 @@ def test_query_jobs_submission_pagination(self):
                     page_token=response.next_page_token,
                     extensions=ExtendedQueryFields(submission=min_time)),
                 [job2])
+
+        def test_aggregtion_jobs(self):
```
calbach: Spelling.

Also nit: would probably name this the same as the method, test_get_job_aggregations

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -18,8 +18,10 @@
 from jobs.models.job_metadata_response import JobMetadataResponse
 from jobs.models.query_jobs_response import QueryJobsResponse
 from jobs.models.query_jobs_request import QueryJobsRequest
+from jobs.models.aggregation_response import AggregationResponse
 
 DOCKER_IMAGE = 'ubuntu:14.04'
+_LABEL_MIN_COUNT_FOR_RANK = 3
```
calbach: I would just set this to 0, I think the logic in this test is too complicated otherwise

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +453,66 @@ def test_query_jobs_submission_pagination(self):
...
+            self.wait_status(job_failed, ApiStatus.SUCCEEDED)
+
+            # Create jobs with specified labels above times that are large enough to
+            # be ranked top and thus appear in the Aggregation
+            for i in xrange(_LABEL_MIN_COUNT_FOR_RANK):
+                self.start_job('echo LABEL', labels=labels, name=name)
+
+            aggregationResponse = self.must_aggregate_job('HOURS_1')
+
+            for aggregation in aggregationResponse.aggregations:
```
calbach: This feels a bit too dynamic. In general you want to form the expected response in code here, e.g. using the AggregationResponse model, then compare it. I'm not sure whether you can check equality directly on this - you may need to call `to_dict()` first.

servers/dsub/jobs/test/test_jobs_controller_google.py
```diff
@@ -115,6 +115,15 @@ def test_query_jobs_by_start(self):
...
         self.assert_query_matches(QueryJobsRequest(start=date), [job])
 
+    def test_aggregtion_jobs_invalid_project_id(self):
+        time_frame = 'DAYS_7'
+        project_id = 'should-be-an-invalid-id'
+        resp = self.client.open(
+            '/aggregations?projectId={}&timeFrame={}'.format(
+                project_id, time_frame),
+            method='GET')
+        self.assert_status(resp, 500)
```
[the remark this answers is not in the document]
zacharyhsia: Added BadRequest exception when project id is needed.

servers/dsub/jobs/test/base_test_cases.py
zacharyhsia: Seems like a yapf thing. Fixed.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -101,6 +105,17 @@ def must_query_jobs(self, parameters):
             self.assert_status(resp, 200)
             return QueryJobsResponse.from_dict(resp.json)
 
+        def must_aggregate_job(self, time_frame):
```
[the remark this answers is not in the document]
zacharyhsia: Accepted.

servers/dsub/jobs/test/base_test_cases.py
zacharyhsia: Accepted.

servers/dsub/jobs/controllers/aggregation_controller.py
zacharyhsia: Changed to 
```
app.app.config[
                'AGGREGATION_FILTER_LABEL'] = "aggregation-testing-unique:" + datetime.datetime.now().strftime('%Y%m%d_%H%M')
```

servers/dsub/jobs/test/base_test_cases.py
zacharyhsia: By using the filter label, we do not need to change this value when testing, reverting it to the last version.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +453,66 @@ def test_query_jobs_submission_pagination(self):
...
+            self.wait_status(job_aborted, ApiStatus.RUNNING)
+            self.must_abort_job(job_aborted)
+            self.wait_status(job_aborted, ApiStatus.ABORTED)
+
+            # Create a failed job
+            job_failed = self.api_job_id(
+                self.start_job('not_a_command', labels=labels, name=name))
+            self.wait_status(job_failed, ApiStatus.FAILED)
+
+            # Created a succeeded job
```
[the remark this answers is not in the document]
zacharyhsia: Change the **Command** to `sleep 30`

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +453,66 @@ def test_query_jobs_submission_pagination(self):
...
+            job_failed = self.api_job_id(
+                self.start_job('echo SUCCEEDED', labels=labels, name=name))
+            self.wait_status(job_failed, ApiStatus.SUCCEEDED)
+
+            # Create jobs with specified labels above times that are large enough to
+            # be ranked top and thus appear in the Aggregation
+            for i in xrange(_LABEL_MIN_COUNT_FOR_RANK):
+                self.start_job('echo LABEL', labels=labels, name=name)
```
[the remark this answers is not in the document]
zacharyhsia: Seems that we cannot guarantee a job that is in status of `Submitted`. Only test the other four statuses for now.

servers/dsub/jobs/test/base_test_cases.py
zacharyhsia: Accepted.

servers/dsub/jobs/test/test_jobs_controller_google.py
```diff
@@ -115,6 +115,15 @@ def test_query_jobs_by_start(self):
         self.wait_status(self.api_job_id(job), ApiStatus.RUNNING)
         self.assert_query_matches(QueryJobsRequest(start=date), [job])
 
+    def test_aggregtion_jobs_invalid_project_id(self):
```
calbach: Typo still

servers/dsub/jobs/test/test_jobs_controller_google.py
```diff
@@ -115,6 +115,15 @@ def test_query_jobs_by_start(self):
...
+        time_frame = 'DAYS_7'
+        project_id = 'should-be-an-invalid-id'
+        resp = self.client.open(
+            '/aggregations?projectId={}&timeFrame={}'.format(
+                project_id, time_frame),
+            method='GET')
+        self.assert_status(resp, 400)
```
calbach: A 400 is fine if the requestor doesn't provide a projectId. But this test passes a project ID that doesn't exist. It should be expecting a 404 or 403 (see other tests).

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -64,6 +66,9 @@ def create_app(self):
             app = connexion.App(__name__, specification_dir='../swagger/')
             app.app.json_encoder = JSONEncoder
             app.add_api('swagger.yaml')
+            app.app.config[
+                'AGGREGATION_FILTER_LABEL'] = "aggregation-testing-unique:" + datetime.datetime.now(
+                ).strftime('%Y%m%d_%H%M')
```
calbach: I would put the second granularity at least. It's definitely possible that one person or multiple people might run this test twice in a minute.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,83 @@ def test_query_jobs_submission_pagination(self):
...
+                }, {
+                    'status': u'Aborted',
+                    'count': 1
+                }]
+            })
+
+            aggregation_resp = self.must_get_job_aggregations('HOURS_1')
+
+            self.assertEqual(
+                self.status_counts_equal(aggregation_resp.summary,
```
calbach: Just do `assert_status_counts_equal()`, and put the asserts in there

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,83 @@ def test_query_jobs_submission_pagination(self):
...
+            counts2Dict = {}
+            for status_count in counts2.counts:
+                counts2Dict[status_count.status] = status_count.count
+
+            return counts1Dict == counts2Dict
+
+        def test_get_job_aggregations(self):
+            name = flask.current_app.config['AGGREGATION_FILTER_LABEL']
+
+            labels = {
```
calbach: I would put different label values on these, and possibly different keys entirely. Currently there's exactly 1 value+key.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,83 @@ def test_query_jobs_submission_pagination(self):
...
+                    self.assertEqual(
+                        self.status_counts_equal(entry.status_counts,
+                                                 status_counts), True)
+
+            # Should have labels above presented in the aggregation response
+            labels = [
+                aggregation.key
+                for aggregation in aggregation_resp.aggregations
+            ]
+            self.assertEqual('aggregation-testing-unique-1' in labels, True)
```
calbach: Should list out the full list of expected keys here, and then do `assertItemsEqual()`. Currently there might be extra stuff returned.

servers/dsub/jobs/__main__.py
```diff
@@ -64,6 +64,7 @@ def split_env_flag(name):
 app = connexion.App(__name__, specification_dir='./swagger/', swagger_ui=False)
 app.app.config['PROVIDER_TYPE'] = args.provider_type
 app.app.config['REQUIRES_AUTH'] = bool(args.requires_auth)
+app.app.config['LABEL_MIN_COUNT_FOR_RANK'] = 10
```
calbach: I don't follow why this is no longer needed, but if it's not needed you should be able to revert this.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -33,6 +38,12 @@ def get_job_aggregations(timeFrame, projectId=None):
     label_summaries = {}
 
     for job in jobs:
+        # AGGREGATION_FILTER_LABEL is a global config parameter only set for aggregation testing
+        # and jobs that do not have this label is discarded when testing
```
calbach: grammar nit: is -> are

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -33,6 +38,12 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+        if 'AGGREGATION_FILTER_LABEL' in current_app.config and job.name.find(
+                current_app.config['AGGREGATION_FILTER_LABEL']) < 0:
```
calbach: I was expecting the filter to be on a label key+value, but I guess name works too if that's easier.

If you're using the name though, you should rename this config value to something like "AGGREGATION_JOB_NAME_FILTER"

servers/dsub/jobs/__main__.py
```diff
@@ -64,6 +64,7 @@ def split_env_flag(name):
 app = connexion.App(__name__, specification_dir='./swagger/', swagger_ui=False)
 app.app.config['PROVIDER_TYPE'] = args.provider_type
 app.app.config['REQUIRES_AUTH'] = bool(args.requires_auth)
+
```
calbach: nit: just revert all changes to this file since you're no longer changing it in this PR.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -32,11 +36,20 @@ def get_job_aggregations(timeFrame, projectId=None):
...
     label_summaries = {}
 
-    for job in jobs:
-        _count_total_summary(job, total_summary)
-        _count_for_key(job, user_summary, job.extensions.user_id)
-        _count_for_key(job, job_name_summary, job.name)
-        _count_top_labels(job, label_summaries)
+    try:
+        for job in jobs:
+            # AGGREGATION_FILTER_LABEL is a global config parameter only set for aggregation testing
```
calbach: Update constant name

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,146 @@ def test_query_jobs_submission_pagination(self):
...
                     extensions=ExtendedQueryFields(submission=min_time)),
                 [job2])
+
+        def assert_status_counts_equal(self, counts1, counts2):
+            counts1Dict = {}
+            for status_count in counts1.counts:
+                counts1Dict[status_count.status] = status_count.count
+
+            counts2Dict = {}
+            for status_count in counts2.counts:
```
calbach: nit: should be able to replace these with dictionary comprehensions

`{sc.status: sc.count for sc in counts2.counts}`

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,146 @@ def test_query_jobs_submission_pagination(self):
...
+
+            # Created a succeeded job
+            job_failed = self.api_job_id(
+                self.start_job('echo SUCCEEDED', labels=labels, name=name))
+            self.wait_status(job_failed, ApiStatus.SUCCEEDED)
+
+            # Created a running job
+            job_running = self.api_job_id(
+                self.start_job('sleep 30', labels=labels, name=name))
+            self.wait_status(job_running, ApiStatus.RUNNING)
```
calbach: I would start all of these jobs first, then wait for all of them.

You should also bump up the sleep higher than, probably more like 180. I could imagine it taking >30s before you actually call the aggregations.

Ideally you also try to cancel all of these after the test runs (you can ignore failures there).

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,146 @@ def test_query_jobs_submission_pagination(self):
...
+                                            status_counts_total)
+
+            for aggregation in aggregation_resp.aggregations:
+                if aggregation.key == 'name' or aggregation.key == 'userId':
+                    expected_counts = status_counts_total
+                else:
+                    expected_counts = status_counts_group
+
+                for entry in aggregation.entries:
+                    # print(entry.label)
```
calbach: remove

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,146 @@ def test_query_jobs_submission_pagination(self):
...
+            self.wait_status(job_running, ApiStatus.RUNNING)
+
+            # Create a running job that has different labels
+            job_running = self.api_job_id(
+                self.start_job('sleep 30', labels=different_labels, name=name))
+            self.wait_status(job_running, ApiStatus.RUNNING)
+
+            status_counts_total = StatusCounts.from_dict({
+                'counts': [{
+                    'status': u'Running',
```
calbach: I don't think you need the `u` here

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +454,146 @@ def test_query_jobs_submission_pagination(self):
...
+                    'count': 1
+                }]
+            })
+
+            aggregation_resp = self.must_get_job_aggregations('HOURS_1')
+            self.assert_status_counts_equal(aggregation_resp.summary,
+                                            status_counts_total)
+
+            for aggregation in aggregation_resp.aggregations:
+                if aggregation.key == 'name' or aggregation.key == 'userId':
```
calbach: In general it is clearer to avoid so much dynamic code in testing. Specifically just generating a single expected aggregations response above would make things much more readable IMO.

You can still make local helpers like `status_counts_single`.

```
status_counts_single =...
status_counts_group = ...
expected_response = AggregationResponse.from_dict({
  'counts': [
    {status: 'Running', count: 3},
    ...
  ],
  aggregations: [{
     key: 'name',
     entries: {
       label: name,
       status_counts: status_counts_total
     },
  }, {
    ...
  }]
})

assert_aggregation_response_equal(aggregation_resp, expected_response)
```

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -97,17 +110,18 @@ def _to_top_labels_aggregations(label_summaries):
     # where valid means a label has jobs more than _LABEL_MIN_COUNT_FOR_RANK.
     label_freq = {}
     for label, item in label_summaries.items():
+        # job-id label contains one job per label value, filter it out
```
bfcrampton: This is not strictly true––one dsub job can have many tasks each of which is represented as a distinct job here. I'm fine with filtering it as an aggregation key though, would just update the comment: "Do not use the job-id label as an aggregation key".

Also, do you also want to filter out the 'task-id' label here?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +453,66 @@ def test_query_jobs_submission_pagination(self):
...
+                'aggregation-testing-unique-3': 'yeah'
+            }
+
+            # Add timestamp to distinguish from old test jobs
+            name = 'aggregation-testing' + str(datetime.datetime.now().minute)
+
+            # Create an aborted job
+            job_aborted = self.api_job_id(
+                self.start_job('sleep 30', labels=labels, name=name))
+            time.sleep(10)
```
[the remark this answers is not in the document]
bfcrampton: Generally we'd want to split out provider-specific code into the two different test files (local and google), though it would be unfortunate to duplicate this whole test in both places just to remove the sleep for google.

I'd suggest you add an abstract method to `BaseTestCases` called `start_running_job` and then have a separate local and google implementation. 

Local:
```
def start_running_job(self, command, name=None, labels={}):
  self.start_job(command, name=name, labels=labels):
  time.sleep(10)
  self.wait_status(job_aborted, ApiStatus.RUNNING)
```

Google:
```
def start_running_job(self, command, name=None, labels={}):
  self.start_job(command, name=name, labels=labels):
  self.wait_status(job_aborted, ApiStatus.RUNNING)
```


Then this method can just call `self.start_running_job(...)`. Additionally, the separate `test_abort_job` tests in `test_jobs_controller_local` and `test_jobs_controller_google` can be refactored to use this method and put in this file.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -32,11 +36,24 @@ def get_job_aggregations(timeFrame, projectId=None):
...
 
-    for job in jobs:
-        _count_total_summary(job, total_summary)
-        _count_for_key(job, user_summary, job.extensions.user_id)
-        _count_for_key(job, job_name_summary, job.name)
-        _count_top_labels(job, label_summaries)
+    try:
+        for job in jobs:
+
+            # TEST_TOKEN_VALIE is a global config value used to distinguish testing jobs from batch to batch by timestemp
```
calbach: VALIE -> VALUE

calbach: timestamp

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -32,11 +36,24 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+            if 'TEST_TOKEN_VALUE' in current_app.config and (
+                    'test_token' not in job.labels or job.labels['test_token']
```
calbach: Don't really need this key check, can just do: `or job.labels.get('test_token') != ...`

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -32,11 +36,24 @@ def get_job_aggregations(timeFrame, projectId=None):
...
+                    != current_app.config['TEST_TOKEN_VALUE']):
```
calbach: I really don't want anything named "TEST" in the production code. Setting that aside - why do you need both this filter AND the aggregation job name filter? Isn't the name filter sufficient?

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -97,17 +114,18 @@ def _to_top_labels_aggregations(label_summaries):
     # where valid means a label has jobs more than _LABEL_MIN_COUNT_FOR_RANK.
     label_freq = {}
     for label, item in label_summaries.items():
+        # Do not use the job-id or task-id label as aggregation key
```
calbach: nit: in general comments should be used to explain "why" rather than "what"; this current comment basically just restates the code. Something like "job-id and task-id are assigned by dsub and are not meaningful for aggregation"

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
...
+
+            another_value = {
+                'aggregation-testing-unique-1': 'different-value',
+            }
+
+            different_labels = {'different-label': 'different-value'}
+
+            labels_no_show = {'no-show': 'cannot-see-me'}
+
+            # Do not change the order of creating jobs since the labels are updated
```
calbach: Is this true? I don't see the labels being updated anywhere

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
...
+            self.wait_status(job_aborted, ApiStatus.RUNNING)
+            self.must_abort_job(job_aborted)
+
+            self.wait_status(job_aborted, ApiStatus.ABORTED)
+
+            self.wait_status(job_failed, ApiStatus.FAILED)
+
+            self.wait_status(job_succeeded, ApiStatus.SUCCEEDED)
+
+            self.wait_status(job_running, ApiStatus.RUNNING)
```
calbach: How about accumulating these running jobs in a list, then waiting for all, then aborting all

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
                     page_token=response.next_page_token,
                     extensions=ExtendedQueryFields(submission=min_time)),
                 [job2])
+
+        def status_counts_to_dict(self, status_counts):
+            status_counts_dict = {}
```
calbach: opt: can you use a dictionary comprehension for this?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -111,7 +137,11 @@ def start_job(self,
                       outputs={},
                       outputs_recursive={},
                       task_count=1,
-                      wait=False):
+                      wait=False,
+                      with_test_token=False):
```
calbach: Wouldn't `True` be a more reasonable default here? Isn't your test the only one that would want this to be false?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
...
+            status_counts_group = {
+                'Running': 1,
+                'Failed': 1,
+                'Succeeded': 1,
+                'Aborted': 1
+            }
+
+            status_counts_single = {'Running': 1}
+
+            expected_aggregations_response_dict = {
```
calbach: `assert_aggregation_response_equal` should be a generic function. So it shouldn't have a specific hard-coded response here that's tied to your test. Would prefer this were defined in the test function, then passed as a parameter to `assert_aggregation_response_equal`.

Would also be nice if that parameter were of the same form (AggregationResponse, rather than dict), but fine to keep it a dict if using the model is to verbose. If you want to leave the expected response as a dict, I'd just remove this function, convert the test response to a dict in your test method and call assert_equals() directly there.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
...
+                    userId: status_counts_total
+                },
+                'name': {
+                    name: status_counts_total
+                },
+                'aggregation-testing-unique-1': {
+                    'testing-rocks': status_counts_group,
+                    'different-value': status_counts_single
+                },
+                'test_token': {
```
calbach: I'm finding test_token to be kind of confusing here, can you just drop it from your test case? (always omit the label). Was this testing something specific?

servers/dsub/jobs/test/base_test_cases.py
calbach: Cool - this line right here is the main thing I wanted, I can now understand pretty easily what the test is looking for.

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +468,174 @@ def test_query_jobs_submission_pagination(self):
...
+            self.wait_status(job_running_diff_labels, ApiStatus.RUNNING)
+
+            userId = self.must_get_job(job_running).extensions.user_id
+            aggregation_resp = self.must_get_job_aggregations('HOURS_1')
+
+            self.assert_aggregation_response_equal(aggregation_resp, userId,
+                                                   name)
+
+            # Cancel jobs for testing
+            self.must_abort_job(job_running)
```
calbach: Not sure I would assert fail here, the main goal would be to just limit runtime, so you want this to run for all jobs and you don't necessarily need the test to fail if this doesn't work. Probably I would just call abort without the assert fail (modified version of `must_abort_job`)

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -38,6 +42,10 @@ def setUpClass(cls):
             cls.provider = None
             cls.wait_timeout = 30
             cls.poll_interval = 1
+            cls.test_token_label = {
+                'test_token':
+                datetime.datetime.now().strftime('%Y%m%d_%H%M%S_%f')
```
calbach: Per offline discussion: move to `setUp` instead

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -64,6 +72,13 @@ def create_app(self):
             app = connexion.App(__name__, specification_dir='../swagger/')
             app.app.json_encoder = JSONEncoder
             app.add_api('swagger.yaml')
+
+            # TEST_TOKEN_VALIE is a global config value used to distinguish testing jobs from batch to batch by timestemp
+            app.app.config['TEST_TOKEN_VALUE'] = self.test_token_label[
+                'test_token']
+            # AGGREGATION_JOB_NAME_FILTER is a global config value used to distinguish aggregation testing jobs from others
+            app.app.config[
+                'AGGREGATION_JOB_NAME_FILTER'] = "aggregation-testing-unique"
```
calbach: Per offline, switch to just check test token

switch naming to be `config['AGGREGATION_JOB_LABEL_FILTER'] = 'test_token=...'`

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -38,14 +38,8 @@ def get_job_aggregations(timeFrame, projectId=None):
...
         for job in jobs:
-
             # TEST_TOKEN_VALIE is a global config value used to distinguish testing jobs from batch to batch by timestemp
-            if 'TEST_TOKEN_VALUE' in current_app.config and (
-                    'test_token' not in job.labels or job.labels['test_token']
-                    != current_app.config['TEST_TOKEN_VALUE']):
-                continue
-            # AGGREGATION_JOB_NAME_FILTER is a global config value used to distinguish aggregation testing jobs from others
-            if 'AGGREGATION_JOB_NAME_FILTER' in current_app.config and job.name != current_app.config['AGGREGATION_JOB_NAME_FILTER']:
+            if 'AGGREGATION_JOB_LABEL_FILTER' in current_app.config and job.labels['test_token'] != current_app.config['AGGREGATION_JOB_LABEL_FILTER']:
```
calbach: Please don't hardcode `"test_token"` here, please see my original comment on proposed syntax, i.e. set this to `test_token=...`, then just parse that = out here. I realize there is exactly one user of this config parameter but still it is messy to depend on entering the exact same string here as in your test file.

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -38,14 +38,8 @@ def get_job_aggregations(timeFrame, projectId=None):
 
     try:
         for job in jobs:
-
             # TEST_TOKEN_VALIE is a global config value used to distinguish testing jobs from batch to batch by timestemp
```
calbach: Update comment

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -548,94 +499,94 @@ def test_get_job_aggregations(self):
...
-            # Create an aborted job
+            # Created a job to be aborted
             job_aborted = self.api_job_id(
-                self.start_job(
-                    'sleep 180',
-                    labels=labels,
-                    name=name,
-                    with_test_token=True))
-            # Create a failed job
+                self.start_job('sleep 180', labels=labels, name=name))
```
calbach: since name doesn't matter anymore, perhaps just make the names descriptive instead, e.g. `name="agg-abort"`, this may be useful for manual debugging if these tests fail

servers/dsub/jobs/test/base_test_cases.py
calbach: Not addressed?

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -548,94 +499,94 @@ def test_get_job_aggregations(self):
...
+                },
+                'test_token': {
+                    self.test_token_label['test_token']: status_counts_total,
+                },
+                'different-label': {
+                    'different-value': status_counts_single
+                }
+            }
+
+            self.assertEqual(aggregation_response_dict,
```
calbach: Nice

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -548,94 +499,94 @@ def test_get_job_aggregations(self):
...
+                             expected_aggregations_response_dict)
 
             # Cancel jobs for testing
-            self.must_abort_job(job_running)
-            self.must_abort_job(job_running_no_token)
-            self.must_abort_job(job_running_diff_labels)
-            self.must_abort_job(job_running_diff_label_value)
-            self.must_abort_job(job_running_no_token_diff_label)
+            self.client.open(
```
calbach: `for j in [job_running, job_running...]:`

Alternatively, accumulate in a list per above comment

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +457,136 @@ def test_query_jobs_submission_pagination(self):
...
+
+        def status_counts_to_dict(self, status_counts):
+            return {
+                status_count.status: status_count.count
+                for status_count in status_counts.counts
+            }
+
+        def aggregation_response_to_dict(self, aggregation_response):
+            # Flatten the nested aggregation response to a dict that is easy to compare equality
+            aggregation_response_dict = {}
```
calbach: nit: I would merge these two lines instead:

```
aggregation_response_dict = {
  'summary': self.status_counts_to_dict(aggregation_response.summary)
}
```

servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -438,3 +457,136 @@ def test_query_jobs_submission_pagination(self):
...
+                    entry_dict[entry.label] = self.status_counts_to_dict(
+                        entry.status_counts)
+
+                aggregation_response_dict[aggregation.key] = entry_dict
+
+            return aggregation_response_dict
+
+        def test_get_job_aggregations(self):
+            # Set the current test_token as global config parameter
+            # so that the aggregation_controller can filter out jobs do not need
```
calbach: grammar nit: I would rephrase as "filter out unrelated jobs"

servers/dsub/jobs/test/test_jobs_controller_local.py
```diff
@@ -33,6 +34,9 @@ def setUp(self):
         # Create logging directory
         self.log_path = '{}/logging'.format(self.testing_root)
         os.mkdir(self.log_path)
+        self.test_token_label = {
```
calbach: This is duplicated, please call `super()` here instead and put this logic into the base_test_cases class

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -36,10 +36,18 @@ def get_job_aggregations(timeFrame, projectId=None):
     job_name_summary = {}
     label_summaries = {}
 
+    # AGGREGATION_JOB_LABEL_FILTER is a global config value used to distinguish testing jobs from batch to batch by timestemp
+    has_aggregation_filter = 'AGGREGATION_JOB_LABEL_FILTER' in current_app.config
+    # aggregation_filter is in string formate 'key=value'
```
calbach: formate -> format

servers/dsub/jobs/controllers/aggregation_controller.py
```diff
@@ -36,10 +36,18 @@ def get_job_aggregations(timeFrame, projectId=None):
     job_name_summary = {}
     label_summaries = {}
 
+    # AGGREGATION_JOB_LABEL_FILTER is a global config value used to distinguish testing jobs from batch to batch by timestemp
```
calbach: timestemp -> timestamp

servers/dsub/jobs/test/test_jobs_controller_local.py
```diff
@@ -8,6 +8,7 @@
 import tempfile
 import time
 import unittest
+import datetime
```
calbach: nit: revert this import

PR #371, opened 2018-07-18
A really simple quickstart script

cjllanwarne:
Plenty of fix-up is possible and my python is, uh, more than a little bit rusty, but you should be able to try this out with the [instruction here](https://github.com/DataBiosphere/job-manager/tree/cjl_quick_start#try-it-out)


README.md
```diff
@@ -4,6 +4,16 @@
 
 _This product is in Alpha and not yet ready for production use. We welcome all feedback!_
 
+
+## Try it out!
+
+This script is not perfect but if you're on a linux-based system you should be able to get started by copying the following command into a terminal window and following the prompts:
+`
+curl -s https://raw.githubusercontent.com/DataBiosphere/job-manager/cjl_quick_start/deploy/quickstart/quick_start.py > /tmp/jmui.py && python3 /tmp/jmui.py
```
cjllanwarne: We probably want a stable, non-feature-branch URL if this is accepted and merged.

deploy/cromwell/capabilities-config.json
```diff
@@ -40,7 +40,7 @@
   ],
   "queryExtensions": [],
   "authentication": {
-    "isRequired": true,
+    "isRequired": false,
```
cjllanwarne: this wasn't deliberate... oops!

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+import urllib.request
+from subprocess import call
+import subprocess
+
+
+def quick_start():
+    version = "v0.2.0"
+
+    home = os.getenv("HOME")
+    install_dir = request_input_path("Select an installation directory", home + "/jmui")
```
rexwangcc: nit: use `'{0} is a {1}'.format(variable0, variable1)` or `'{v0} is a {v1}'.format(v0=variable0, v1=variable1)`throughout the script for better readability and performance.

README.md
rexwangcc: We can modify the URL here to use master branch before this PR gets merged in.

rexwangcc: I think it might be necessary to add a note here saying "you have to have python3 installed on your machine", otherwise  `python3 /tmp/jmui.py` will fail. Besides, `import urllib.request` will fail on Python2.x.

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+def make_or_replace_or_reuse_directory(path_to_make):
+    valid = False
+    time_to_exit = False
+    while not time_to_exit:
+        if path.isdir(path_to_make):
+            print("Directory " + path_to_make + " already exists. I can...")
+            print("1. Default: Re-use the existing directory which might leave behind some old files")
+            print("2. Delete the entire directory and all of its contents - and then remake it, completely empty")
+            provided = input("Replace or re-use? (1/2) [ 1 ] > ") or "1"
+            if provided is 1:
```
rexwangcc: Note,  this will run into a dead loop since provided is a `string` and it is checking if a `string` == `int` here. To resolve it, either use `provided = int(input(...))` or use `if provided is str(1):` or `if provided is "1"`

rexwangcc: Besides, usually we use == when comparing values and `is` when comparing identities, when comparing ints (or immutable types, like str in general), you might want the former to avoid some traps. 

There's an "optimization" which lets small ints to be compared with `is`, but that will be problematic for larger cases:
```
>>> a = 19999999998
>>> b = 19999999997 + 1
>>> a is b
False
>>> a == b
True
```
For boolean comparisions `is` is always more pythonic tho (`if not`, `if True`...).

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+def request_input_path(message, default):
+    valid = False
+    absolute_path = path.abspath(default)
+    while not valid:
+        provided = input(">> " + message + " [ " + absolute_path + " ] > ") or absolute_path
+        absolute_path = path.abspath(provided)
+        print("Using: " + provided)
+
+        make_or_replace_or_reuse_directory(provided)
+
```
rexwangcc: missing a `valid = True` here, which will cause infinite loop i think

rexwangcc: You probably want to assign the` return value` from `make_or_replace_or_reuse_directory()`(which currently doe not have any return values) to the local `valid` to True?

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+def make_or_replace_or_reuse_directory(path_to_make):
+    valid = False
+    time_to_exit = False
+    while not time_to_exit:
+        if path.isdir(path_to_make):
+            print("Directory " + path_to_make + " already exists. I can...")
+            print("1. Default: Re-use the existing directory which might leave behind some old files")
+            print("2. Delete the entire directory and all of its contents - and then remake it, completely empty")
+            provided = input("Replace or re-use? (1/2) [ 1 ] > ") or "1"
```
rexwangcc: nit: "Re-use or replace" since 1 => Re-use and 2 => Replace

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+    config_dir = install_dir + "/config"
+    make_or_replace_or_reuse_directory(config_dir)
+
+    os.chdir(install_dir)
+
+    service_type = request_input_from_list("Select a shim to use", ["Cromwell", "dsub"], "Cromwell")
+    options = {
+        "Cromwell": quick_start_cromwell
+    }
+    options[service_type](version, install_dir, bin_dir, config_dir)
```
rexwangcc: typing `dsub` here will raise a `KeyError: 'dsub'` since you are using  `options[service_type]` here, you may want to the call `quick_start_dsub()` here in a `try..except KeyError..` block

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,211 @@
...
+                    valid = False
+                    time_to_exit = True
+            else:
+                print("Invalid input: got " + provided + " but expected one of (1/2)")
+                valid = False
+                time_to_exit = False
+        else:
+            try:
+                os.mkdir(path_to_make)
+                valid = True
```
rexwangcc: I'm not sure if you want a `time_to_exit = False` here?

cjllanwarne:
Thanks @rexwangcc - I'm not a "native python speaker" so pointing out all these gotchas and bad behaviors are very much appreciated!

deploy/quickstart/quick_start.py
cjllanwarne: Oh yeah, I switched the order because I thought "re-use" was a slightly safer default if people are just accepting defaults and not reading properly. I must've forgotten to re-order these. Good spot!

deploy/quickstart/quick_start.py
cjllanwarne: So I made the `valid` result the return value of `make_or_replace_or_reuse_directory` so if an OS error occurs we can bounce it back to the caller to decide whether to try again, fail, ask for an alternative directory, etc

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,245 @@
+import os
+from os import path
+import urllib.request
+from subprocess import call
+import subprocess
+
+
+def quick_start():
+    version = "v0.2.0"
```
bfcrampton: nit: We use single quotes in python everywhere else in this codebase, could you update throughout for consistency?

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,245 @@
...
+        'Enter the raw Cromwell IP address, eg "10.1.10.100" (DO NOT USE "LOCALHOST" or "127.0.0.1"!)',
+        cromwell_service_clue)
+
+    os.chdir(install_dir)
+    start_script_file = '{0}/jmui_start.sh'.format(bin_dir)
+    start_script_contents = [
+        'if docker ps | grep job-manager', 'then',
+        'echo \'Stopping current instance(s)\'',
+        'docker stop $(docker ps | grep \'job-manager\' | awk \'{ print $1 }\')',
+        'fi', 'export CROMWELL_URL=http://{0}:8000/api/workflows/v1'.format(
```
rexwangcc: I'd let users specify the port number(maybe as part of the IP address) here as well (we can still leave 8000 as a default value)

deploy/quickstart/quick_start.py
```diff
@@ -0,0 +1,245 @@
...
+            valid = False
+    print('Using: ' + provided)
+    print()
+    return provided
+
+
+def make_or_replace_or_reuse_directory(path_to_make, exit_if_invalid):
+    valid = False
+    time_to_exit = False
+    while not time_to_exit:
```
rexwangcc: nit: It would be great to have some separators before each call of this function, since this function will be called more than once, I was confused(thinking it's an infinite loop) before I looked into the actual msg. A possible approach is to use a long line based on the console width: 
```
import os
print('=' * int(os.popen('stty size', 'r').read().split()[1]))
```

PR #372, opened 2018-07-20
[Controversial] Add .style.yapf, widen python columns

cjllanwarne:
Adds a `.style.yapf` file with some sensible seeming (to me) values, but mine are not always the most globally recognized opinions w.r.t. coding styles so I'm 100% open to adjustments.

In particular I raised `COLUMN_WIDTH` from 79 to 139 to make my own IDE experience good. Other value suggestions are welcome.


.style.yapf
```diff
@@ -0,0 +1,247 @@
...
+# would reformat to:
+#
+#    call_func_that_takes_a_dict({
+#        'key1': 'value1',
+#        'key2': 'value2',
+#    })
+coalesce_brackets=False
+
+# The column limit.
+column_limit=139
```
calbach: 139 seems... high. Would prefer 99 personally

geoffjentry:
PEP8 specifies 79 for column width, IIRC

calbach:
@geoffjentry today this repo enforces pep8. This PR is proposing to deviate from that.

I have a preference towards just adopting existing standard style guidelines (i.e. keeping what we had already), but I won't strongly object to overriding some cases like this if others feel it is causing excessive pain.

geoffjentry:
@calbach Right, I was directing that at @cjllanwarne 

In general I'm all for really long columns, but in this case there exists an official style guide for the language and it **is** opinionated on this matter, so I personally defer to it

cjllanwarne:
Yes, I'm aware of PEP8.

This PR came about because the linter-enforced PEP8 in https://github.com/DataBiosphere/job-manager/pull/371/files was breaking lines in ways that I thought were:

- Reducing readability
  - Eg the seemingly arbitrary line break on [line 73](https://github.com/DataBiosphere/job-manager/pull/371/files#diff-a1c0034ef00b9de31abc6015ae73efbcR73) vs its surrounding lines
- Increasing inconsistency
  - Eg: The inconsistent `)` placement on [lines 160/161](https://github.com/DataBiosphere/job-manager/pull/371/files#diff-a1c0034ef00b9de31abc6015ae73efbcR161) vs [lines 162/163/164](https://github.com/DataBiosphere/job-manager/pull/371/files#diff-a1c0034ef00b9de31abc6015ae73efbcR164)

That's probably a side effect of my generally trusting linters less than humans to write something readable (maybe if I spent a little more time, I could have refactored the examples above fit under 80 characters by hand)

I'll also say that I'm also not normally a python developer - so if inconsistency with PEP8 makes this less portable, I'm happy to roll this PR back and accept the above as something I just need to get used to.

calbach:
I love being able to thwart an entire class of pedantic code review comments with an autoformatter/linter. It is certainly a tradeoff as it does make code ugly sometimes, but you can generally work around those cases, e.g. the class of issues with strings you mention don't exactly get solved by increasing the line limit, you'll have longer strings that just barely need to be wrapped given any limit.

As for whether we use vanilla PEP8 style, that's my preference as not to set a precedent for overruling PEP8 whenever there's a particularly egregious style applied. Line length is unlikely to be the only thing in the style guide you will find offensive. But as I've said above, I don't feel strongly (and my personal style preference is indeed for >80). Others should add opinions.

PR #375, opened 2018-07-24
Display shard statuses for scattered tasks

rsasch:
TODO
- [x] add test case(s)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,88 @@ def get_job(id, **kwargs):
 
 
 def format_task(task_name, task_metadata):
+    # check to see if task is scattered
+    is_scattered_task = len(
+        set([item.get('shardIndex') for item in task_metadata])) > 1
```
cjllanwarne: This should probably be more like "is there a `shardIndex` that isn't -1?"

It's possible that this wouldn't work properly on a scatter that's got a width of 1

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,88 @@ def get_job(id, **kwargs):
...
+    if is_scattered_task:
+        return format_scattered_task(task_name, task_metadata)
+    else:
+        latest_attempt = task_metadata[-1]
```
cjllanwarne: ToL: is it guaranteed that the last entry is the most recent attempt?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,88 @@ def get_job(id, **kwargs):
...
+            shard_statuses=None)
+
+
+def format_scattered_task(task_name, task_metadata):
+    temp_status_collection = {}
+    current_shard = ''
+    minStart = _parse_datetime(task_metadata[0].get('start'))
+    maxEnd = _parse_datetime(task_metadata[0].get('end'))
+
+    # go through calls in reverse to grab the latest attempt if there are multiple
```
cjllanwarne: technically, to:

  - grab the lastest attempt
  - find the earliest start time
  - find the latest end time

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,88 @@ def get_job(id, **kwargs):
...
+            if status not in temp_status_collection:
+                temp_status_collection[status] = 1
+            else:
+                temp_status_collection[
+                    status] = temp_status_collection[status] + 1
+        current_shard = shard.get('shardIndex')
+        if minStart > _parse_datetime(shard.get('start')):
+            minStart = _parse_datetime(shard.get('start'))
+        if maxEnd < _parse_datetime(shard.get('end')):
+            maxEnd = _parse_datetime(shard.get('end'))
```
cjllanwarne: If one shard is done and one is still running, does it make sense to display the end-time?

servers/cromwell/jobs/models/__init__.py
```diff
@@ -19,6 +19,7 @@
 from .query_jobs_request import QueryJobsRequest
 from .query_jobs_response import QueryJobsResponse
 from .query_jobs_result import QueryJobsResult
+from .shard_status_count import ShardStatusCount
```
cjllanwarne: Is this used in this file?

ui/src/app/job-details/tasks/tasks.component.css
```diff
@@ -74,3 +76,38 @@ clr-icon {
...
+  line-height: 1.5rem;
+  margin: 0 0.5rem 0 0.5rem;
+  font-weight: bold;
+}
+
+.shards-statuses {
+  margin: 0.5rem;
+}
+
+.shard-count {
```
cjllanwarne: ToL (naive question because I am 100% *not* a UI person):

Does it make sense to be this granular with styles? Is there not a general purpose "info box number" style, or similar?

ui/src/app/job-details/tasks/tasks.component.html
```diff
@@ -19,9 +19,26 @@
...
+              <clr-tooltip-content clrPosition="right"  clrSize="xs" *clrIfOpen>
+                <span>task shards</span>
+              </clr-tooltip-content>
+            </clr-tooltip>
+          </button>
+          <mat-menu #shardMenu="matMenu" class="wide-menu" xPosition="after" yPosition="below" [overlapTrigger]="false">
+            <div class="shards-header">Scattered: {{ getScatteredCountTotal(t) }} shards</div>
+            <div class="shards-container" *ngIf="isScattered(t)">
+              <div class="shards-statuses" *ngFor="let jobStatus of getShardStatuses()">
+                <div class="shard-status"><clr-icon [attr.shape]="getStatusIcon(jobStatus)" size="18"></clr-icon></div>
```
cjllanwarne: A few more probably-naive style questions:
  - Should `size` be hard-coded here?
  - How many digits Is '18' wide enough to hold ?

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: fixed

gemmalam:
@rexwangcc @samanehsan could either of you give this PR a look?

cjllanwarne:
@rsasch I've deleted my "just for info" questions - a few comments remain that I think are worth thinking about

cjllanwarne:
Is the "add test" TODO still TODO? TODOne? "One for later"?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,90 @@ def get_job(id, **kwargs):
...
+            if status not in temp_status_collection:
+                temp_status_collection[status] = 1
+            else:
+                temp_status_collection[
+                    status] = temp_status_collection[status] + 1
+        if minStart > _parse_datetime(shard.get('start')):
+            minStart = _parse_datetime(shard.get('start'))
+        if shard.get('executionStatus') not in ['Failed', 'Done']:
+            maxEnd = None
+        if maxEnd is not None and maxEnd < _parse_datetime(shard.get('end')):
```
rexwangcc: What happens if `_parse_datetime(shard.get('end'))` is `None`? Is it possible?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,90 @@ def get_job(id, **kwargs):
...
+            maxEnd = _parse_datetime(shard.get('end'))
+        current_shard = shard.get('shardIndex')
+
+    shard_status_counts = [
+        ShardStatusCount(status=status, count=temp_status_collection[status])
+        for status in temp_status_collection
```
rexwangcc: nit: `ShardStatusCount(status=status, count=count) for status, count in temp_status_collection.items()` might be more "pythonic" here?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,90 @@ def get_job(id, **kwargs):
...
-    """ Remove the workflow name from the beginning of task, input and output names.
+    """ Remove the workflow name from the beginning of task, input and output names (if it's there).
     E.g. Task names {workflowName}.{taskName} => taskName
          Input names {workflowName}.{inputName} => inputName
          Output names {workflowName}.{taskName}.{outputName} => taskName.outputName
     """
-    return '.'.join(name.split('.')[1:])
+    if "." in name:
+        return '.'.join(name.split('.')[1:])
+    return name
```
rexwangcc: An optional nit that might save invisible computation time here by using [`str.partition()`](https://docs.python.org/3/library/stdtypes.html#str.partition): 
```
partitioned = name.partition('.')
name = partitioned[2] if partitioned[2] != '' else partitioned[0]
return name
```

ui/src/app/shared/pipes/duration.pipe.ts
```diff
@@ -6,13 +6,15 @@ import {Pipe, PipeTransform} from '@angular/core';
...
   transform(start: Date, end: Date): string {
-    let duration: number;
-    if (end) {
-      duration = end.getTime() - start.getTime();
-    } else {
-      duration = new Date().getTime() - start.getTime();
-    }
-    return Math.floor(duration/3600000) + "h " +
-      Math.floor(duration/60000)%60 + "m";
+   if (start) {
```
rexwangcc: If I understand it correctly, there are indeed some situations that `start` is missing?

rsasch: There were, but I believe I've addressed them, so I've removed this check – thanks for the catch.

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: ok, done

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: ok, done

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: `_parse_datetime(None)` will return `None` (see line 334)

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -128,29 +129,90 @@ def get_job(id, **kwargs):
 
 
 def format_task(task_name, task_metadata):
+    # check to see if task is scattered
+    if task_metadata[-1].get('shardIndex') != -1:
```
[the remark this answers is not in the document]
rsasch: if the task isn't scattered, `shardIndex` will be set to `-1`

PR #377, opened 2018-07-25
Stop tracking generated swagger files

cjllanwarne:
Since they are now easy to generate thanks to #374, this PR removes all of the generated build artifacts made by `swagger-codegen`.


cjllanwarne:
@calbach good point! I can imagine remembering to run `rebuild` when switching between branches is particularly likely to annoy us...

Question - is `docker-compose` intended to be a de-facto build script? 

  - If so should we call it `build-docker-compose` to distinguish from the "deployment" versions of `docker-compose`?
  - If not would a one-line "build and deploy" (as a replacement for running `docker-compose up` when doing dev work) make more sense?

calbach:
> Question - is docker-compose intended to be a de-facto build script?

Yes.

> If so should we call it build-docker-compose to distinguish from the "deployment" versions of docker-compose?

Perhaps you added something new in the quick start PR, but the intent for docker-compose has always been for local development only. Deployment are done via Kubernetes or app engine (and for k8s, not via swarm or anything that would integrate with docker-compose).

> If not would a one-line "build and deploy" (as a replacement for running docker-compose up when doing dev work) make more sense?

I do like the simplicity of just using docker-compose, but if we'd need to bend over backwards to integrate this new step, then wrapping in a development script is probably reasonable.

cjllanwarne:
The quick-start does indeed use `docker-compose` but it's based on a different compose file in `deploy/` so it shouldn't need to interact with `common-docker-compose.yml`.

I'll have a look at introducing that build step into the compose sequence.

servers/dsub/Dockerfile
```diff
@@ -16,4 +16,4 @@ RUN rm -rf jm_utils
 
 # Missing required arguments -b PORT, -e ... which must be provided by the
 # docker image user.
-ENTRYPOINT ["/env/bin/gunicorn", "jobs.__main__:app"]
+ENTRYPOINT ["/bin/bash", "/scripts/await_md5_match.sh", "/app/jobs/models/.jobs.yaml.md5", "--", "/env/bin/gunicorn", "jobs.__main__:app"]
```
cjllanwarne: This is awkward in the `Dockerfile`. I think I ought to remove `ENTRYPOINT` altogether and set the entire command in the `docker-compose.yml`

cjllanwarne:
@calbach I think this is what you meant:

- The `rebuild_swagger.sh` script now records the md5 sum of the `jobs.yaml` it used, alongside the models it generates.
- It only rebuilds the fields if that `.md5` file doesn't validate against `api/jobs.yaml`
- There's a new service which mounts the job-manager directory to gain access to `jobs.yaml`, `rebuild_swagger.sh` and the destinations for the model generation. It runs `rebuild_swagger.sh` and then exists.
- The other services are also given the `jobs.yaml` file and won't begin their real work until an `.md5` file exists where they expect their model files to be mounted, *and* that md5 file validates against `api/jobs.yaml`

It feels like there's a lot of moving parts here and a lot of "things happening to be in the right places", but maybe that can be consolidated with some shuffling of what goes in `Dockerfile`s vs `compose.yml`s... what do you think?

common-compose.yml
```diff
@@ -1,5 +1,10 @@
 version: '2.1'
 services:
+  rebuild-swagger:
+    image: java
+    volumes:
+      - ./:/base
+    entrypoint: ["/bin/bash", "-c", "cd base && /base/scripts/rebuild_swagger.sh"]
```
calbach: nit: Does `working_dir: "/base"` work?

scripts/rebuild_swagger.sh
```diff
@@ -18,28 +18,59 @@ then
...
-  -o servers/dsub \
-  -DsupportPython2=true,packageName=jobs
-
-java -jar swagger-codegen-cli.jar generate \
-  -i api/jobs.yaml \
-  -l python-flask \
-  -o servers/cromwell \
-  -DsupportPython2=true,packageName=jobs
+MD5=$(md5sum api/jobs.yaml)
+#MD5=$(md5sum api/jobs.yaml | awk ' { print $1 }' )
```
calbach: nit: rm commented out lines

servers/dsub/Dockerfile
calbach: Agreed - for dsub we do currently use this Dockerfile for deployment, which sort of depends on the entrypoint being setup like this (not sure about the Cromwell shim). I think for this specific Dockerfile, you can just leave it as is (perhaps adding a call to run the swagger generation a few lines up).

For the other local dev Dockerfiles, perhaps just remove the entrypoints and move the logic into docker-compose entrypoint or command?

calbach:
This is very cool - thanks. I didn't consider the MD5, but I see this may be necessary since depends_on doesn't actually wait for the command to run.

servers/dsub/Dockerfile
calbach: @rexwangcc should probably verify the changes to the Cromwell shim Dockerfile; I'm not sure what image k8s is actually using.

rexwangcc: Mint's k8s cluster is using the dockerfile `servers/cromwell/Dockerfile`. When removing the `ENTRYPOINT`, please make sure add docs about it, otherwise the person who deploys the docker image has no idea what command to add to either docker-compose file or kubernetes yaml file.

cjllanwarne:
@calbach I've unified on having the `await_md5_match.sh` in the Dockerfile to prevent startup until the swagger files are generated. It's up to whoever uses the Dockerfile to provide those swagger files somehow - either with a manual build or by including the `rebuild_swagger` service in their `docker_compose.yml`.

@calbach @rexwangcc because of this, the *interface* to the images is unchanged except that it will delay your command from starting until those swagger files are generated

I think I'm pretty happy with this now - let me know what you think

PR #378, opened 2018-07-25
Allow backend to tell UI the total number of results it has

cjllanwarne:
Cromwell always knows the total number of results that can be listed, so we should always be able to display it.

This PR (hopefully!) continues the existing `dsub` behavior of displaying `many` until the jobs list knows that its results set is exhaustive


api/jobs.yaml
```diff
@@ -407,10 +407,18 @@ definitions:
     properties:
       results:
         type: array
+        required: true
         items:
           $ref: '#/definitions/QueryJobsResult'
+      totalKnownResults:
+        type: number
+        required: false
```
calbach: opt_nit: This should be the default so I'd probably leave it unset.

api/jobs.yaml
```diff
@@ -407,10 +407,18 @@ definitions:
     properties:
       results:
         type: array
+        required: true
         items:
           $ref: '#/definitions/QueryJobsResult'
+      totalKnownResults:
```
calbach: I'd probably prefer `totalSize`, or if not that then at least `totalResults`. This API mostly follows the Google Cloud API style guide for pagination, which recommends `totalSize`:

https://cloud.google.com/apis/design/design_patterns#list_pagination

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -194,11 +194,15 @@ def query_jobs(body, **kwargs):
 
     now = datetime.utcnow()
     jobs_list = [format_job(job, now) for job in response.json()['results']]
+    total_known_results = int(response.json()['totalResultsCount'])
```
calbach: There's already a local variable with this value, `total_results`

ui/src/app/job-list/job-list.component.ts
```diff
@@ -32,6 +32,7 @@ export class JobListComponent implements OnInit {
   // whenever we change query parameters, via a subscription.
   jobs = new BehaviorSubject<JobListView>({
     results: [],
+    totalKnownResults: 0,
```
calbach: nit: If it's valid for this to be null/undefined, then I'd probably start with that (just change the type in `JobListView` to `number|null` or `number|undefined`).

ui/src/app/shared/header/header.component.ts
```diff
@@ -279,16 +279,31 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
...
-    if (this.backendJobs.value.exhaustive) {
+
+    let knownLength = null;
+
+    if (this.backendJobs.value.totalKnownResults != null) {
+      knownLength = this.backendJobs.value.totalKnownResults;
+    } else if (this.backendJobs.value.exhaustive) {
+      knownLength = length;
+    } else {
+      knownLength = null;
```
calbach: redundant

ui/src/app/shared/header/header.component.ts
```diff
@@ -279,16 +279,31 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
   }
 
   getRangeLabel = (page: number, pageSize: number, length: number) => {
-    if (this.backendJobs.value.exhaustive) {
+
+    let knownLength = null;
+
+    if (this.backendJobs.value.totalKnownResults != null) {
```
calbach: This line scares me a bit. See this page: http://zero.milosz.ca/

I think perhaps you should more clearly define the expected type union of `JobListView.totalKnownResults`, and then do a `!==` accordingly.

ui/src/app/shared/header/header.component.ts
```diff
@@ -279,16 +279,31 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
...
+      knownLength = null;
+    }
+
+    if (knownLength != null) {
       // Can't use proper inheritance here, since MatPaginatorIntl only defines
       // properties, rather than class methods.
-      return this.defaultIntl.getRangeLabel(page, pageSize, length);
+
+
+      console.log(page, pageSize, knownLength)
```
calbach: rm logger and newlines

ui/src/app/shared/header/header.component.ts
```diff
@@ -279,16 +279,31 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
...
+      return this.defaultIntl.getRangeLabel(page, pageSize, knownLength);
+    } else {
```
calbach: nit: can just remove the "else" part and reduce indentation, since you're returning above.

ui/src/app/shared/job-stream.ts
```diff
@@ -38,11 +39,14 @@ export class JobStream extends BehaviorSubject<JobListView> {
       let pageSize = Math.max(
         JobStream.minBackendPageSize, atLeast - this.value.results.length);
       return this.queryJobs(pageSize, prevResp.nextPageToken).then(resp => {
+        console.log(resp);
```
calbach: rm

ui/src/app/shared/job-stream.ts
```diff
@@ -38,11 +39,14 @@ export class JobStream extends BehaviorSubject<JobListView> {
...
         this.next({
           results: this.value.results.concat(resp.results),
+          totalKnownResults: resp.totalKnownResults,
           exhaustive: !resp.nextPageToken,
           stale: false
         });
+        console.log(this.value);
```
calbach: rm

ui/src/app/shared/header/header.component.ts
calbach: This is fine to leave as-is, I should have flagged this as an optional comment. Definitely personal preference - for me, it adds more visual noise (indentation) and cognitive load for following the "normal path" through the code. This is mostly coming from a background in Go where this kind of style is preferred. In this case, that argument's not as compelling as there's arguably no "normal path". But possibly slightly better with inversion:

```
if (knownLength == null) {
  ...
}
return this..defaultIntl...
```

PR #444, opened 2018-07-30
google provider test fixing

zacharyhsia:
Seems that Google provider test is excluded on last merge, fixing the bug.

### DIFF
--- a/servers/dsub/jobs/test/test_jobs_controller_google.py
+++ b/servers/dsub/jobs/test/test_jobs_controller_google.py
@@ -4,6 +4,7 @@ from dsub.providers import google
 import flask
 import operator
 import unittest
+import datetime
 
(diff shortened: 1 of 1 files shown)


cjllanwarne:
👍 on the change (cf #445...)

Any insight as to why this wasn't failing the PR versions of the tests? Do they not get run in PRs (do we risk regressions?)

zacharyhsia:
@calbach Not sure why the other test (Google provider) is not run on the last PR. Was the CircleCI configuration modified?

calbach:
Replied on slack as well, but reposting: "we changed this to be post-submit only since these were less frequently modified and take a long time to run".

The exclusion happens here FTR: https://github.com/DataBiosphere/job-manager/blob/master/.circleci/config.yml#L73

PR #446, opened 2018-07-31
Untrack even more swagger codegen

cjllanwarne:
Missed this in the first round (#377)


cjllanwarne:
@calbach the `.gitignore` stops new files from being automatically added, but doesn't automatically delete any existing files (at least not on my `git` version). For my PR I had to manually delete the files inside these directories (and because git works with files not directories, my delete was really "a list of individual files inside this directory", not just "this directory").

What I think happened here is that a new PR was merged *after* my `git rm` which added a few new files that weren't included in the "to delete" list

PR #449, opened 2018-08-01
Usage documentation for the rebuild-swagger service

cjllanwarne:
I've removed the "manual process" section since that wouldn't generate the `.md5` files which the services are keying off for "should I start yet".

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,17 +147,20 @@ https://github.com/angular/angular-cli/issues/6349).
 
 ### Updating the API using swagger-codegen
 
-We use [swagger-codegen](https://github.com/swagger-api/swagger-codegen) to automatically generate the API classes
-defined in `api/jobs.yaml` for the servers and for the UI. 
-Whenever the API is updated, re-run `docker-compose up rebuild-swagger` to trigger a rebuild.
+* We use [swagger-codegen](https://github.com/swagger-api/swagger-codegen) to transform the API 
+defined in `api/jobs.yaml` into appropriate classes for the servers and the UI to use.
+* Whenever the API is updated, run this to trigger a rebuild: 
+```sh
+docker-compose up rebuild-swagger
+```
  
-**Note:** the `rebuild-swagger` job does nothing if the file `api/jobs.yaml` has not changed since the
+**Notes** 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,61 +147,18 @@ https://github.com/angular/angular-cli/issues/6349).
...
- - Use `swagger-codegen-cli.jar` to regenerate them. 
-
-If you prefer to do this manually you can fallow the instructions below.
-
-#### Manual Process
+We use [swagger-codegen](https://github.com/swagger-api/swagger-codegen) to automatically generate the API classes
+defined in `api/jobs.yaml` for the servers and for the UI. 
+Whenever the API is updated, re-run `docker-compose up rebuild-swagger` to trigger a rebuild.
+ 
+**Note:** the `swagger-rebuild` job does nothing if the file `api/jobs.yaml` has not changed since the
```
rsasch: isn't it `rebuild-swagger`  instead of `swagger-rebuild` ?

rsasch: Also – if `docker-compose up` will automatically run `rebuild-swagger` if needed, do we need this note?

cjllanwarne: This note is still useful in case you want to rebuild the swagger model without also bringing up the service.

Say you just made a change to the model in a way that you know will break the UI and the services. You'll want to rebuild everything from `jobs.yaml` but there's no point bringing up the other services that you know will now have compilation errors.

README.md
```diff
@@ -147,61 +147,18 @@ https://github.com/angular/angular-cli/issues/6349).
...
-      -o servers/dsub \
-      -DsupportPython2=true,packageName=jobs
-      java -jar swagger-codegen-cli.jar generate \
-      -i api/jobs.yaml \
-      -l python-flask \
-      -o servers/cromwell \
-      -DsupportPython2=true,packageName=jobs
-      ```
+**Note:** after regenerating the model files, you'll need to test and update the server implementations to 
+resolve any broken dependencies on old API definitions or implement additional functionality to match the new specs.
```
rsasch: nit: I'd change these **Note:** 's to one "Notes" header and a list (see https://github.com/DataBiosphere/job-manager/blob/master/README.md#notes)

PR #450, opened 2018-08-01
Re-add link to timing diagram

### LINKED ISSUE #318: Display Cromwell timing diagram
- Add a panel to the details page to display timing diagrams for jobs
- In terms of least-initial effort, this could probably just be a frame containing the default cromwell timing diagram (eg `http://127.0.0.1:8000/api/workflows/v1/<<Workflow ID>>/timing`) 
- Eventually we probably want to migrate the diagram out of Cromwell altogether and let JMUI be its new home, so any progress now towards that goal will be fantastic.

cjllanwarne:
~~Shows up in the status pane. Closes #318~~ 

screen shot 2018-08-01 at 5 03 08 pm

---

EDIT: Ignore that picture! I moved this back to the "Tasks" section where it was before. That doesn't feel like a great final location for it because ultimately we should be able to draw a timing diagram for `dsub` jobs too, and they don't use the tasks panel, but for now it's good enough

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -85,6 +85,8 @@ def get_job(id, **kwargs):
     """
     url = '{cromwell_url}/{id}/metadata'.format(
         cromwell_url=_get_base_url(), id=id)
+    timing_url = '{cromwell_url}/{id}/timing'.format(
+        cromwell_url=_get_base_url(), id=id)
     response = requests.get(
         url, auth=kwargs.get('auth'), headers=kwargs.get('auth_headers'))
     job = response.json()
@@ -125,7 +127,7 @@ def get_job(id, **kwargs):
         outputs=update_key_names(job.get('outputs', {})),
         labels=job.get('labels'),
         failures=failures,
-        extensions=ExtendedFields(tasks=sorted_tasks))
+        extensions=ExtendedFields(tasks=sorted_tasks, timing_url=timing_url))
 
 
 def format_task(task_name, task_metadata):
--- a/servers/cromwell/jobs/test/test_jobs_controller.py
+++ b/servers/cromwell/jobs/test/test_jobs_controller.py
@@ -422,7 +422,8 @@ class TestJobsController(BaseTestCase):
                     'returnCode': return_code,
                     'attempts': attempts,
                     'jobId': subworkflow_id
-                }]
+                }],
+                'timingUrl': "https://test-cromwell.org/id/timing"
             }
         }  # yapf: disable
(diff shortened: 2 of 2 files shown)


cjllanwarne:
Note to reviewers - this seemed a little bit **too** easy to me so I'm suspicious that it was deliberately removed rather than it just "stopped working one day"?

ui/src/app/job-details/panels/panels.component.html
```diff
@@ -28,6 +28,9 @@
       Ended: {{ job.end | jmShortDateTime }}
       ({{ job.start | jmDuration: job.end }})
     </p>
+    <p *ngIf="job.extensions && job.extensions.timingUrl">Graph View:<a href="{{ job.extensions.timingUrl }}" target="_blank">
+      <div class="title-link">Timing Diagram</div>
```
rsasch: I wouldn't put this on two lines because it can get a little scrunched for room:
<p>
screen shot 2018-08-01 at 5 46 13 pm
</p>

cjllanwarne: 1. Fun fact - I actually intended it to be on a single line but wasn't sure what incantation of HTML tags would achieve that
2. I see you have extra `Started` and `Ended` times which I'm missing. I wonder whether I should look into why those don't exist for me (since on *my* Status panel, there's plenty of room!)

samanehsan:
> Note to reviewers - this seemed a little bit too easy to me so I'm suspicious that it was deliberately removed rather than it just "stopped working one day"?

If I remember correctly, when we added the capabilities endpoint to Cromwell it was hard-coded to always return the same information at first, but the timing diagram link is different depending on which Cromwell instance you're using!

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -125,7 +127,7 @@ def get_job(id, **kwargs):
         outputs=update_key_names(job.get('outputs', {})),
         labels=job.get('labels'),
         failures=failures,
-        extensions=ExtendedFields(tasks=sorted_tasks))
+        extensions=ExtendedFields(tasks=sorted_tasks, timing_url=timing_url))
```
samanehsan: Where does the timing_url get displayed now?

cjllanwarne: Same as before - it becomes a second link next to "List View" on the Tasks panel

samanehsan: Oh, I see now! I didn't realize that link was still there

cjllanwarne: Yeah, that's what I mean about seeming "too easy"... everything was already there except the URL!

PR #452, opened 2018-08-06
Add error panel and tab to task details page

rsasch:
- [x] Display first few error messages (including task name and links to logs & execution dir) on the task details page
- [x] Add a tab to the "Tasks" tab group that displays all the errors' info and associated links
- [x] Add a way to easily collapse the first few error messages and switch to the "Errors" tab with one click

errors-mockup-edit


api/jobs.yaml
```diff
@@ -573,13 +573,25 @@ definitions:
...
         type: string
         format: date-time
         description: The time at which this failure occurred
+      stdout:
+        type: string
+        description: Path to the standard output file for this task or shard
+      stderr:
+        type: string
+        description: Path to the standard error file for this task or shard
+      callRoot:
```
calbach: This language might be obvious for Cromwell users, but in terms of the general jobs API I have no idea what a "callRoot" is and what I should expect to find here. Is there any way we can encode this information more generally e.g. by renaming this?

api/jobs.yaml
```diff
@@ -573,13 +573,25 @@ definitions:
     required:
       - failure
     properties:
+      taskName:
+        type: string
+        description: The name of the task that generated the failure message
```
calbach: This should specify "if any" - This object is referenced both on the task and on the job, and dsub doesn't have tasks.

ui/src/app/core/local-storage.service.ts
```diff
@@ -0,0 +1,5 @@
+import {Injectable} from "@angular/core";
+
+@Injectable()
+export class LocalStorageService {
```
calbach: Did you mean to include this?

rsasch:
@calbach The description was in the linked issue; I've moved it to the PR.

api/jobs.yaml
rsasch: "callRoot" was introduced to `jobs.yaml` in #271 (see https://github.com/DataBiosphere/job-manager/blob/master/api/jobs.yaml#L558)

api/jobs.yaml
rsasch: done

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -100,7 +100,9 @@ def get_job(id, **kwargs):
     failures = None
     if job.get('failures'):
         failures = [
-            FailureMessage(failure=f['message']) for f in job['failures']
+            format_failure(name, m)
+            for name, metadata in job.get('calls', {}).items()
+            for m in metadata if m.get('failures') is not None
```
samanehsan: Wow! I've actually never seen a nested list-comprehension in python before 😮

samanehsan:
As discussed in person, the subworkflow detail pages are not showing the error panel & tab but there is error information in the metadata about failed tasks:

screen shot 2018-08-07 at 3 35 05 pm

ui/src/app/job-details/failures/failures-tab/failures-tab.component.html
```diff
@@ -0,0 +1,49 @@
...
+    <mat-cell *matCellDef="let f" class="failure-links">
+      <clr-tooltip *ngIf="getResourceUrl(f.stdout)">
+        <a class="log-item" href="{{ getResourceUrl(f.stdout) }}">
+          <clr-icon clrTooltipTrigger shape="file" size="24"></clr-icon>
+          <clr-tooltip-content clrPosition="top-left" clrSize="xs" *clrIfOpen>
+            <span>stdout log</span>
+          </clr-tooltip-content>
+        </a>
+      </clr-tooltip>
+      <clr-icon *ngIf="!getResourceUrl(f.stdout)" shape="file" size="24" style="color: #ccc;"></clr-icon>
```
bfcrampton: Per my other comment about logs on `FailureMessage` this should extract the correct log file out of the `job` or `task` (if a taskName is present).

rsasch:
@samanehsan I believe I've resolved the subworkflow issue; please pull the latest code and let me know.

samanehsan:
Thanks @rsasch! I see the error messages for the task that failed in the subworkflow now 🎉 

screen shot 2018-08-07 at 5 44 25 pm



And the errors tab: 
screen shot 2018-08-07 at 5 45 30 pm

rsasch:
@bfcrampton Thanks for the feedback; I'm going to look into whether or not it makes sense to pass `stdout`, `stderr` and `callRoot` per failure and on making the UI more flexible when it comes to displayed columns. I will ping you for a re-review once I do.

rsasch:
@bfcrampton I've moved the (renamed) `failure-table` component to a `common` directory under `job-details`, added some additional inputs and now call it in both the failures panel and the failures task tab. I've also added logic in the UI which should hide the "task name" and "links" columns if there is no data in them. And I discussed the passing of the `stdout`, `stderr` and `callRoot` per failure with @ruchim; we both think that it makes sense for the JM to rely on the Cromwell backend to supply these paths, as the location scheme might change in the future and this keeps that logic in only one place. I look forward to your feedback.

rsasch:
@bfcrampton Thanks for the feedback; the list of errors for a task will include all shards and attempts, if applicable.

PR #458, opened 2018-08-09
Initial implementation of local storage for UI options

rsasch:
- Added new `settings` service that uses local storage to save browser-based settings for UI
- Added a "gear" icon that allows the user to hide/show available columns, first acquired from capabilities endpoint.

screen shot 2018-08-09 at 1 18 03 pm


ui/src/app/job-list/job-list.component.ts
```diff
@@ -151,6 +163,15 @@ export class JobListComponent implements OnInit {
...
       .catch((error) => this.handleError(error));
   }
+
+  toggleDisplayColumn(field: DisplayField) {
+    const newValue = !field.showInListView;
+    this.displayFields = this.settingsService.setDisplayColumn(field, newValue);
+  }
+
+  saveSettings() {
+    this.jobTable.ngOnInit();
```
bfcrampton: I don't think you should call this directly, can you make `setUpFieldsAndColumns` public and just call that?

api/jobs.yaml
```diff
@@ -200,6 +200,9 @@ definitions:
         type: array
         items:
           type: string
+      showInListView:
```
bfcrampton: nit: I would change this to `defaultShow` or something similar, the 'list view' part is coupling this pretty tightly to the UI (we may use this field on the dashboard page, for example). Also mind adding a description?

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (this.currentSettings === null) {
+      const capabilities = this.capabilitiesService.getCapabilitiesSynchronous();
+      let user: string;
+      if (this.authService.isAuthenticated()) {
+        user = this.authService.userId;
+      }
+      this.currentSettings = {
+        'user' : user,
+        'project' : null,
+        'v1': { 'displayColumns': capabilities.displayFields } };
```
calbach: One thing to consider here is whether we actually want to store the default value. I think the answer is probably "no". If I've never modified my settings, and the server changes the capabilities displayFields, I should get whatever the latest default view is. Else we're somewhat arbitrarily binding the user to a view of the columns we happened to be displaying when they first logged in.

api/jobs.yaml
rsasch: giving it a name with 'default' in it doesn't work, because it can get updated and the UI checks it to see if it should display as a column at any given time

ui/src/app/core/settings.service.ts
rsasch: How about, if `'displayColumns'` isn't set in their settings, going through each `displayField` from `capabilities` and setting it be be viewable as a column?

calbach: Would need to see the code but that sounds like it would probably be fine. Ultimately I think as long as the following holds I'm good: store an array of `displayColumns` in LocalStorage IFF the user manually modified their display columns

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,58 @@
+import {Injectable} from '@angular/core';
+import {AuthService} from '../core/auth.service';
+import {CapabilitiesService} from "./capabilities.service";
+import {DisplayField} from "../shared/model/DisplayField";
+
+type Settings = {
+  // Unset for basic auth users
+  user: string
+  // Unset for non-project-based job manager
+  project: string
```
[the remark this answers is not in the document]
rsasch: Should it scope by project first and then by user or vice versa?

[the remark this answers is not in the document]
bfcrampton: I would opt to only store settings for the most recent user, i.e. if we see a new user just reset to the default settings, rather than trying to store the settings for all users the have logged into a machine. 

If that's a use case we expect (maybe users often have different accounts they sign in with), then I would probably scope by user first though I guess it doesn't matter too much.

api/jobs.yaml
bfcrampton: Ah I see what you're saying, I was just viewing this as the initial capabilities response. 

Thinking about this more, I'm don't think we should store full `DisplayField` objects in local storage. If the capabilities endpoint updates one of the other properties (say `editable`), we don't want to be relying on a cached version of the `DisplayField`. It also limits our ability to update this model in the future, every update to this API model would require a version update and migration for the user settings. 

Instead, just store a list of 'field' strings which we selected and use that to filter the full set of `DisplayField`s. When the list is blank we default to using the display fields with this field is `true`, otherwise we ignore this field.

ui/src/app/core/settings.service.ts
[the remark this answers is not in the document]
samanehsan: In terms of scoping by user vs. project first, I think either one will work fine! Although project-first may make more sense for non-project-based backends, otherwise each user will have a project key that isn't really necessary

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  public getDisplayColumns(projectId = ''): DisplayField[] {
+    if (this.getSettingsForProject(projectId) && this.getSettingsForProject(projectId).displayColumns) {
+      return this.getSettingsForProject(projectId).displayColumns;
+    }
+    const cleanDisplayColumns = this.getDefaultDisplayFieldSettings();
+    this.currentSettings.v1.projects =[{
+      'projectId': projectId,
+      'displayColumns': cleanDisplayColumns
+    } as ProjectSettings];
+    this.updateLocalStorage();
```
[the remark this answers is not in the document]
rsasch: The reason is that if there are no saved settings, the UI needs something to go by.  Rather than generate defaults each time (if the user makes no changes), they are stored for future use.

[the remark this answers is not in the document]
calbach: I think it's desirable to generate defaults each time. We shouldn't pin the user to the default columns that happened to be configured when they first logged in (unless they explicitly save that configuration).

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,83 @@
...
+          'projects': [{
+            'projectId': '',
+            'displayColumns': []
+          }]} };
+      this.updateLocalStorage();
+    }
+  }
+
+  /** return the display columns settings for a project, or if they don't exist, set all
+   * possible columns from capabilities service to be displayed and return that */
```
bfcrampton: this comment seems out of date, if the settings don't exist you're just returning `undefined` here (which the caller interprets as display all the columns with `primary = true`). Also, tiny nit, but prefer to start comments with a capital letter.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -75,6 +81,18 @@ export class JobListComponent implements OnInit {
...
+
+    // set project ID (if any) and get display field info for list columns
+    const req = URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q']);
+    this.projectId = req.extensions.projectId || '';
+    const savedColumnSettings = this.settingsService.getDisplayColumns(this.projectId);
+    let field:DisplayField;
+    this.capabilities.displayFields.forEach((df) => {
+      field = df;
+      field.primary = true;
+      field.primary = !(savedColumnSettings.length && !savedColumnSettings.includes(df.field));
```
bfcrampton: The double negation here is a little confusing, can you do:
```
field.primary = !savedColumnSettings.length || savedColumnSettings.includes(df.field)
```

api/jobs.yaml
```diff
@@ -200,6 +200,9 @@ definitions:
         type: array
         items:
           type: string
+      primary:
```
samanehsan: A `description` that this field controls whether or not the field should be displayed in the column view would be useful here!

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,83 @@
...
+  /** If there are browser settings for the current user (or for '' if none) get them;
+   * otherwise, empty out the browser settings and set up a clean scaffolding */
+  constructor(
+    private readonly authService: AuthService,
+    private readonly capabilitiesService: CapabilitiesService,
+    private readonly localStorage: Storage
+  ) {
+    const savedSettings = JSON.parse(this.localStorage.getItem('settings'));
+    if ((authService.userId && savedSettings && (savedSettings.v1.userId == this.authService.userId))
+      || !authService.userId && savedSettings && savedSettings.v1.userId === '') {
```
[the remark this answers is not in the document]
rsasch: `authService.userId && authService.userId == savedSettings.v1.userId` is the case where the current user is authenticated and the saved settings are for that user; `!authService.userId && savedSettings.v1.userId == ''` is for when there is no authentication (`userId` is `''`) and there are saved settings for that "user".

ui/src/app/job-list/job-list.component.ts
rsasch: good idea!

ui/src/app/core/settings.service.ts
rsasch: fixed

ui/src/app/core/settings.service.ts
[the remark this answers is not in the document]
rsasch: I think I've cleaned this up so it's a little more readable; please let me know what you think.

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,83 @@
...
+      if (p.projectId === projectId) {
+        p.displayColumns = fields;
+      }
+    });
+    this.updateLocalStorage();
+  }
+
+  private getSettingsForProject(projectId: string): ProjectSettings {
+    let settings: ProjectSettings = null;
+    this.currentSettings.v1.projects.forEach((p) => {
```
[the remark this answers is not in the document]
rsasch: thanks!

ui/src/app/core/settings.service.ts
```diff
@@ -0,0 +1,83 @@
...
+            'displayColumns': []
+          }]} };
+      this.updateLocalStorage();
+    }
+  }
+
+  /** return the display columns settings for a project, or if they don't exist, set all
+   * possible columns from capabilities service to be displayed and return that */
+  getDisplayColumns(projectId: string): string[] {
+    if (this.getSettingsForProject(projectId) && this.getSettingsForProject(projectId).displayColumns) {
```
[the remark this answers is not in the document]
rsasch: good idea!

servers/cromwell/jobs/swagger/swagger.yaml
```diff
@@ -217,6 +217,12 @@ definitions:
         description: "A list of values the field can be set to"
         items:
           type: "string"
+      primary:
```
calbach: Sidenote for a follow-up: this file is autogenerated should also be untracked.

ui/src/app/testing/fake-auth.service.ts
```diff
@@ -0,0 +1,11 @@
+import {AuthService} from "../core/auth.service";
+import {CapabilitiesService} from "../core/capabilities.service";
+
+export class FakeAuthService extends AuthService {
```
calbach: Unless this causes compilation issues otherwise, I probably wouldn't extend AuthService as it's probably making a bunch of remote calls.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -75,6 +81,18 @@ export class JobListComponent implements OnInit {
...
     });
+
+    // set project ID (if any) and get display field info for list columns
+    const req = URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q']);
+    this.projectId = req.extensions.projectId || '';
+    const savedColumnSettings = this.settingsService.getDisplayColumns(this.projectId);
+    let field: DisplayField;
+    this.capabilities.displayFields.forEach((df) => {
+      field = df;
+      field.primary = true;
```
calbach: mutating the original capabilities DisplayField object does seem like it may eventually lead to bugs, e.g. if we ever wanted to have a "reset to defaults" option on the column settings. I would either fix or at least note this in a comment for now though.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -75,6 +81,18 @@ export class JobListComponent implements OnInit {
...
       }
     });
+
+    // set project ID (if any) and get display field info for list columns
+    const req = URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q']);
+    this.projectId = req.extensions.projectId || '';
+    const savedColumnSettings = this.settingsService.getDisplayColumns(this.projectId);
+    let field: DisplayField;
+    this.capabilities.displayFields.forEach((df) => {
+      field = df;
```
calbach: I don't think there's any reason to do this assignment. You're still pointing to the same value (by reference) as df, so you may as well use df.

ui/src/app/job-list/job-list.component.html
```diff
@@ -5,11 +5,37 @@
...
+  </mat-menu>
+  </span>
   <jm-job-list-table
     class="job-list-table"
     [ngStyle]="{'opacity': tableOpacity}"
     [dataSource]="dataSource"
-    (onJobsChanged)="handleJobsChanged()">
+    (onJobsChanged)="handleJobsChanged()"
+    (onColumnsChanged)="handleColumnsChanged()"
+    [(displayFields)]="displayFields">
```
calbach: Why the parens?

ui/src/app/job-list/job-list.component.ts
```diff
@@ -133,6 +151,10 @@ export class JobListComponent implements OnInit {
     this.reloadJobs(this.route.snapshot.queryParams['q']);
   }
 
+  handleColumnsChanged() {
```
calbach: What is the purpose of this function? You seem to already be binding `this.displayFields` into the child component in the template

ui/src/app/job-list/job-list.component.ts
```diff
@@ -151,6 +173,26 @@ export class JobListComponent implements OnInit {
...
+
+  saveSettings() {
+    let fields: string[] = [];
+    this.displayFields.forEach((field) => {
+      if (field.primary) {
+        fields.push(field.field);
+      }
+    })
+    this.settingsService.setDisplayColumns(fields, this.projectId);
+    this.jobTable.setUpFieldsAndColumns();
```
calbach: Do you really need to call into the child component here? Can't you just update `this.displayFields`, thereby updating the child component which binds that value?

ui/src/app/core/settings.service.ts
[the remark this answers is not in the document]
bfcrampton: Yup looks great, thanks!

PR #461, opened 2018-08-14
More defensive 'of many' logic

### LINKED ISSUE #457: Header not correctly displaying total number of listed jobs
When the user jumps from dashboard page to job-list page by aggregation numbers, total number of listed jobs on the **top-right** corner is not correctly displayed.

It always displays `1 - 50 of NaN`

calbach:
Backends which don't implement `totalSize` will most likely return undefined (not return the property at all). Adjust the structure to better match. Reject all "falsy" `totalSize` values in the paginator (except 0).

Fixes #457


calbach:
@cjllanwarne just asked on Slack but then it apparently went down so not sure if you saw it - please review.

ui/src/app/shared/header/header.component.ts
```diff
@@ -281,8 +281,8 @@ export class JobsPaginatorIntl extends MatPaginatorIntl {
   getRangeLabel = (page: number, pageSize: number, length: number) => {
 
     let knownLength = null;
-
-    if (this.backendJobs.value.totalSize !== null) {
+    if (this.backendJobs.value.totalSize ||
```
cjllanwarne: This looks like it's intended to check that `totalSize` "is defined", or at least, "is not undefined"? And that need a special case because `0` counts as false?

I guess I'm a little bit surprised there's not a neater way to express an "is defined" check (eg `this.backendJobs.value.totalSize !== undefined`)?

ui/src/app/shared/job-stream.ts
```diff
@@ -19,7 +19,7 @@ export class JobStream extends BehaviorSubject<JobListView> {
               private request: QueryJobsRequest) {
     super({
       results: [],
-      totalSize: 0,
+      totalSize: undefined,
```
cjllanwarne: Not complaining (because I do think `undefined` *is* more appropriate in this case), just curious - would just changing this to `null` instead of `0` have solved the issue too?

calbach: This line apparently doesn't really matter AFAICT. Just something I noticed while making the change. It just felt better to me that the "unloaded state" (which I think is never displayed), would show "of many" rather than "of 0".

ui/src/app/shared/header/header.component.ts
calbach: Yes, that was the goal (and 0 is considered false).

I guess I'm just being extra safe since there was already a bug here. It's pretty common that a field would be null/undefined even though it's not documented as such. Your check should also have worked fine (`!== undefined`). If somehow a null showed up here (e.g. if a backend returned `totalResults: null`), this would catch it.

calbach: It is enforced by the compiler, but there's a cast going on in the API  layer when we receive a JSON response. It would have been more correct if that code instead placed nulls for these missing properties (given the type signature of our models), but it doesn't do that currently.

cjllanwarne:
My questions are satisfied - seems ready to merge as far as I'm concerned!

PR #463, opened 2018-08-15
Implement local storage

rsasch:
see #458 for details


ui/src/app/job-list/job-list.component.ts
```diff
@@ -75,6 +81,19 @@ export class JobListComponent implements OnInit {
...
     });
+
+    // set project ID (if any) and get display field info for list columns
+    const req = URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q']);
+    this.projectId = req.extensions.projectId || '';
+    const savedColumnSettings = this.settingsService.getDisplayColumns(this.projectId);
+
+    // assign this.displayFields to a copy of this.capabilities.displayFields and then
+    // update with saved settings, if any
+    this.capabilities.displayFields.forEach((df) => this.displayFields.push(Object.assign({}, df)));
```
calbach: `map` would probably be more natural here.

```
this.displayFields = this.capabilities.displayFields.map((df) => Object.assign({}, df));
```

calbach: Nice find on `Object.assign()`

ui/src/app/job-list/job-list.component.ts
```diff
@@ -151,6 +170,26 @@ export class JobListComponent implements OnInit {
...
+    this.displayFields.forEach((df) => {
+      if (df.field == field.field) {
+        df.primary = newValue;
+      }
+    })
+  }
+
+  saveSettings() {
+    let fields: string[] = [];
+    this.displayFields.forEach((field) => {
```
calbach:
```
fields = this.displayFields.filter(...)
```

rsasch: How would that work? `fields` is an array of string and `this.displayFields` is an array of DisplayField.

calbach: Good point, I missed that it was pulling out a subfield. filter + map would work, but I'm not sure I would recommend that over your current approach.

PR #464, opened 2018-08-20
Add page count to job list settings

rsasch:
Add "Displayed rows/page" to view settings on job list page, to be saved in local browser storage, which changes how many jobs are shown per page in the list view.

### DIFF
--- a/ui/src/app/core/settings.service.ts
+++ b/ui/src/app/core/settings.service.ts
@@ -45,13 +45,11 @@ export class SettingsService {
 
   /** return the display columns settings for a project or a clean scaffolding, if they don't exist */
   getDisplayColumns(projectId: string): string[] {
-    let settings = this.getSettingsForProject(projectId);
-    return settings.displayColumns;
+    return this.getSettingsForProject(projectId).displayColumns;
   }
 
   getPageSize(projectId: string): number {
-    let settings = this.getSettingsForProject(projectId);
-    return settings.pageSize;
+    return this.getSettingsForProject(projectId).pageSize;
   }
 
   /** update the display columns settings for a project */
@@ -89,9 +87,9 @@ export class SettingsService {
 
   private createEmptySettingsForProject(projectId: string): ProjectSettings {
     this.currentSettings.v1.projects.push({
-      'projectId': projectId,
-      'displayColumns': null,
-      'pageSize': null
+      projectId: projectId,
+      displayColumns: null,
+      pageSize: null
     });
     this.updateLocalStorage();
     return this.getSettingsForProject(projectId);
--- a/ui/src/app/core/settings.service.ts
+++ b/ui/src/app/core/settings.service.ts
@@ -13,7 +13,8 @@ type Settings = {
 type ProjectSettings = {
   // set to '' for non-project-based job manager
   projectId: string,
-  displayColumns: string[]
+  displayColumns: string[],
+  pageSize: number
   // ... other browser-based settings to be saved go here
 }
 
@@ -45,10 +46,12 @@ export class SettingsService {
   /** return the display columns settings for a project or a clean scaffolding, if they don't exist */
   getDisplayColumns(projectId: string): string[] {
     let settings = this.getSettingsForProject(projectId);
-    if (!settings) {
-      settings = this.createEmptySettingsForProject(projectId);
-    }
     return settings.displayColumns;
+  }
+
+  getPageSize(projectId: string): number {
+    let settings = this.getSettingsForProject(projectId);
+    return settings.pageSize;
   }
 
   /** update the display columns settings for a project */
@@ -62,8 +65,22 @@ export class SettingsService {
     this.updateLocalStorage();
   }
 
+  setPageSize(pageSize: number, projectId: string): void {
+    this.currentSettings.v1.projects.forEach((p) => {
+      if (p.projectId === projectId) {
+        p.pageSize = pageSize;
+        return;
+      }
+    });
+    this.updateLocalStorage();
+  }
+
   private getSettingsForProject(projectId: string): ProjectSettings {
-    return this.currentSettings.v1.projects.find(p => p.projectId === projectId);
+    let settings = this.currentSettings.v1.projects.find(p => p.projectId === projectId);
+    if (!settings) {
+      settings = this.createEmptySettingsForProject(projectId);
+    }
+    return settings;
   }
 
   private updateLocalStorage(): void {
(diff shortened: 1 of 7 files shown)


ui/src/app/job-list/job-list.component.html
```diff
@@ -23,10 +23,18 @@ <h2 class="settings-header">Displayed columns</h2>
...
       </mat-list-item>
     </mat-list>
-   <mat-divider></mat-divider>
-   <div class="settings-button-container">
-     <button class="apply-link" (click)="saveSettings()">Done</button>
-   </div>
+    <mat-divider></mat-divider>
+    <div class="field-container">
+      <h2 class="settings-header">Displayed rows/page</h2>
+      <mat-form-field class="page-count">
```
calbach: Is there a reason not to use the built-in page size picker? https://github.com/DataBiosphere/job-manager/blob/72fd13bfec97ed0abf211e514bc76a803a3527c6/ui/src/app/shared/header/header.component.html#L59

rsasch: Is there a way to use that and allow the user to enter in their own numeric value (and not a dropdown of options)?

calbach: Not that I'm aware of - here are the docs: https://material.angular.io/components/paginator/api

Is setting an arbitrary value really necessary?

rsasch:
@calbach Checked in with @ruchim and am waiting to hear back regarding what options make the most sense for a diverse set of users. Will re-factor to use `mat-paginator` in the meantime – thanks!

rsasch:
@rexwangcc @calbach I've re-worked the code to use `mat-paginator`; the exact page number options are still TBD, but the rest of it is ready for review (again).

ui/src/app/job-list/job-list.component.ts
```diff
@@ -67,8 +67,11 @@ export class JobListComponent implements OnInit {
         this.reloadJobs(this.route.snapshot.queryParams['q'], true);
       }
     });
+    const req = URLSearchParamsUtils.unpackURLSearchParams(this.route.snapshot.queryParams['q']);
+    this.projectId = req.extensions.projectId || '';
 
     this.header.pageSubject.subscribe(resp => this.onClientPaginate(resp));
+    this.pageSize = this.settingsService.getPageSize(this.projectId) || 50;
```
calbach: If you're going to leave it as an `@Input`, you should keep the initialization to 50 as it was before, and then only override that value here if getPageSize() is truthy.

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -133,6 +133,8 @@ describe('JobListComponent', () => {
     tick();
 
     testComponent = fixture.debugElement.query(By.css('jm-job-list')).componentInstance;
+    testComponent.pageSize = 3;
```
calbach: Hopefully you can revert per my other comment.

ui/src/app/job-list/job-list.module.ts
```diff
@@ -15,7 +15,7 @@ import {
   MatSelectModule,
   MatSortModule,
   MatTableModule,
-  MatTooltipModule
+  MatTooltipModule,
```
calbach: I would probably just revert this file since it's not relevant to the change anymore

ui/src/app/job-list/job-list.component.ts
```diff
@@ -165,6 +166,7 @@ export class JobListComponent implements OnInit {
   }
 
   private onClientPaginate(e: PageEvent) {
+    this.settingsService.setPageSize(e.pageSize, this.projectId);
```
calbach: This seems a bit chatty - perhaps compare to `this.pageSize` and only update if it has changed? This also means you won't be saving saving settings for someone who has never touched the page size controls.

ui/src/app/core/settings.service.ts
```diff
@@ -73,7 +90,8 @@ export class SettingsService {
   private createEmptySettingsForProject(projectId: string): ProjectSettings {
     this.currentSettings.v1.projects.push({
       'projectId': projectId,
-      'displayColumns': null
+      'displayColumns': null,
```
calbach: Why the string keys?

calbach: I think omitting these keys would be a better simulation of "not having settings". This is what older versions of the settings will look like.

ui/src/app/core/settings.service.ts
```diff
@@ -45,12 +46,14 @@ export class SettingsService {
...
   getDisplayColumns(projectId: string): string[] {
     let settings = this.getSettingsForProject(projectId);
-    if (!settings) {
-      settings = this.createEmptySettingsForProject(projectId);
-    }
     return settings.displayColumns;
   }
 
+  getPageSize(projectId: string): number {
+    let settings = this.getSettingsForProject(projectId);
```
calbach: nit: I wouldn't bother with this local variable.

ui/src/app/job-list/job-list.component.ts
rsasch: done

ui/src/app/job-list/job-list.component.spec.ts
rsasch: yup, thanks

ui/src/app/job-list/job-list.component.ts
rsasch: done

ui/src/app/core/settings.service.ts
rsasch: no real reason, changed

rsasch: When I take `displayColumns` and `pageSize` out of this function, I get the error
> Argument of type '{ projectId: string; }' is not assignable to parameter of type 'ProjectSettings'.
>  Property 'displayColumns' is missing in type '{ projectId: string; }'.

ui/src/app/core/settings.service.ts
rsasch: fixed for `getDisplayColumns` and `getPageSize `

ui/src/app/job-list/job-list.component.ts
```diff
@@ -166,7 +169,9 @@ export class JobListComponent implements OnInit {
   }
 
   private onClientPaginate(e: PageEvent) {
-    this.settingsService.setPageSize(e.pageSize, this.projectId);
+    if (e.pageSize != this.pageSize) {
+      this.settingsService.setPageSize(e.pageSize, this.projectId);
```
calbach: This should also update `this.pageSize`, else you won't save if the user switches back to 50

calbach: - User changes page size to 100
-> this saves
- User changes page size back to 50
-> this doesn't save (`this.pageSize` is still 50)

rsasch: ok, all set

PR #465, opened 2018-08-22
Add full support for submission time

rsasch:
- add submission to the regular query builder and API
 - take submission out of the queryExtensions endpoint for dsub
 - add submission to the query request for cromwell

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -321,9 +321,7 @@ def cromwell_query_params(query, page, page_size):
     return query_params
 
 
-def format_job(job, now):
-    if now is None:
-        now = datetime.utcnow()
+def format_job(job, now=datetime.utcnow()):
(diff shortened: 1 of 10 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -318,16 +321,15 @@ def cromwell_query_params(query, page, page_size):
     return query_params
 
 
-def format_job(job, now):
+def format_job(job, now=datetime.utcnow()):
```
calbach: I would recommend reverting this, I'm not sure when this actually gets evaluated but I know there are definitely some gotchas with non-primitive default parameters.

rsasch: ok, done

calbach: Here's more detail on this particular issue: http://blog.thedigitalcatonline.com/blog/2015/02/11/default-arguments-in-python/#default-arguments-evaluation

PR #467, opened 2018-08-24
Fix prod docker files.

rexwangcc:
I noticed that the docker files for prod are broken, for they are missing the generated models, which are required by `ng build --prod`. 

I think the reason why it was not identified for a while is that we used `ng build --target=production --env=dev` in the UI test, which doesn't have the exact same build mechanism with `ng build --prod`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,17 +180,17 @@ From v0.2.0, Job Manager starts to release stock docker images on [DockerHub](ht
 
 - To build the `job-manager-ui` image with `$TAG` from the root of this Github repository:
     ```
-    cd ui && docker build -t job-manager-ui:$TAG . -f Dockerfile
+    docker build -t job-manager-ui:$TAG . -f ui/Dockerfile
     ```
     
 - **Cromwell:** To build the `job-manager-api-cromwell` image with `$TAG` from the root of this Github repository:
     ```
-    docker build -t job-manager-api-cromwell:v0.2.0 . -f servers/cromwell/Dockerfile
+    docker build -t job-manager-api-cromwell:$TAG . -f servers/cromwell/Dockerfile
     ```
     
 - **dsub:** To build the `job-manager-api-dsub` image with `$TAG` from the root of this Github repository:
     ```
-    cd servers && docker build -t job-manager-api-dsub:v0.2.0 . -f dsub/Dockerfile
+    docker build -t job-manager-api-dsub:$TAG . -f servers/dsub/Dockerfile
     ```
 
 ### Add a github release pointing to the dockerhub images
--- a/servers/cromwell/Dockerfile
+++ b/servers/cromwell/Dockerfile
@@ -1,14 +1,24 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
+ADD ./ /job-manager
+
+RUN /bin/bash -c scripts/rebuild_swagger.sh
+RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5
+
 FROM gcr.io/google-appengine/python
 
 WORKDIR /app
 
-ADD servers/jm_utils /app/jm_utils
-ADD servers/cromwell/jobs /app/jobs
-COPY servers/cromwell/requirements.txt /app/jobs
+COPY --from=0 /job-manager/servers/jm_utils /app/jm_utils
+COPY --from=0 /job-manager/servers/cromwell/jobs /app/jobs
+COPY ./servers/cromwell/requirements.txt /app/jobs
+
 RUN cd jobs && pip install -r requirements.txt
 # We installed jm_utils so don't need local copy anymore, which breaks imports
 RUN rm -rf jm_utils
 
 # Missing required arguments -b PORT, -e ... which must be provided by the
 # docker image user.
-ENTRYPOINT ["/bin/bash", "/scripts/await_md5_match.sh", "/app/jobs/models/.jobs.yaml.md5", "--", "gunicorn", "jobs:run()"]
+ENTRYPOINT ["gunicorn", "jobs:run()"]
--- a/servers/dsub/Dockerfile
+++ b/servers/dsub/Dockerfile
@@ -1,3 +1,12 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
+ADD ./ /job-manager
+
+RUN /bin/bash -c scripts/rebuild_swagger.sh
+RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5
+
 FROM gcr.io/google-appengine/python
 RUN virtualenv --no-download /env -p python
 
@@ -7,13 +16,13 @@ ENV VIRTUAL_ENV /env
 ENV PATH /env/bin:$PATH
 
 WORKDIR /app
-ADD jm_utils /app/jm_utils
-ADD dsub/jobs /app/jobs
-ADD dsub/requirements.txt /app/jobs
+COPY --from=0 /job-manager/servers/jm_utils /app/jm_utils
+COPY --from=0 /job-manager/servers/dsub/jobs /app/jobs
+COPY ./servers/dsub/requirements.txt /app/jobs
 RUN cd jobs && pip install -r requirements.txt
 # We installed jm_utils so don't need local copy anymore, which breaks imports
 RUN rm -rf jm_utils
 
 # Missing required arguments -b PORT, -e ... which must be provided by the
 # docker image user.
-ENTRYPOINT ["/bin/bash", "/scripts/await_md5_match.sh", "/app/jobs/models/.jobs.yaml.md5", "--", "/env/bin/gunicorn", "jobs.__main__:app"]
+ENTRYPOINT ["/env/bin/gunicorn", "jobs.__main__:app"]
--- a/ui/Dockerfile
+++ b/ui/Dockerfile
@@ -1,14 +1,28 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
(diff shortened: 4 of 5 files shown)


calbach:
@cjllanwarne should review, but based on my understanding I don't think this is going to work properly for local dev. The build step only runs if you've never built the image, or else explicitly are passing `--build` (uncommon). Therefore await_md5 won't always run, which is what we want for local dev.

rexwangcc:
@calbach Thanks for the comments here, I thought UI should be OK since it is using `Dockerfile.dev` for local builds. For SHIM dockers, as you said, I think I need some help/inputs from @cjllanwarne about how to harmonize this.

cjllanwarne:
The problem (I think) is that there are two use cases here which are fundamentally different:

1. Developer: 
  - I want a docker image with a big hole where the model classes should be, which can be filled in when I run `docker-compose up` by the rebuild swagger task docker image, and regenerated live if and when I make changes to the swagger API definition.
  - I don't want to prebuild the model because otherwise there's no way to override with local development changes without a full `docker build`
2. Deployer
  - I want a full, finished and self contained docker image
  - I therefore *do* want the model to be prebuilt into the docker image.

I think the way to resolve this going to have to be separate `dockerfile`s - one for "developer"s and one for "deployer"s, each with the appropriate amount of prebuilding done and built-into the image

rexwangcc:
@cjllanwarne I strongly second having separate `dockerfile`s, the **only** concern is that this will somehow "lose" one of the advantages of using docker that having 1 environment from development to deployment, and may potentially bring hidden bugs to production docker images, due to the difference between development(testing) images and prod images.

However, I cannot think of a better way to address the problem for now, so let me start the iteration in this direction, and see how it works?

rexwangcc:
@calbach @bfcrampton Any objections for having separate docker images(like it's discussed above) on dsub shim side? I saw `Dockerfile.local` and `Dockerfile` in dsub shim folder already, does that mean we need to have:
- `Dockerfile.local.dev`
- `Dockerfile.local.prod`
- `Dockerfile.dev`
- `Dockerfile.prod`
docker files? Please correct me if I'm wrong!

calbach:
@rexwangcc yep, sounds good to me. It's a reasonable thing to do, we just haven't needed to for the shims yet because Python dev and prod look so similar.

I'd just make them match what we've done in the UI directory which is to leave off a suffix for the prod version (or else you could update UI to match your new proposal):

- `Dockerfile`
- `Dockerfile.dev`
- `Dockerfile.local`

Don't worry about a prod+local variant, it wouldn't really make sense to run that.

rexwangcc:
@calbach Thanks a lot for the inputs here! I will go ahead and make the changes!

README.md
```diff
@@ -180,17 +180,17 @@ From v0.2.0, Job Manager starts to release stock docker images on [DockerHub](ht
...
 - To build the `job-manager-ui` image with `$TAG` from the root of this Github repository:
     ```
-    cd ui && docker build -t job-manager-ui:$TAG . -f Dockerfile
+    docker build -t job-manager-ui:$TAG . -f ui/Dockerfile
     ```
     
 - **Cromwell:** To build the `job-manager-api-cromwell` image with `$TAG` from the root of this Github repository:
     ```
-    docker build -t job-manager-api-cromwell:v0.2.0 . -f servers/cromwell/Dockerfile
+    docker build -t job-manager-api-cromwell:$TAG . -f servers/cromwell/Dockerfile
```
cjllanwarne: Maybe mention setting `$TAG` to (eg) `0.2.0` somewhere above?

README.md
```diff
@@ -180,17 +180,17 @@ From v0.2.0, Job Manager starts to release stock docker images on [DockerHub](ht
 
 - To build the `job-manager-ui` image with `$TAG` from the root of this Github repository:
     ```
-    cd ui && docker build -t job-manager-ui:$TAG . -f Dockerfile
+    docker build -t job-manager-ui:$TAG . -f ui/Dockerfile
```
cjllanwarne: I assume this works? I just want to check the `Dockerfile` doesn't have any relative paths that are upset because the PWD is now different...?

servers/dsub/Dockerfile
```diff
@@ -1,3 +1,12 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
+ADD ./ /job-manager
+
+RUN /bin/bash -c scripts/rebuild_swagger.sh
+RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5
```
cjllanwarne: You probably don't need this, since it only exists to delay startup until the swagger is rebuilt (which is guaranteed to have already happened here)

ui/Dockerfile
```diff
@@ -1,14 +1,28 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
+ADD ./ /job-manager
+
+RUN /bin/bash -c scripts/rebuild_swagger.sh
+RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5
```
cjllanwarne: Same comment here

servers/cromwell/Dockerfile
```diff
@@ -1,14 +1,24 @@
+FROM openjdk
+
+WORKDIR /job-manager
+
+ADD ./ /job-manager
+
+RUN /bin/bash -c scripts/rebuild_swagger.sh
+RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5
```
cjllanwarne: Same comment here

cjllanwarne:
This all looks good to me, but I'll allow @calbach or @bfcrampton give it the final thumb.

As a side note, I assume you've tested that all of these `Dockerfile`s build successfully based on the instructions in the README?

servers/dsub/Dockerfile
cjllanwarne: - Specifically, you probably don't need the `RUN scripts/await_md5_match.sh /job-manager/ui/src/app/shared/model/.jobs.yaml.md5` line 
- You definitely *do* need the `RUN /bin/bash -c scripts/rebuild_swagger.sh` line!

README.md
rexwangcc: Good point, will do!

servers/cromwell/Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
+# Note: This is the dockerfile for development purposes
+
+FROM gcr.io/google-appengine/python
```
samanehsan: Why does this use a different base image than the cromwell `Dockerfile`?

servers/dsub/Dockerfile
rexwangcc: Oh, I see what you mean. I didn't realize the functionality of this script. I think we can take it out. Same for other images

ui/nginx.prod.conf
```diff
@@ -13,8 +13,8 @@ http {
 
     # If we're running on App Engine, logs will appear on the Google Developer's
     # Console when logged to this directory.
-    access_log /var/log/app_engine/app.log;
-    error_log /var/log/app_engine/app.log;
+    # access_log /var/log/app_engine/app.log;
```
samanehsan: Why is this commented out now? If it's not required anymore could we remove it instead?

README.md
rexwangcc: Yes this works. I have to change the `PWD` because some of the commands within the Dockerfile actually refers to files in the upper lever to the Dockerfile itself.

servers/cromwell/Dockerfile.dev
rexwangcc: I copied this from the master branch. so this was the current `Dockerfile` in master branch.
The reason that the new production `Dockerfile` using `openjdk` is that `rebuild_swagger.sh` requires `java`. So I have to use the `multi-stage` docker build to make the production image work.

samanehsan: Got it! Thanks for the explanation!

ui/nginx.prod.conf
rexwangcc: This line is required by the App Engine, but will break the image if you don't use the GAE and you don't override this default nginx conf file. I commented out just as a fix, but need inputs from @bfcrampton or @calbach to determine whether we want to consolidate it in this way.

calbach: I would just remove these lines, we (dsub deploy) should be overriding the nginx file on our end, rather than cromwell doing it since the GAE is more deployment-specific.

rexwangcc: done!

rexwangcc:
@cjllanwarne @calbach I made the changes you suggested, and I made sure the following docker images can be built correctly, could you take another look and give 👍 or comments? Thanks!
- `docker build -t job-manager-ui:$TAG . -f ui/Dockerfile`
- docker build -t job-manager-api-cromwell:$TAG . -f servers/cromwell/Dockerfile
- docker build -t job-manager-api-dsub:$TAG . -f servers/dsub/Dockerfile

deploy/quickstart/quick_start.py
```diff
@@ -190,9 +190,8 @@ def make_or_replace_or_reuse_directory(path_to_make, exit_if_invalid):
                     valid = False
                     time_to_exit = True
             else:
-                print(
-                    'Invalid input: got {0} but expected one of (1/2)'.format(
-                        provided))
+                print('Invalid input: got {0} but expected one of (1/2)'.
```
bfcrampton: minor fyi: `{}` works the same as `{0}`, but don't need to fix

rexwangcc:
I'm going to merge this PR, any objections?

PR #471, opened 2018-08-31
Support retryable dsub jobs

bfcrampton:
Added a new "Attempt" column for the dsub job manager. Also supports querying by attempt, and now doesn't break when you try to see the details page for a retried job.

### DIFF
--- a/servers/dsub/jobs/controllers/jobs_controller.py
+++ b/servers/dsub/jobs/controllers/jobs_controller.py
@@ -28,8 +28,8 @@ def abort_job(id):
     """
     # Attempt is unused in aborting because only one attempt can be running at
     # a time.
-    project, job, task, _ = job_ids.api_to_dsub(id, _provider_type())
-    provider = providers.get_provider(_provider_type(), project, _auth_token())
+    proj_id, job_id, task_id, _ = job_ids.api_to_dsub(id, _provider_type())
+    provider = providers.get_provider(_provider_type(), proj_id, _auth_token())
 
     # TODO(bryancrampton): Add flag to ddel to support deleting only
     # 'singleton' tasks.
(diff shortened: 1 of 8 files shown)


servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -48,7 +48,9 @@ def abort_job(id):
     # hacky re-routing of stdout once dsub removes it from the python API
     deleted = execute_redirect_stdout(lambda:
         ddel.ddel_tasks(
-            provider=provider, job_ids={job_id}, task_ids=task_ids))
+            provider=provider,
+            job_ids={job},
+            task_ids= {task} if task else None))
```
mbookman: Remove the space after the equal sign.

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -73,28 +75,26 @@ def get_job(id):
     Returns:
         JobMetadataResponse: Response containing relevant metadata
     """
-    proj_id, job_id, task_id = job_ids.api_to_dsub(id, _provider_type())
-    provider = providers.get_provider(_provider_type(), proj_id, _auth_token())
+    project, job, task, attempt = job_ids.api_to_dsub(id, _provider_type())
+    provider = providers.get_provider(_provider_type(), project, _auth_token())
```
mbookman: In the `dsub` code base, we have tried to keep to the nomenclature that a `job` or `task` is an object of some sort (typically a dict), whereas `job_id` or `task_id` is used for the actual ID. Is this intentionally different here?

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Good catch, done.

servers/dsub/jobs/controllers/jobs_controller.py
bfcrampton: Good point, thinking about it more I agree its better to leave as `job_id` and `task_id`. Was just trying to shorten the line so it didn't wrap, fixed.

PR #479, opened 2018-09-20
fixed job status conversion and added test

rsasch:
Looks like I left off a status in my `controllers/utils/job_statuses.py` – fixing that and adding a test for them.

### DIFF
--- a/servers/cromwell/jobs/test/test_job_statuses.py
+++ b/servers/cromwell/jobs/test/test_job_statuses.py
@@ -1,0 +1,24 @@
+# coding: utf-8
+
+from __future__ import absolute_import
+
+from . import BaseTestCase
+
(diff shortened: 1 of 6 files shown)


servers/cromwell/jobs/test/test_job_statuses.py
```diff
@@ -0,0 +1,21 @@
...
+
+from __future__ import absolute_import
+
+from . import BaseTestCase
+
+from jobs.controllers.utils import job_statuses
+
+
+class TestJobStatuses(BaseTestCase):
+    def test_cromwell_execution_statuses_conver_correctly(self):
```
calbach: converts

rsasch: thanks for catching that!

Issue #316, opened 2018-07-06
Archiving Jobs


cjllanwarne:
As a user of the job manager UI, sometimes a job will fail and after investigation, I want to never see that job again in my standard views. i.e. when looking at a page of all failed jobs in the UI, I only want to see new or unresolved failures. In previous Zamboni UI at Broad, this was called 'abandoning' the job.

This would need to be a hard coded filter, where any jobs that have this flag set are automatically not shown in the UI. We will use a settings menu to turn on or off seeing archived jobs by default.

When a user does a query, archived jobs would be included in the query results.

Cromwell 33 has been released with the ability to filter out based on labels. This will enable us to use our archive feature. 

AC:
- [x] Archiving works in JMUI. This means when a user flags a job as archive, the job is automatically filtered out from view. It will appear as a filter chip and can be added/removed through the query builder.
- [x] Ability to `Remove from archive`

![image.png](https://images.zenhubusercontent.com/5a9f2a304b5806bc2bc9b485/10cfe074-4084-4f3c-910f-893ea0383c32)

PR #481, opened 2018-09-25
Add hideArchived to query builder

rsasch:
- added hideArchived as possible query extension
- added support in Cromwell shim for hideArchived
- added support in UI for boolean filter chips
- added hideArchived and all other query filters to local browser storage


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -321,6 +321,10 @@ def cromwell_query_params(query, page, page_size):
     query_params.append({'additionalQueryResultFields': 'parentWorkflowId'})
     query_params.append({'additionalQueryResultFields': 'labels'})
     query_params.append({'includeSubworkflows': 'false'})
+    if query.extensions:
```
bfcrampton: nit: single if:
```
if query.extensions and query.extensions.hide_archived:
```
(can switch back to two if/when more query extensions are needed or cromwell)

ui/src/app/core/settings.service.ts
```diff
@@ -52,6 +55,10 @@ export class SettingsService {
     return this.getSettingsForProject(projectId).pageSize;
   }
 
+  getHideArchived(projectId: string): boolean {
+    return this.getSettingsForProject(projectId).hideArchived;
+  }
+
   /** update the display columns settings for a project */
```
bfcrampton: nit: I'd remove this comment even though you didn't add it. It's pretty clear what this function is doing and the other similar ones don't have comments.

ui/src/app/core/settings.service.ts
```diff
@@ -86,10 +103,16 @@ export class SettingsService {
   }
 
   private createEmptySettingsForProject(projectId: string): ProjectSettings {
+    const capabilities = this.capabilitiesService.getCapabilitiesSynchronous();
+    let hideArchivedValue;
+    if (capabilities.queryExtensions) {
+      hideArchivedValue = capabilities.queryExtensions.includes('hideArchived');
```
bfcrampton: So, just to be clear, the default behavior for a project it to hide archived jobs? And you'd have to explicitly set it to **not** do that.

ui/src/app/shared/common.ts
```diff
@@ -38,7 +39,8 @@ export const queryDataTypes: Map<string, FieldDataType> = new Map([
 
 export const queryExtensionsDataTypes: Map<string, FieldDataType> = new Map([
   ['projectId', FieldDataType.Text],
-  ['userId', FieldDataType.Text]
+  ['userId', FieldDataType.Text],
```
bfcrampton: It feels to me like the `FieldDataType` of a query extensions should be provided as part of the `CapabilitiesResponse` rather than encoded separately in a static UI file. I know I initially added this, and we can for sure leave it to another PR, but do you have any sense of how difficult it would be to refactor the field type into part of the capabilities response?

ui/src/app/core/settings.service.ts
```diff
@@ -14,7 +15,8 @@ type ProjectSettings = {
   // set to '' for non-project-based job manager
   projectId: string,
   displayColumns: string[],
-  pageSize: number
+  pageSize: number,
+  hideArchived: boolean
```
bfcrampton: This should be more generic. Would it be difficult to generalize it to something like:
```
{
  ...
  filters: Filter[] 
}

type Filter = {
  key: string
  value: any
}
```

That way we don't have to special case `hideArchived` in the UI, we can just store all current filters in settings and when you return back to the JM for that project the same filters will automatically be in place. I realize this is a much bigger ask, but I'm pretty against hard coding things like this in the UI since it sends us down the wrong path of adding code specific to a given backend. Ideally this would either be implemented in a general way or not stored in local storage at first.

From a product standpoint, why is this the only filter we want to store? It seems like either all your previous filters should show up, or none of them... @ruchim

rsasch: @bfcrampton for the associated user story, see issue #316

ui/src/app/shared/common.ts
rsasch: I don't think it would be too difficult.  There's already the concept of type for the `displayFields` part of capabilities. Can you create a new issue for this?

ui/src/app/core/settings.service.ts
rsasch: @ruchim How unique is the `hideArchived` scenario?  Would users like to have all the attributes of the last query they used show up in new windows or the next time they use JMUI?

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: k, done

ui/src/app/core/settings.service.ts
rsasch: k, done

ui/src/app/core/settings.service.ts
rsasch: If it's in the query extensions, yes.

ui/src/app/shared/common.ts
bfcrampton: Done https://github.com/DataBiosphere/job-manager/issues/483

rsasch: thanks!

ui/src/app/core/settings.service.ts
rsasch: @bfcrampton Got a clarification offline – new requirement is to retain all the query filters in local storage.  Will notify you when it's ready for review again

ui/src/app/core/settings.service.ts
```diff
@@ -85,11 +106,22 @@ export class SettingsService {
     return this.localStorage.setItem('settings', JSON.stringify(this.currentSettings));
   }
 
-  private createEmptySettingsForProject(projectId: string): ProjectSettings {
+  createEmptySettingsForProject(projectId: string): ProjectSettings {
+    const capabilities = this.capabilitiesService.getCapabilitiesSynchronous();
+    let filters: Filter[] = [];
+    if (capabilities.queryExtensions) {
+      capabilities.queryExtensions.forEach((f) => {
+        const dataType = queryDataTypes.has(f) ? queryDataTypes.get(f) : queryExtensionsDataTypes.get(f);
```
bfcrampton: Since we're iterating through queryExtensions don't we know this field is in `queryExtensionsDataTypes` and not `queryDataTypes` (otherwise it wouldn't be an extension)? Might be missing something though.

ui/src/app/core/settings.service.ts
```diff
@@ -85,11 +106,22 @@ export class SettingsService {
     return this.localStorage.setItem('settings', JSON.stringify(this.currentSettings));
   }
 
-  private createEmptySettingsForProject(projectId: string): ProjectSettings {
+  createEmptySettingsForProject(projectId: string): ProjectSettings {
+    const capabilities = this.capabilitiesService.getCapabilitiesSynchronous();
+    let filters: Filter[] = [];
+    if (capabilities.queryExtensions) {
```
bfcrampton: Am I reading this correctly in that any queryExtension which is for a boolean field will automatically be applied as `= True`? Doesn't matter right now since we only have the one, but I'd put a comment in `common.ts` around `queryExtensionsDataTypes` indicating that any new fields added there which are of the type `Boolean` will automatically be applied. We may have to just special case adding `hideArchived` here if we ever add any boolean queryExtensions in the future which we do not want to be auto applied.

ui/src/app/core/settings.service.ts
```diff
@@ -73,6 +84,16 @@ export class SettingsService {
     this.updateLocalStorage();
   }
 
+  setFilters(filters: Filter[], projectId: string): void {
+    this.currentSettings.v1.projects.forEach((p) => {
```
bfcrampton: tiny nit: I'd prefer to just use a regular loop and break rather than a `forEach` here, feels a bit more readable to me:
```
for (let p of this.currentSettings.v1.projects) {
  if (p.projectId === projectId) {
    p.filters = filters;
    break;
  }
}
```

ui/src/app/projects/projects.component.html
```diff
@@ -4,7 +4,7 @@
       <input matInput
              autofocus
              class="input-field"
-             placeholder="Project ID"
+             placeholder="ProjectID"
```
bfcrampton: Why the change? I think it looks cleaner with the space

ui/src/app/shared/header/header.component.ts
```diff
@@ -245,7 +261,7 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
       j => this.onHoldStatuses.includes(j.status)).length;
   }
 
-  private refreshChips(query: string): void {
+  refreshChips(query: string): void {
```
bfcrampton: nit: why the change?

ui/src/app/core/settings.service.ts
rsasch: I've added a comment; please let me know if you don't think it's clear enough.

ui/src/app/projects/projects.component.html
rsasch: No good reason – happened during debugging the tests; fixed.

ui/src/app/shared/header/header.component.ts
rsasch: Happened during debugging the tests; fixed.

PR #489, opened 2018-10-05
Add 'hide archived' setting to job list settings UI

rsasch:
- added support in Cromwell shim for hideArchived
- added slide toggle to settings menu in job list for 'hide archived'
- added hideArchived to local browser storage


api/jobs.yaml
```diff
@@ -407,6 +407,10 @@ definitions:
         type: string
         description: >
           Returns only jobs with the specified user ID.
+      hideArchived:
+        type: string
```
calbach: Shouldn't this be a boolean?

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -44,6 +45,8 @@ describe('JobListComponent', () => {
 
   // Jobs with IDs JOB0 -> JOB4.
   function testJobs(count: number): QueryJobsResult[] {
+    const flags = ['archive','','archive','','archive'];
```
calbach: opt: `new Array(3).fill('archive');`

ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -329,6 +335,28 @@ describe('JobListComponent', () => {
     expect(de.queryAll(By.css('.fake-projects')).length).toEqual(1);
   }));
 
+  it('does not display the hide archived setting without the right project setting', async(() => {
```
calbach: Nice

ui/src/app/job-list/job-list.component.ts
```diff
@@ -71,9 +73,9 @@ export class JobListComponent implements OnInit {
     this.projectId = req.extensions.projectId || '';
 
     this.header.pageSubject.subscribe(resp => this.onClientPaginate(resp));
-    const savedPageSettings = this.settingsService.getPageSize(this.projectId);
-    if (savedPageSettings) {
-      this.pageSize = savedPageSettings;
+    this.savedProjectSettings = this.settingsService.getSettingsForProject(this.projectId);
```
calbach: Hmm, I think I probably prefer calling settingsService each time. I would worry about this local copy of the settings going out of sync over time.

api/jobs.yaml
rsasch: fixed

ui/src/app/job-list/job-list.component.spec.ts
rsasch: actually, this isn't needed anymore, so I removed it altogether

ui/src/app/job-list/job-list.component.ts
rsasch: ok – changed

ui/src/app/core/settings.service.ts
```diff
@@ -43,37 +46,30 @@ export class SettingsService {
...
-    this.currentSettings.v1.projects.forEach((p) => {
-      if (p.projectId === projectId) {
-        p.displayColumns = fields;
-        return;
+  getSavedSettingValue(settingKey: string, projectId: string): any {
+    let value = null;
+    const settings = this.getSettingsForProject(projectId);
+    for (let key in settings) {
+      if (key == settingKey) {
+        value = settings[key];
```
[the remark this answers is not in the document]
rsasch: fixed

ui/src/app/job-list/job-list.component.html
```diff
@@ -23,6 +23,10 @@ <h2 class="settings-header">Displayed columns</h2>
                       (click)="$event.stopPropagation()">{{ df.display }}</mat-checkbox>
       </mat-list-item>
     </mat-list>
+    <div *ngIf="settingsService.getSavedSettingValue('hideArchived', projectId) != null">
```
calbach: nit: I would generally consider this too heavy to put in a template, would probably extract this into a helper method on the component.

Issue #442, opened 2018-07-28
Handle Cromwell outage more gracefully


ruchim:
Today, trying to navigate to the caas dev job manager resulted in a plain text page that just said 500 server error. This was due to problems in CaaS (which were in turn due to problems in Sam).

The user experience could be improved. We could show a custom error page with a more user friendly message. Or we could show the regular job manager landing page and just leave the jobs table empty, with a message saying that we're having trouble talking to CaaS.

cjllanwarne:
I think there's a whole class of "make initial connections easier to set up" since this is closely related to my experience in #441

PR #495, opened 2018-10-18
Slightly improve upon the default error message

aednichols:
Reassure people that they have indeed reached Job Manager, and that it is running, but there's a problem with navigation or reaching Cromwell (or dsub... neutrally worded).

This will probably get restyled along with the rest of Job Manager anyway, hence the "enterprise CRM system circa 1995" aesthetic.

Closes https://github.com/DataBiosphere/job-manager/issues/442

screen shot 2018-10-18 at 5 19 49 pm

### DIFF
--- a/ui/src/app/core/initial-error/initial-error.component.ts
+++ b/ui/src/app/core/initial-error/initial-error.component.ts
@@ -8,7 +8,7 @@ import {NavigationError, Router} from '@angular/router';
   styleUrls: ['./initial-error.component.css']
 })
 export class InitialErrorComponent {
-  initialLoadErrorHeader: string;
+  errorMessage: string;
   initialLoadFailure = false;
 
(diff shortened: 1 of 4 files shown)


ui/src/app/core/initial-error/initial-error.component.ts
```diff
@@ -23,7 +23,7 @@ export class InitialErrorComponent {
         this.initialLoadFailure = true;
         const status = e.error.status || 'unknown';
         const title = e.error.title || 'Unknown error';
-        this.initialLoadErrorHeader = `${status}: ${title}`
+        this.errorMessage = `${status}: ${title}`
```
calbach: Will also need to rename property at line 11

rsasch: :arrow_up:  this is why the UI test is failing

PR #497, opened 2018-10-22
fix for whitespace in label values

rsasch:
Filtering by a label value with whitespace in it currently doesn't work. I added URL decoding to the UI and Cromwell shim layer.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -247,7 +247,8 @@ def query_jobs(body, **kwargs):
     auth = kwargs.get('auth')
     headers = kwargs.get('auth_headers')
     query = QueryJobsRequest.from_dict(body)
-    query.labels = _format_query_labels(query.labels)
+    if query.labels is not None:
+        query.labels = _format_query_labels(query.labels)
     query_page_size = query.page_size or _DEFAULT_PAGE_SIZE
     offset = 0
     if query.page_token is not None:
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -247,8 +247,7 @@ def query_jobs(body, **kwargs):
     auth = kwargs.get('auth')
     headers = kwargs.get('auth_headers')
     query = QueryJobsRequest.from_dict(body)
-    if query.labels is not None:
-        query.labels = _format_query_labels(query.labels)
+    query.labels = _format_query_labels(query.labels)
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -246,6 +247,8 @@ def query_jobs(body, **kwargs):
     auth = kwargs.get('auth')
     headers = kwargs.get('auth_headers')
     query = QueryJobsRequest.from_dict(body)
+    if query.labels is not None:
```
aednichols: It might be nice to make `_format_query_labels` resilient to empty input instead of protecting it at each call site (which, granted, is only one right now)

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -143,7 +143,7 @@ import {TimeFrame} from "../model/TimeFrame";
     let chips: Map<string, string> = new Map();
     urlSearchParams.paramsMap.forEach((values: string[], key: string) => {
       if (values && key) {
-        chips.set(key, values.toString());
+        chips.set(key, decodeURIComponent(values.toString()));
```
aednichols: No chance of a key having spaces? (probably true, but I can't immediately see where that's enforced)

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -367,3 +369,10 @@ def _parse_datetime(date_string):
 
 def _get_base_url():
     return current_app.config['cromwell_url']
+
+
+def _format_query_labels(orig_query_labels):
+    query_labels = {}
+    for key, val in orig_query_labels.items():
+        query_labels[key] = urllib.unquote(val).decode('utf8')
```
aednichols: Are you sure this `decode()` is necessary? 

It seems that `unquote()` returns a string, while `decode()` [operates on bytes](https://docs.python.org/3/library/stdtypes.html#bytes.decode).

I think we may be taking the UTF-8 encoded string returned by `unquote()`, casting it to an array of bytes, and then decoding it into the same UTF-8 string.

servers/cromwell/jobs/controllers/jobs_controller.py

ui/src/app/shared/utils/url-search-params.utils.ts
rsasch: Good point – fixed in shim layer, too

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: removed `decode()`, and it still worked

PR #505, opened 2018-10-31
reclassify 'Aborted' as a 'Failed' job state

### LINKED ISSUE #380: Aborted jobs go to failed filter
After more discussion with users, there were feelings that aborted jobs should go to failed filter. The idea would be that users would abort the jobs and then mark them as archived later.

In reality we may want an aborted status of it's own... need to talk to Users

AC:
Aborted jobs go to failed filter, or failed counts in future dashboard view

rsasch:
Closes #380

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
           name: Install detox and yapf
           # detox fails with newer versions of tox: https://github.com/tox-dev/detox/issues/33
           command: |
-            pip install detox tox==3.4.0 yapf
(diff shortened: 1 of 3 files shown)


rsasch:
@samanehsan Maybe I'm misunderstanding your comment, but I made changes to `header.component.ts` [here](https://github.com/DataBiosphere/job-manager/pull/505/files#diff-7394b459d815f59b294bb1c40c03b543).

samanehsan:
Ah nevermind! For some reason I thought this was a different file 🤦‍♀️

PR #507, opened 2018-10-31
Added icon to task list that indicates task was call-cached

### LINKED ISSUE #498: Add indication of task being call-cached
Add new column to the job detail page task list where an icon can go if that task was call-cached.  Possible icons to use are:
screen shot 2018-10-23 at 1 37 43 pm
screen shot 2018-10-23 at 1 37 48 pm
screen shot 2018-10-23 at 1 38 25 pm
Tooltip text should be: "This job was cached".

rsasch:
- added icon to indicate a call-cached task
- moved shard status icon into the same position for consistency

Closes #498

screen shot 2018-10-31 at 5 09 10 pm

### DIFF
--- a/ui/src/app/job-details/tasks/tasks.component.html
+++ b/ui/src/app/job-details/tasks/tasks.component.html
@@ -9,23 +9,36 @@
           <mat-cell *matCellDef="let t" class="task-title">
             <a class="title-link" *ngIf="t.jobId" href="{{ '/jobs/' + t.jobId }}">{{ t.name }}</a>
             <p *ngIf="!t.jobId" class="title-nolink">{{ t.name }}</p>
-            <button *ngIf="isScattered(t)" mat-icon-button class="shard-icon" [matMenuTriggerFor]="shardMenu">
+          </mat-cell>
+        </ng-container>
+
+        <!-- Shard Statuses column -->
+        <ng-container matColumnDef="taskInfoIcons">
+          <mat-header-cell *matHeaderCellDef class="slim-cell"></mat-header-cell>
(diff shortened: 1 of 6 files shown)


ui/src/app/job-details/tasks/tasks.component.html
```diff
@@ -9,23 +9,36 @@
...
+
+        <!-- Shard Statuses column -->
+        <ng-container matColumnDef="taskInfoIcons">
+          <mat-header-cell *matHeaderCellDef class="slim-cell"></mat-header-cell>
+          <mat-cell *matCellDef="let t" class="slim-cell">
+            <button *ngIf="t.shardStatuses" mat-icon-button class="shard-icon" [matMenuTriggerFor]="shardMenu">
               <clr-tooltip>
                 <clr-icon clrTooltipTrigger shape="layers" size="24"  class="is-solid"></clr-icon>
-                <clr-tooltip-content clrPosition="right"  clrSize="xs" *clrIfOpen>
+                <clr-tooltip-content clrPosition="left"  clrSize="xs" *clrIfOpen>
```
rexwangcc: Not familiar with the UI here, so I have a curious question: what would the task look like if a task is both call-cached and scattered? My guess is that the "scatter" icon will sit to the left?

rsasch: I believe a scattered task cannot be call-cached, but I might be wrong. @ruchim Can you chime in on this one?

rexwangcc: Per the offline discussion, this shouldn't be a problem, and we can also update it in a separate PR!

PR #508, opened 2018-11-02
Include authuser on GCS links

### LINKED ISSUE #486: Google Auth should use the appropriate Google account for bucket links
I'm not sure of the best way to handle this, but currently if we click on a bucket link in the UI, it always uses the first user logged in in the browser, so the console UI displays a 403 if that user doesn't have access to the bucket.

When GCP console handles this in Console UI, they use `authuser=x` param in all links. I think it should be possible to use that here as well, based on the list of logged in users on the client.

calbach:
closes #486

### DIFF
--- a/deploy/quickstart/quick_start.py
+++ b/deploy/quickstart/quick_start.py
@@ -191,8 +191,9 @@ def make_or_replace_or_reuse_directory(path_to_make, exit_if_invalid):
                     valid = False
                     time_to_exit = True
             else:
-                print('Invalid input: got {0} but expected one of (1/2)'.
-                      format(provided))
+                print(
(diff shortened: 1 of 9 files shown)


ruchim:
@calbach -- any update here? Thanks!

gemmalam:
@ruchim CH messaged me on slack that he would circle back to finish this work this week.

ruchim:
perfect!

rsasch:
@calbach Is this change still necessary since the log links are now going to Google Console, instead of Google Storage (see #561)?

calbach:
Yes. Cloud console will 403 if you are multilogged in and you don't have access to the bucket on your primary account. This should be the case with either Cloud Console or the direct storage URL.

calbach:
Thanks

PR #509, opened 2018-11-06
Add better UI for labels of type 'list' in query builder

### LINKED ISSUE #501: Make 'flag' field a drop down in query builder
In my mind it makes sense for the flag field in the query builder to be a drop down of options just like it is in the list view and bulk editor instead of free text. It would prevent silly mistakes like the one I just made, where I searched for `flag:archived` instead of `flag:archive`. 

image

image

rsasch:
Currently, the query builder treats all labels as being of type `text`, so even if it's of type `list` (which means there is a set of possible values to choose from), the user has to get the value just right.  

screen shot 2018-11-06 at 5 48 22 pm

This PR adds `enum` type of filter chip (radio buttons) for labels of type `list`.

screen shot 2018-11-09 at 4 48 26 pm

Closes #501


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -382,7 +380,13 @@ def _get_base_url():
...
     if orig_query_labels is None:
         return None
-    query_labels = {}
+    query_labels = []
     for key, val in orig_query_labels.items():
-        query_labels[urllib.unquote(key)] = urllib.unquote(val)
+        if isinstance(val, list):
+            for value_or in val:
+                query_labels.append({
+                    'labelor': str(key) + ':' + str(value_or)
```
rexwangcc: Not sure but as I remembered `lableor` is not supported by the current CaaS 🤔

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -86,20 +100,20 @@ import {TimeFrame} from "../model/TimeFrame";
             break;
           }
           case FieldDataType.Enum: {
-            // Handle enum data types. Currently this is only statuses, if we
-            // add additional ones this has to be updated.
-            let statuses: JobStatus[] = [];
-            for (let status of urlSearchParams.getAll(key)) {
-              statuses.push(JobStatus[status]);
+            let options: string[] = [];
+            for (let option of urlSearchParams.getAll(key)) {
```
calbach: nit: `value = urlSearchParams.getAll(key)` ?

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -64,18 +65,31 @@ import {TimeFrame} from "../model/TimeFrame";
...
-      if (queryDataTypes.has(key) || queryExtensionsDataTypes.has(key)) {
-        // If this is a known field, handle the data type explicitly
-        var value: any;
-        let dataType = queryDataTypes.has(key) ? queryDataTypes.get(key) : queryExtensionsDataTypes.get(key);
+      let labelField: DisplayField = null;
+      if (capabilities) {
+        labelField = capabilities.displayFields.find(f => f.field === 'labels.' + key);
+      }
+      if (queryDataTypes.has(key) || queryExtensionsDataTypes.has(key) || labelField) {
+        // If this is label, use the settings to determine type
```
calbach: I'm not necessarily opposed, but why do we need to do this? The fallback is just to treat any unknown as a label

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -382,7 +380,13 @@ def _get_base_url():
 def _format_query_labels(orig_query_labels):
     if orig_query_labels is None:
         return None
-    query_labels = {}
+    query_labels = []
     for key, val in orig_query_labels.items():
-        query_labels[urllib.unquote(key)] = urllib.unquote(val)
+        if isinstance(val, list):
```
calbach: This PR adds undocumented support to the API for single OR list values in label queries. This should be made explicit in the documentation. Personally I'd prefer to document using array values across the board instead of supporting a union type (string|[]string). dsub would need to fail if it saw >1 value. Arguably this should be something in the capabilities response, whether these kinds of "or" queries are supported.

ui/src/app/shared/header/chips/enum-selection.component.ts
```diff
@@ -0,0 +1,41 @@
...
+      this.currentChipValues.splice(this.currentChipValues.indexOf(optionValue), 1);
+    }
+    this.updateValue.emit(this.currentChipValues.join(','));
+  }
+
+  isChecked(status: string): boolean {
+    return this.currentChipValues.indexOf(status) > -1;
+  }
+
+  trimOptions(options: string[]): string[] {
```
calbach: private, if possible

ui/src/app/shared/header/chips/enum-selection.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    let newOptions: string[] = [];
+    options.forEach((option) => {
```
calbach: nit: feels like a `return options.filter(...)`

rsasch:
@calbach @rexwangcc Thank you for your comments.  I've re-worked this change to use radio buttons instead of checkboxes for `list` type labels.

servers/cromwell/jobs/test/test_jobs_controller.py
```diff
@@ -539,7 +539,7 @@ def _request_callback(request, context):
                 'stdout': std_out,
                 'taskName': 'analysis',
                 'timestamp': response_timestamp
-             },{
+            },{
```
calbach: unrelated?

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -154,7 +154,14 @@ import {TimeFrame} from "../model/TimeFrame";
     let queryFields = new Map<string, FieldDataType>(queryDataTypes);
 
     if (capabilities.commonLabels) {
-      capabilities.commonLabels.map(label => queryFields.set(label, FieldDataType.Text));
+      capabilities.commonLabels.map(label => {
```
calbach: existing issue, but would use `forEach` if you're not consuming the return value of the `map()` call

ui/src/app/shared/header/chips/enum-selection.component.ts
```diff
@@ -0,0 +1,32 @@
...
+export class EnumSelectionComponent implements OnInit {
+  @Input() initialChipValue: string;
+  @Input() chipKey: string;
+  @Input() chipOptions: string[];
+  @Output() updateValue: EventEmitter<string> = new EventEmitter();
+
+  currentChipValue: string;
+
+  ngOnInit(): void {
+    if (this.initialChipValue != null) {
```
calbach: Is this check really needed? Isn't the `currentChipValue` just going to be `undefined` anyways?

ui/src/app/shared/utils/url-search-params.utils.ts
```diff
@@ -154,7 +154,14 @@ import {TimeFrame} from "../model/TimeFrame";
...
+        let currentField = capabilities.displayFields.find( f => f.field == 'labels.' + label);
+        if (currentField && currentField.validFieldValues) {
+          queryFields.set(label, FieldDataType.Enum);
```
calbach: The function doc should be updated, currently says "non-label fields"

ui/src/app/shared/header/header.component.ts
```diff
@@ -76,8 +79,7 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
       this.chips = URLSearchParamsUtils.getChips(this.route.snapshot.queryParams['q']);
     }
 
-    this.options = URLSearchParamsUtils.getQueryFields(
-      this.capabilitiesService.getCapabilitiesSynchronous());
+    this.options = URLSearchParamsUtils.getQueryFields(this.capabilities);
```
calbach: nit: I probably prefer the previous way to avoid adding more instance variables

ui/src/app/shared/header/chips/filter-chip.component.ts
```diff
@@ -57,6 +61,14 @@ export class FilterChipComponent implements OnInit {
     return this.chipKey + ': ' + this.currentChipValue;
   }
 
+  getChipValues(): string[] {
+    const capabilities = this.capabilitiesService.getCapabilitiesSynchronous();
+    let labelField = capabilities.displayFields.find(f => f.field === 'labels.' + this.chipKey);
```
calbach: const

ui/src/app/shared/header/chips/enum-selection.component.ts
```diff
@@ -0,0 +1,32 @@
+import {Component, EventEmitter, Input, OnInit, Output} from "@angular/core";
+
+@Component({
+  selector: 'jm-enum-selection',
+  templateUrl: './enum-selection.component.html',
+})
+export class EnumSelectionComponent implements OnInit {
```
calbach: Should the `Status` chip use this?

rsasch: Can you clarify?  The `Status` chip is a special case where there can be multiple values (e.g. checkboxes in the UI).  All other fields and labels have only one value, so ones with set values to choose from should use this filter chip type (e.g. radio buttons in UI).

calbach: Good point - forgot about this change. Ignore

servers/cromwell/jobs/test/test_jobs_controller.py
rsasch: unrelated, but it's a formatting fix I spotted while working on this

ui/src/app/shared/header/chips/enum-selection.component.ts
rsasch: It's not undefined if there is a relevant parameter in the query string (e.g. `?q=flag%3Dfollow-up`)

ui/src/app/shared/utils/url-search-params.utils.ts
rsasch: Ok, done

ui/src/app/shared/utils/url-search-params.utils.ts
rsasch: nice catch – updated the comment

ui/src/app/shared/header/header.component.ts
rsasch: makes sense

ui/src/app/shared/header/chips/filter-chip.component.ts
rsasch: changed

PR #514, opened 2018-11-09
Improve usability of query builder

### LINKED ISSUE #510: Improvements to the Query Builder
- [ ] Filter icon instead of magnifying glass
- [ ] Make sure suggestion appears whenever field has focus

- [ ] Border all four sides
- [ ] Prompt text (possibly something generic like `Filter Workflows`)
inspired by google cloud console:
screen shot 2018-11-07 at 2 13 34 pm

rsasch:
Addresses issues brought up in usability testing:
-  Filter icon instead of magnifying glass
-  Make sure suggestion appears whenever field has focus
-  Border all four sides
-  Prompt text (possibly something generic like Filter Workflows)

before:
query-builder-before

after:
screen shot 2018-11-09 at 2 36 56 pm

Closes #510

### DIFF
--- a/ui/src/app/shared/header/header.component.css
+++ b/ui/src/app/shared/header/header.component.css
@@ -1,9 +1,33 @@
 .content {
   margin-left: 1rem;
+  margin-right: 1rem;
 }
 
 .query-form {
-  width: 100%;
+  width: 60%;
+  min-width: 60rem;
+  padding-top: 0.5rem;
+  line-height: 1.5rem;
+}
+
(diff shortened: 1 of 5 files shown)


ui/src/app/shared/header/header.component.css
```diff
@@ -19,6 +43,7 @@
   display: flex;
   justify-content: space-between;
   align-items: center;
+  min-width: 60rem;
```
calbach: This is quite large - are you sure about this value as a minimum?

ruchim:
We walked through the UI design together, looks good to me!

Issue #340, opened 2018-07-13
Support clickable label filters on the table view


ruchim:
![image](https://user-images.githubusercontent.com/14941133/42712709-a5b5471e-86ba-11e8-9b3c-4096fab8e3d3.png)

As a user, I want to be able to quickly filter by a label that's on a job I'm looking at. See attached wireframe.

rsasch:
How does this work with the current functionality where clicking on a label value allows users to edit that value?

rsasch:
@ruchim This currently works for all `text` fields that aren't `editable`.  Does that satisfy this requirement?

ruchim:
@rsasch I think its great that the filtering works for fields that aren't editable!

As a suggestion, for something like Labels -- is it possible to do both filter on a value as is, and the option to edit it?

rsasch:
@ruchim It's possible, but we don't have a user experience/design for how this would look/happen.

ruchim:
Closing as these changes have been merged to v0.5.1 and above.

Issue #319, opened 2018-07-06
Configurable List View


cjllanwarne:
~~@kbergin, @rsasch, @ruchim: could you add a quick description of what "Configurable List View" means to you in Q3?~~

From elastc issue 301:

---

As a user of a job manager where it is used by anyone besides myself, I want to be able to set the labels that I see in the list view based on what I find the most useful or based on what my project uses (in the case of JMUI aaS). The label options will be from a set of best practice suggestions (white list), although a user could use other labels and see them on the details view. 

AC:
- [x] User can select (in UI) up to 5 columns to view in list view from white list
- [x] User can set order of columns preferably by drag and drop but if that's difficult than another method works.
- [x] JMUI remembers the columns they want to see when they return
- [ ] Any columns they have in view are options for filtering

![15_columnsettings_5ad1160090bddf04e8fd3e40_207112933149202157168_5132_](https://user-images.githubusercontent.com/13006282/42525839-36f08eb2-8442-11e8-86fd-3575e8d5e2f7.png)

kbergin:
@cjllanwarne https://elastc.com/c/vHHUknxz/301-configure-list-view-labels-in-ui contains the description. Could we consider using the existing tickets to create new ones here, as the content is all there in elastc already.

cjllanwarne:
Thanks @kbergin - I've copied the content here (is that what you had in mind when you said "using the existing tickets to create new ones here"?)

ruchim:
@rsasch -- what's left to complete in this issue? Seems like the functionality already exists.

rsasch:
@ruchim The last checkbox, "Any columns they have in view are options for filtering," hasn't been completed, because:
 - label columns can't be both editable and filterable
 - there's no user story for filtering by `status` or date-based columns (submission, start, end)

ruchim:
@rsasch -- I tried filtering by status in the query bar and it worked, and same for submission/start/end.

I think it's okay if one can't filter by editable filters for now -- I'll file it as a separate issue since the last thing in the AC that's incomplete, and the feature around configuring columns in JMUI is complete, so I'd like to close this issue.

PR #516, opened 2018-11-14
Add more functionality to the job details menu on the job list page

### LINKED ISSUE #512: Enhance pop-up panel for workflow details
Today, when using the pop-up panel on the list view when getting details for a particular workflow, the only option that exists is to get workflow details:

The acceptance criteria for this issue to enhance the capability of this button to include:
- [ ] link to the timing diagram
- [ ] the ability to filter for workflows with that name

While this particular design includes other content/features, just attaching it from the perspective of re-using the design/layout of this pop-up panel:

rsasch:
Addresses issues brought up in usability testing regarding the utility of the job details menu and what additional functionality might be helpful.
 
Before:
job-details-before

After:
screen shot 2018-11-14 at 5 03 28 pm

Closes #512


ui/src/app/job-list/table/table.component.html
```diff
@@ -41,23 +41,38 @@
...
+                </a>
+              </mat-list-item>
+              <mat-list-item *ngIf="hasTimingUrl(j)">
+                <a href="{{ j.extensions.timingUrl }}" target="_blank">
+                  <clr-icon shape="hourglass" size="24"></clr-icon>
+                  <span>Timing Diagram</span>
+                </a>
+              </mat-list-item>
+              <mat-list-item *ngIf="j.name">
+                <a [href]="'?q=name%3D' + j.name">
```
calbach: I'm not sure this is going to work if you already have filters applied. Also should clarify what the desired behavior would be in that case:

- Should it intersect with existing filters?
- Or, discard existing filters and just apply this

ui/src/app/job-list/table/table.component.html
```diff
@@ -41,23 +41,38 @@
...
+                  <clr-icon shape="filter-2" size="24"></clr-icon>
+                  <span>Filter by <strong>name:{{ j.name }}</strong></span>
```
calbach: I'm not sure I follow the product thinking here. Why are we expediting "filter by name" here vs other criteria? Is this because the button happens to hang off of the job name column?

rsasch: The ability to filter the job list by the `key:value` pair for a particular cell in the job table has been a proposed UI idea (see https://github.com/DataBiosphere/job-manager/issues/340 and
https://github.com/DataBiosphere/job-manager/issues/319, for example).  Because the `name` column already has a menu attached to it, it is easier to add it here first.  Eventually, the idea is that this will be something the user can do anywhere in the table.

ui/src/app/job-list/table/table.component.html
rsasch: Those are good points – I will re-work how this is implemented and let you know when it's ready for re-review.

rsasch: @calbach I've generalized the functionality into something that can be re-used for other column's filters.

ui/src/app/job-list/job-list.component.ts
```diff
@@ -159,6 +159,17 @@ export class JobListComponent implements OnInit {
     this.reloadJobs(this.route.snapshot.queryParams['q']);
   }
 
+  handleFiltersChanged(filter: string) {
+    const [key, value] = filter.split(':');
+    if (this.header.chips.has(key)) {
+      this.header.updateValue('statuses', value);
```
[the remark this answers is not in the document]
rsasch: ugh, thanks for catching that!

ui/src/app/job-list/table/table.component.ts
```diff
@@ -116,6 +117,12 @@ export class JobsTableComponent implements OnInit {
       .catch((error) => this.handleError(error));
   }
 
+  filterOnFieldValue(field: string, value: string) {
+    if (field && value) {
+      this.onFiltersChanged.emit(field + ':' + value);
```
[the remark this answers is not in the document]
rsasch: ok – changed to an array

PR #522, opened 2018-11-16
set different filter behavior for status checkboxes

### LINKED ISSUE #437: Query builder checkbox UX
Noticed while using the query builder that you can only check one box at a time and then the window closes. You have to click back in to the filter chip to check another item. It would also be helpful to have a way to say "show me all except this one" with fewer clicks. Right now I had to click back into the chip 5 times to select each status except submitted.

AC:
Improved UX for query builder checkbox functionality

rsasch:
Because a user can select multiple statuses to filter the job list by, change the `click` behavior so that the user can do so without having to choose `statuses` from the query builder menu each time.

Closes #437

### DIFF
--- a/ui/src/app/shared/header/chips/filter-chip.component.ts
+++ b/ui/src/app/shared/header/chips/filter-chip.component.ts
@@ -74,8 +74,11 @@ export class FilterChipComponent implements OnInit {
   }
 
   setChipValue(value: string): void {
+    console.log(this.getCurrentChipType());
     if (value) {
-      this.chipMenuTrigger.closeMenu();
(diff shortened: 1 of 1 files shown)


ui/src/app/shared/header/chips/filter-chip.component.ts
```diff
@@ -74,8 +74,11 @@ export class FilterChipComponent implements OnInit {
   }
 
   setChipValue(value: string): void {
+    console.log(this.getCurrentChipType());
```
rexwangcc: Is this a debugging line?

PR #521, opened 2018-11-16
Make job/task status icons more intuitive

rsasch:
- change the `aborted`, `aborting`, `failed`, `on hold`, and `succeeded` icons in response to user testing
- added tooltips for shard status icons

screen shot 2018-11-16 at 4 16 29 pm

Closes #517

PR #525, opened 2018-11-20
improve the clarity of the the status buttons

### LINKED ISSUE #518: Status options for quick filtering
Today on Job Manager, the quick filters exist for statuses.

- Active = Running
- Completed = Aborted & Succeeded
- On Hold = On Hold

Usability testing shows that:
1. The term "Completed" was very confusing and not easy to associate with statuses Succeeded and/or Aborted.
2. It was hard to make the link when the quick filters have a different name than the actual status.

To ease those concerns, it may be best to make the association between status & quick filters more obvious by displaying each of the most prominent statuses as its own quick filter:

For example, a quick filter called "Running" to show all workflows in "Running" status. And the same for Succeeded, Failed, Aborted & On Hold.

rsasch:
After noting issues during user testing, change the quick filter status buttons to correspond to just one status.  After feedback from demo, add status icon.

Before:
screen shot 2018-11-20 at 1 53 29 pm

After:
screen shot 2018-11-26 at 3 15 26 pm

Closes #518

### DIFF
--- a/ui/src/app/job-list/job-list.component.spec.ts
+++ b/ui/src/app/job-list/job-list.component.spec.ts
@@ -184,7 +184,7 @@ describe('JobListComponent', () => {
 
     // We don't finish loading until we complete a query that matches the
     // the ActiveRoute; so trigger a real navigate to clear the loading state.
-    de.query(By.css('.active-button')).nativeElement.click();
+    de.query(By.css('.running-button')).nativeElement.click();
     tick();
     fixture.detectChanges();
     expect(de.queryAll(By.css('.spinner-container')).length).toEqual(0);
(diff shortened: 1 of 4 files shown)


rsasch:
@rexwangcc Users found the fact that the status buttons didn't correspond to actual statuses confusing (i.e. what does "active" really mean?).  This PR uses whatever icons are returned by the `JobStatusIcon` enum.  A different PR (#521) will change those icons.

rexwangcc:
👍 for 
> A different PR (#521) will change some of those icons.

And sorry for the conusfion around my first question, I was asking if `Aborting` and `Submitted` are not as important as other statuse for users to filter by shortcut buttons(tehy can still filter from the query builder)? The answer seems to be yes?

rsasch:
@ruchim Could you reply to :arrow_up: ?

ruchim:
@rexwangcc -- As both "Submitted" and "Aborting" are short-term transition statuses, we wanted to limit the quick filtered status to be what would be most common filters, which seem to be active workflows, terminal workflows, or intentionally on-hold workflows. This is meant to narrow down the options to simplify the UI. And exactly like you said, they are still valid statuses to query by in the query builder!

rexwangcc:
@ruchim Thanks for the explanation, it makes sense to me!

PR #537, opened 2018-12-11
Add button to clear the query builder chips

### LINKED ISSUE #388: Clear all filters
As a user, I want to be able to quickly clear the filters I've applied to my list view.

AC:
Clear button clears filter chips
Should not clear the `projectId` chip.

rsasch:
before:
qb-before

after:
screen shot 2018-12-11 at 12 21 55 pm

Will not clear the `projectId` chip, as requested.

Closes #388

### DIFF
--- a/ui/src/app/shared/header/header.component.ts
+++ b/ui/src/app/shared/header/header.component.ts
@@ -198,6 +198,18 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
     this.search();
   }
 
+  hasChipsToClear(): boolean {
+    return this.getChipKeys().filter(key => key != 'projectId').length > 0;
+  }
+
+  removeAllChips(): void {
+    this.chips.forEach((value: string, key: string) => {
+      if (key != 'projectId') {
+        this.chips.delete(key);
+      }
+    });
+  }
+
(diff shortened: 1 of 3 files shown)


ui/src/app/shared/header/header.component.ts
```diff
@@ -198,6 +198,19 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
     this.search();
   }
 
+  hasChipsToClear(): boolean {
+    return this.getChipKeys().filter(key => key != 'projectId').length > 0;
```
rexwangcc: Just a curious question, why exclude `projectId ` from the query keys? Is that an invisible but fixed key for the query builder?

rsasch: It was part of the request (see #388); I'm assuming it's because if it's being used (possibly only for dsub?) it's not likely to change while a user is looking at jobs. @kbergin ?

kbergin: I do believe the reasoning for this was related to dsub. For dsub when you log in as a certain project it would put a project id chip in the query builder, but if you close out that chip I think it logs you out of the job manager. I'm not sure if that still holds true, my information is probably outdated :)

Issue #434, opened 2018-07-27
Subworkflow links should be router links


kbergin:
@samanehsan FYI; I ( @calbach )  noticed that clicking a subworkflow resulted in a full page load, which is going to be a lot slower than using a routerLink. See https://github.com/DataBiosphere/job-manager/blob/master/ui/src/app/job-list/table/table.component.html#L38

This would become irrelevant if we move to collapsible subworkflows and no longer have a separate details page.

rsasch:
@kbergin Was this addressed by #540?

kbergin:
I'm not sure, @rsasch . I think this was something that is an implementation detail that I'm not familiar with, I think I was making the ticket for @calbach

calbach:
I don't know where the subworkflow link is if it still exists. If that button still exists and clicking that refreshes the whole page, then the bug still exists.

rsasch:
@calbach Here is the call to the `router` that I was referring to in that PR: https://github.com/DataBiosphere/job-manager/pull/540/files#diff-28e7af551c1526a3e1844c2c1727b115R61

rsasch:
Resolved by #540

PR #540, opened 2019-01-07
Add 'Navigate Up' arrow to sub-job details

### LINKED ISSUE #536: Add ability to navigate from sub workflow details to parent workflow details
image
Tooltips for icons will be  “Go back to parent job” and “Go back to list view"

rsasch:
- add icon to sub-job details page
<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/1713505/49829865-95b12300-fd5d-11e8-8b94-b84862726f2e.png">image</a>
- maintain query filters from job list page

Closes #536

### DIFF
--- a/ui/src/app/job-details/job-details.component.ts
+++ b/ui/src/app/job-details/job-details.component.ts
@@ -40,6 +40,14 @@ export class JobDetailsComponent implements OnInit {
     });
   }
 
+  handleNavUp(): void {
+    this.router.navigate(['/jobs/' + this.job.extensions.parentJobId], {
+      queryParams: {
+        'q': this.route.snapshot.queryParams['q']
+      }
+    });
+  }
+
   hasResources(): boolean {
     return (this.job.inputs && Object.keys(this.job.inputs).length !== 0)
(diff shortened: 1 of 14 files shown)


ui/src/app/job-details/job-details.component.ts
```diff
@@ -40,6 +42,46 @@ export class JobDetailsComponent implements OnInit {
     });
   }
 
+  handleNavUp(): void {
+    if (this.job.extensions.parentJobId) {
+      this.router.navigate(['/jobs/' + this.job.extensions.parentJobId], {
```
calbach: Should just use a routerLink in the template if possible so that t behaves like a proper link, e.g. ctrl+click does the right thing

calbach: (in context of #434  - but I guess this did address the bug - opening in a new window)

rsasch: I tried using a `routerLink`, but it wouldn't reload the components in the page, it would only update the URL.

PR #543, opened 2019-01-09
Move inputs & outputs from Resources panel to Tabs

### LINKED ISSUE #535: Move the inputs and outputs tabs int the tab group that contains the list of tasks
Inputs and outputs are currently too prominent in the job details page.  Adding them to the last list tab set will keep them accessible but free up some vertical space.

rsasch:
To increase the available real estate on the job details page, move the inputs and outputs into their own tabs and out of the Resources panel.

Before:
screen shot 2019-01-09 at 2 19 12 pm

After:
screen shot 2019-01-09 at 2 18 26 pm

Closes #535


ui/src/app/job-details/tabs/tabs.component.ts
```diff
@@ -87,6 +87,18 @@ export class TaskDetailsComponent implements OnInit, OnChanges {
     return false;
   }
 
+  hasFailures(): boolean {
+    return this.job.failures && (this.job.failures.length > 1)
+  }
+
+  hasInputs(): boolean {
+    return this.job.inputs && (Object.keys(this.job.inputs).length > 1)
```
rexwangcc: this was `Object.keys(this.job.inputs).length !== 0` before, I think there are no differences between them based on my understanding right?

rsasch: changed

PR #545, opened 2019-01-14
simplify display of day/times

### LINKED ISSUE #474: Sometimes date appears for today, sometimes not
This could be intended behavior, but I noticed that the date appeared next to a lot of our workflows from today (Sept 12) alongside the time. And then the two most recent workflows only had the time. Not sure why!

image

rsasch:
The handling of dates is confusing to users; simplify so all day/times are displayed the same.

- treat current day, current year and another year differently
- add tooltip rollover to show complete date and time

Closes #474


ui/src/app/shared/pipes/short-date-time.pipe.ts
```diff
@@ -6,20 +6,8 @@ import {Pipe, PipeTransform} from '@angular/core';
 })
 export class ShortDateTimePipe extends DatePipe implements PipeTransform {
 
-  isToday(date: Date): boolean {
-    if (date) {
-      let today: Date = new Date();
-      return date.getUTCDate() == today.getUTCDate() &&
```
calbach: I read the issue. It sounds to me @kbergin was reporting a bug, which is that we were showing timestamps from the current day with and without the date. THe source of the bug is likely that we're using UTC time here rather than local time. I don't think @kbergin was requesting you remove the differential timestamp entirely

rsasch: I agree that @kbergin was not asking for the display to be simplified, but the complexity of logic that is required to figure out time zones just to display the day/time differently if it's today doesn't seem commensurate with actual value to the user.

calbach:
I disagree, I think it's a useful visual distinction and it's a pretty
common UI pattern to omit the date in this case. If this is what the rest
of the team wants though I'll concede.

rsasch:
@ruchim Could you weigh in?

ruchim:
@rsasch @calbach I went traversing through some other UIs that our users utilize to see if we can find a solution that's consistent with the pattern they're familiar with.

screen shot 2019-01-22 at 3 13 52 pm
screen shot 2019-01-22 at 3 13 03 pm


I prefer a solution that's least disruptive to the existing users, so I would prefer to always include date & timestamp for that reason. I'll also tag @martygardner if he has any strong preferences for any UX considerations worth bringing up...

calbach:
Note: A pattern we've followed often in the jmui is the Gmail inbox (in
this case, it of course won't show a full date for an email received
today). I don't think users will find either approach very surprising. As a
user I just prefer to minimize redundant content on the page.

martygardner:
I think there are a couple issues here. 

To me the problem described in the bug is that we're not showing _any_ date reference for workflows submitted today. That, I agree, seems wrong: there should always be an indication of the date and time.

A separate question is whether we should show the exact same format for all dates. Here, I think there are benefits to altering the format in different cases to simplify the display. For example, saying a job was created "Today" will be understood more quickly than saying "Jan 23", which the user may have to check against a calendar to confirm is today.

So the rules I would propose are:
- Always include a date reference
- If the date is today (and possibly also yesterday), use those terms instead of the calendar date format -- _e.g. Yesterday, 7:44 PM_
- Only include the year in the date if it differs from the current year
- Use three-letter month abbreviations
- Don't include seconds in the timestamps

rsasch:
@martygardner @calbach @ruchim How about this:

- Current day
screen shot 2019-01-24 at 1 57 35 pm

- Not current day, but current year
screen shot 2019-01-24 at 1 59 39 pm


- Not current year (I figure if it's not the current year, time doesn't add much value)
screen shot 2019-01-24 at 1 58 08 pm

calbach:
I like it. Ideally full timestamp on hover in all cases as well (could be a follow-up).

ruchim:
I love the suggestion Bec, it's very little text and looks clean. I also like the idea of full timestamp on hover, but I don't want to delay this PR further --so Bec I'll let you pick if you want me to file a new issue for it or just add it along here.

rsasch:
@calbach @ruchim @martygardner Adding in a hover timestamp shouldn't be to onerous and doesn't seem too off-topic.  Just to confirm the format, I'm thinking
> Jan 23, 2019 8:19PM

That OK?

calbach:
I'd suggest including seconds on hover as well - usually if I'm digging in like this I'm debugging something and I want high precision, e.g. I might be correlating this with another timestamp from other logs.

martygardner:
LGTM

ui/src/app/job-details/panels/panels.component.html
```diff
@@ -37,13 +37,31 @@
...
+        <span clrTooltipTrigger> {{ job.submission | jmShortDateTime }}</span>
+        <clr-tooltip-content clrPosition="top-right" clrSize="xs" *clrIfOpen>
+          <span>{{ job.submission | jmFullDateTime }}</span>
+        </clr-tooltip-content>
+      </clr-tooltip>
     </p>
     <p id="started" *ngIf="job.start">
-      Started: {{ job.start | jmShortDateTime }}
+      Started:
+      <clr-tooltip>
```
calbach: opt: A lot of repetition - this could be extracted to a shared component

rsasch: @calbach Good idea – done.

PR #546, opened 2019-01-14
Return more accurate scattered task status

### LINKED ISSUE #542: Scatter jobs report as Succeeded even when shards are still Running
screen shot 2019-01-09 at 3 22 36 pm

In this image, the `BaseRecalibrator` task is scattered over many shards. When one of those shards completes, the entire task is reported as `Succeeded` even when all the other shards are still `Running`. Thus, icon and status denoted at the task level is not accurate in reflecting what is actually going on. The task should wait until all shards have succeeded before reporting its status as `Succeeded`

rsasch:
Take all returned shard statuses into account when figuring out the status of a scattered task.

Closes #542

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -403,7 +403,7 @@ def _get_scattered_task_status(metadata):
     }
     # return status by ranked applicability
     for status in [
-            'Failed', 'Aborted', 'Aborting', 'Running', 'Submitted',
+            'Failed', 'Aborting', 'Aborted', 'Running', 'Submitted',
             'Succeeded'
     ]:
         if status in statuses:
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -397,7 +397,10 @@ def _format_query_labels(orig_query_labels):
 
 def _get_scattered_task_status(metadata):
     # get all shard statuses
-    statuses = {shard.get('executionStatus') for shard in metadata}
+    statuses = {
+        task_statuses.cromwell_execution_to_api(shard.get('executionStatus'))
+        for shard in metadata
+    }
     # return status by ranked applicability
     for status in [
             'Failed', 'Aborted', 'Aborting', 'Running', 'Submitted',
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -394,3 +393,15 @@ def _format_query_labels(orig_query_labels):
     for key, val in orig_query_labels.items():
         query_labels[urllib.unquote(key)] = urllib.unquote(val)
     return query_labels
+
+
+def _get_scattered_task_status(metadata):
```
aednichols: This is a nice pure function that seems like a great candidate for a unit test

rsasch:
@aednichols Thanks for catching that;`task_statuses.cromwell_execution_to_api` call has been put back in.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -394,3 +393,18 @@ def _format_query_labels(orig_query_labels):
...
+    # get all shard statuses
+    statuses = {
+        task_statuses.cromwell_execution_to_api(shard.get('executionStatus'))
+        for shard in metadata
+    }
+    # return status by ranked applicability
+    for status in [
+            'Failed', 'Aborted', 'Aborting', 'Running', 'Submitted',
```
aednichols: Should `Aborted` and `Aborting` switch places? It seems that one wouldn't want to display `Aborted` if some aborts haven't finished

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: 👍 Added

servers/cromwell/jobs/controllers/jobs_controller.py

servers/cromwell/jobs/test/test_task_statuses.py
```diff
@@ -27,3 +30,83 @@ def test_cromwell_execution_to_api_maps_all_task_execution_statuses_correctly(se
     def test_unrecognized_task_status_causes_exception(self):
         with self.assertRaises(ValueError):
             task_statuses.cromwell_execution_to_api('Not a valid task status')
+
+    def test_scattered_task_status(self):
```
aednichols: Test looks great

PR #553, opened 2019-01-23
take out check for exhaustive when contructing a new paginator

### LINKED ISSUE #476: Total job count doesn't refresh when you change filter
Small thing we noticed today. We have two project ids we're filtering on that are very similar at first glance but have different time stamps. If we filter by the first one, ending in 45Z, it shows 996 total workflows and displays those workflows in the list view. 

image

When I change the filter to the other time stamp, ending in 55Z and click enter, it refreshes the list of filtered workflows but not the total number of workflows for the pagination.
image
image

If I refresh the page, there are actually 999 total workflows that match that filter
image

If I x out the filter chip entirely, the same bug is observed. It still says 999 workflows even though there are 3649 after I refresh the page.
image
image

rsasch:
Closes #476

### DIFF
--- a/ui/src/app/shared/header/chips/filter-chip.component.ts
+++ b/ui/src/app/shared/header/chips/filter-chip.component.ts
@@ -7,7 +7,6 @@ import {URLSearchParamsUtils} from "../../utils/url-search-params.utils";
 import {CapabilitiesService} from "../../../core/capabilities.service";
 import {MatMenuTrigger} from "@angular/material";
 import {DatepickerInputComponent} from "./datepicker-input.component";
-import {JobStatus} from "../../model/JobStatus";
 
 @Component({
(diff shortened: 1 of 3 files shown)


ui/src/app/shared/header/chips/filter-chip.component.ts
```diff
@@ -7,7 +7,6 @@ import {URLSearchParamsUtils} from "../../utils/url-search-params.utils";
 import {CapabilitiesService} from "../../../core/capabilities.service";
 import {MatMenuTrigger} from "@angular/material";
 import {DatepickerInputComponent} from "./datepicker-input.component";
-import {JobStatus} from "../../model/JobStatus";
```
calbach: nit: unrelated - would revert to keep this PR clean (even if it is unused)

PR #555, opened 2019-01-31
added sign out button to job list when user is authenticated

### LINKED ISSUE #348: Sign out button

rsasch:
Closes #348

### DIFF
--- a/ui/src/app/shared/header/header.component.ts
+++ b/ui/src/app/shared/header/header.component.ts
@@ -26,6 +26,7 @@ import {
   PageEvent,
 } from '@angular/material';
 
+import {AuthService} from '../../core/auth.service';
 import {CapabilitiesService} from '../../core/capabilities.service';
 import {URLSearchParamsUtils} from '../utils/url-search-params.utils';
 import {JobStatus} from '../model/JobStatus';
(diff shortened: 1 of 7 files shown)


ui/src/app/shared/header/header.component.ts
```diff
@@ -272,6 +274,19 @@ export class HeaderComponent implements OnInit, AfterViewInit, AfterViewChecked,
     this.onDisplayFieldsChanged.emit(this.displayFields);
   }
 
+  isSignedIn(): boolean {
+    return !!this.authService.userId;
```
aednichols: My interpretation of this is
```
!this.authService.userId
```
casts and negates `userId` to `false` if it exists, then the next `!` negates that again.

Is there a way to cast without negating twice, such as (pseudo-code)
```
return (boolean)this.authService.userId
```
?

aednichols: ( not a blocker though )

rsasch: This is the way I've seen it done in other parts of the codebase, so I copied that. It looks like `return Boolean(this.authService.userId);` will behave the same. Would that be preferable?

aednichols: Consistency > nitpicking

Issue #355, opened 2018-07-13
Embed Timing Diagram into the UI


ruchim:
It will look nicer to have the timing diagram inside of our UI instead of linking out. Maybe it is possible to reuse some of cromwell's code that makes the timing diagram now

AC
Timing diagram included within UI

PR #559, opened 2019-02-06
Embed timing diagram within job details page

rsasch:
- uses ng2-google-charts library
- does not use `/timing` endpoint, instead is based on metadata that is already there

Closes: #355


ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
```diff
@@ -0,0 +1,36 @@
...
+  buildTimelineData(metadata: TaskMetadata[]): void {
+    this.timelineChart = {
+      chartType: 'Timeline',
+      dataTable: []
+    };
+    this.timelineChart.dataTable.push(['Num', 'Name', 'Start', 'End']);
+    let counter = 1;
+
+    metadata.forEach((task) => {
+      this.timelineChart.dataTable.push([counter.toString(), task.name, task.start, task.end]);
```
calbach: Are these guaranteed to already be in a reasonable sorted order?

ui/src/app/job-details/tabs/tabs.component.html
```diff
@@ -128,13 +128,9 @@
     <mat-tab *ngIf="hasOutputs()" label="Outputs">
       <jm-resources-table [entries]="job.outputs"></jm-resources-table>
     </mat-tab>
-    <!-- Disable the graph view tab so the indicator doesn't switch on click -->
-    <mat-tab *ngIf="hasTimingUrl()" label="Graph View" [disabled]=true>
-      <ng-template mat-tab-label>
-        <a style="all: unset" href="{{ job.extensions.timingUrl }}" target="_blank">
```
cjllanwarne: Can we now remove this extension from the Cromwell shim code?

rsasch: Good point – removed.

ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
rsasch: I believe https://github.com/DataBiosphere/job-manager/blob/066fc41b6a4e885c058fbd5da42688d5d6662caf/servers/cromwell/jobs/controllers/jobs_controller.py#L112 sorts the tasks by start time.

PR #560, opened 2019-02-07
added back logging import and implementation (because it shouldn't have been removed)

rsasch:


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -21,11 +21,14 @@
...
 import urllib
+import logging
 
 _DEFAULT_PAGE_SIZE = 64
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('{module_path}'.format(module_path=__name__))
+
 
-@requires_auth
```
aednichols: Is removing auth from abort intentional?

PR #563, opened 2019-02-11
made amount of jobs loaded in the UI more flexible

### LINKED ISSUE #496: Pagination lag at 100 jobs/page view
Adapted from slack conversation: 

I am running into a pagination bug (?) where I am on the home page (i.e. no filters applied) viewing 100 jobs and it has 26k total jobs I could paginate through and the arrow to go to page 2 is greyed like I can't use it. If I change the view to 50 jobs it will appear, but only let me go up to 100 jobs and then it's grey again. If I leave it on page 2 and go back to size 100, it will let me click the paginate button once and show jobs 101-200, but then goes gray again. It's very strange.

It turns out, the paginate button goes ungrey eventually, it seems like it's trying to preload the next 100 jobs but there is a lag. The lag is less noticeable at 10 jobs per page, noticeable at 50 and 100.

From @rsasch :  it might have something to do with the fact that `minBackendPageSize` (see https://github.com/DataBiosphere/job-manager/blob/master/ui/src/app/shared/job-stream.ts#L10) is set to 100

rsasch:
Increased `initialBackendPageSize` and used it instead of separately set values in UI.

Closes #496

### DIFF
--- a/ui/src/app/job-list/job-list-resolver.service.ts
+++ b/ui/src/app/job-list/job-list-resolver.service.ts
@@ -43,9 +43,8 @@ export class JobListResolver implements Resolve<JobStream> {
     }
 
     const jobStream = new JobStream(this.jobManagerService, jobsRequest);
-    const numToLoad = settings.pageSize < initialBackendPageSize ? initialBackendPageSize : settings.pageSize * 2;
     return jobStream
-        .loadAtLeast(numToLoad)
+        .loadAtLeast(initialBackendPageSize)
         .then(resp => jobStream);
   }
 }
--- a/ui/src/app/job-list/job-list-resolver.service.ts
+++ b/ui/src/app/job-list/job-list-resolver.service.ts
@@ -43,8 +43,9 @@ export class JobListResolver implements Resolve<JobStream> {
     }
 
     const jobStream = new JobStream(this.jobManagerService, jobsRequest);
+    const numToLoad = settings.pageSize < initialBackendPageSize ? initialBackendPageSize : settings.pageSize * 2;
     return jobStream
-        .loadAtLeast(initialBackendPageSize)
+        .loadAtLeast(numToLoad)
         .then(resp => jobStream);
   }
 }
--- a/ui/src/app/job-list/job-list.component.ts
+++ b/ui/src/app/job-list/job-list.component.ts
@@ -124,9 +124,7 @@ export class JobListComponent implements OnInit {
 
     this.streamSubscription.unsubscribe();
     const nextStream = new JobStream(this.jobManagerService, req);
-    const pageSize = this.settingsService.getSavedSettingValue('pageSize', this.projectId);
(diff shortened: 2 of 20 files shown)


ui/src/app/job-list/job-list.component.ts
```diff
@@ -124,7 +124,9 @@ export class JobListComponent implements OnInit {
 
     this.streamSubscription.unsubscribe();
     const nextStream = new JobStream(this.jobManagerService, req);
-    nextStream.loadAtLeast(initialBackendPageSize)
+    const pageSize = this.settingsService.getSavedSettingValue('pageSize', this.projectId);
```
calbach: Getting it from `this.pageSize` is probably simpler

ui/src/app/shared/job-stream.ts
```diff
@@ -3,12 +3,11 @@ import {BehaviorSubject} from 'rxjs/BehaviorSubject';
 import {JobManagerService} from '../core/job-manager.service';
 import {QueryJobsResponse, QueryJobsResult} from './model/models';
 import {QueryJobsRequest} from "./model/QueryJobsRequest";
+import {initialBackendPageSize} from "./common";
 
 // An observable stream of the client's materialized jobs, where each update
 // contains all jobs that have been loaded so far.
 export class JobStream extends BehaviorSubject<JobListView> {
-  private static readonly minBackendPageSize = 100;
```
calbach: These are conceptually different, I think it would be better to keep the names separate, even if you just set the value here like `minBackendPageSize = initialBackendPageSize`. One is the initial request size, the other is the minimum total request size (first request or otherwise).

ui/src/app/job-list/job-list-resolver.service.ts
```diff
@@ -43,8 +43,9 @@ export class JobListResolver implements Resolve<JobStream> {
     }
 
     const jobStream = new JobStream(this.jobManagerService, jobsRequest);
+    const numToLoad = settings.pageSize < initialBackendPageSize ? initialBackendPageSize : settings.pageSize * 2;
```
calbach: I don't totally follow this logic, but I'd just take `max(initialBackendPageSize, settings.pageSize * 2)`

ui/src/app/job-list/job-list.component.ts
```diff
@@ -124,7 +124,9 @@ export class JobListComponent implements OnInit {
...
+    const numToLoad = initialBackendPageSize > pageSize ? initialBackendPageSize : pageSize * 2;
```
calbach: Same max comment as before.

rsasch:
@calbach Do you think simply increasing `initialBackendPageSize` (and similar) values currently set to `100` to be `200` would be enough to address the issue?  Or is there value to checking the `pageSize` setting?

calbach:
On the initial request, I don't think there's much value in considering the page size, as the user can just quickly adjust it and paginate - inducing the bug (based on my understanding). On subsequent requests, there may be some small value in considering the page size, but I don't know whether it's worth the added complexity.

PR #564, opened 2019-02-19
Add events to timing diagram

### LINKED ISSUE #562: Add events to new timing diagram
Within the Cromwell metadata, there are `executionEvents` within tasks, which could provide more information to the user if included within the timing diagram.

rsasch:
Adds granularity to timing diagram for tasks/subworkflows in Cromwell.

Closes #562

### DIFF
--- a/ui/src/app/job-details/job-details.component.ts
+++ b/ui/src/app/job-details/job-details.component.ts
@@ -26,8 +26,8 @@ export class JobDetailsComponent implements OnInit {
   }
 
   hasTabs(): boolean {
-    if (this.job.inputs || this.job.outputs) {
-      return true;
+    if (( this.job.inputs && Object.keys(this.job.inputs).length !== 0) || (this.job.outputs  && Object.keys(this.job.outputs).length !== 0)) {
+       return true;
     }
     if (this.job.extensions) {
       let tasks: TaskMetadata[] = this.job.extensions.tasks || [];
--- a/ui/src/app/job-details/tabs/tabs.component.ts
+++ b/ui/src/app/job-details/tabs/tabs.component.ts
@@ -34,7 +34,7 @@ export class JobTabsComponent implements OnInit, OnChanges {
   @Output() navDown: EventEmitter<string> = new EventEmitter();
   @ViewChild(JobFailuresTableComponent) failuresTable: JobFailuresTableComponent;
   @ViewChild(JobTimingDiagramComponent) timingDiagram: JobTimingDiagramComponent;
-
+  @ViewChild('tabsPanel') tabsPanel;
 
   database = new TasksDatabase(this.tasks);
   dataSource: TasksDataSource | null;
@@ -46,6 +46,7 @@ export class JobTabsComponent implements OnInit, OnChanges {
     'attempts',
     'files',
   ];
+  tabWidth: number = 1024;
 
   constructor(private authService: AuthService) {};
 
@@ -53,6 +54,9 @@ export class JobTabsComponent implements OnInit, OnChanges {
     this.dataSource = new TasksDataSource(this.database);
     if (this.hasCallCachedTask() || this.hasScatteredTask()) {
       this.displayedColumns.splice(1, 0, "taskInfoIcons");
+    }
+    if (this.tabsPanel) {
+      this.tabWidth = this.tabsPanel._viewContainerRef.element.nativeElement.clientWidth;
     }
   }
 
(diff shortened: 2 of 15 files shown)


ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
```diff
@@ -20,23 +21,42 @@ export class JobTimingDiagramComponent implements OnInit {
...
+
     this.timelineChart.options = {
-      title: 'Tasks',
-      width: 800,
-      height: counter * 50
+      timeline: {
+        showRowLabels: false,
+        showBarLabels: false
+      },
+      width: 1470,
```
calbach: Where did this value come from? Please add a comment, e.g. if it's matching something in CSS somewhere.

rsasch: I've added in some logic to find the browser width and calculate the graph width based on that.

ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
```diff
@@ -20,23 +21,42 @@ export class JobTimingDiagramComponent implements OnInit {
...
+
     this.timelineChart.options = {
-      title: 'Tasks',
-      width: 800,
-      height: counter * 50
+      timeline: {
+        showRowLabels: false,
+        showBarLabels: false
+      },
+      width: this.tabWidth - 65,
```
calbach: Looks like this only runs on `ngOnInit` so I'm not sure I'd expect this to resize itself properly. Perhaps a percentage based width would work better (if supported?).

Anyways - I can't really test this out, so I'll leave it to you or another reviewer to ensure this is functional

ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
```diff
@@ -20,23 +21,42 @@ export class JobTimingDiagramComponent implements OnInit {
...
+      height: (counter * 42) + 100,
```
calbach: nit: "counter" is pretty generic - I'd name it something more semantically relevant, such as `numRows`, or whatever it's actually referring to.

ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
rsasch: It isn't set to re-size, just to figure out how wide to make it when the component is first rendered.

ui/src/app/job-details/tabs/timing-diagram/timing-diagram.component.ts
rsasch: good point – changed

ui/src/app/job-details/job-details.component.ts
```diff
@@ -26,8 +26,8 @@ export class JobDetailsComponent implements OnInit {
   }
 
   hasTabs(): boolean {
-    if (this.job.inputs || this.job.outputs) {
-      return true;
+    if (( this.job.inputs && Object.keys(this.job.inputs).length !== 0) || (this.job.outputs  && Object.keys(this.job.outputs).length !== 0)) {
```
cjllanwarne: Is this change related?

ui/src/app/job-details/tabs/tabs.component.ts
```diff
@@ -54,6 +55,9 @@ export class JobTabsComponent implements OnInit, OnChanges {
     if (this.hasCallCachedTask() || this.hasScatteredTask()) {
       this.displayedColumns.splice(1, 0, "taskInfoIcons");
     }
+    if (this.tabsPanel) {
+      this.tabWidth = this.tabsPanel._viewContainerRef.element.nativeElement.clientWidth;
```
rexwangcc: Does this add the responsiveness to the timing panel?

rsasch: It allows the timing diagram component to determine the width to be initially rendered at.  It does not change if the size of the browser window is changed without a reload.

ui/src/app/job-details/job-details.component.ts
rsasch: tangentially – this function will determine if the tab group, which the timing diagram is in, will be displayed

ui/src/app/job-details/tabs/tabs.component.ts
rsasch: no, it's just passed along to the function that builds the timing diagram when the component is loaded

PR #568, opened 2019-02-21
Handle auth failure more elegantly

### LINKED ISSUE #441: Better experience when auth is configured but unnecessary
Pity the fool who might:

- Have Cromwell running with no auth
- Accidentally leave their capabilities.conf expecting auth:
```yaml
  "authentication": {
    "isRequired": true,
    "scopes": [
      "openid",
      "email",
      "profile"
    ]
  }
```
- On loading the landing page, they would see "Loading...", but no other sign that they did something wrong.

Maybe "redirecting to the authentication site" would make it more obvious what they did wrong?
Maybe forwarding errors from the shim to a UI popup would at least aid in debugging?

rsasch:
Now, when user has an `authorization` section in their capabilities config (or something else that causes `gapi.auth2` to fail, the error is caught and displayed to the user:

screen shot 2019-02-21 at 10 29 16 am

instead of just failing and dumping the error message in the javascript console.

Closes #441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Job Manager Change Log
 
+## v0.5.7 Release Notes
+
+### Added events to timing diagram
+
+Timing diagram now includes events within each task/subworkflow and the rendered width is no longer hard-coded.
+
+### Added better error handling to authentication
+
 ## v0.5.6 Release Notes
 
 ### Embedded timing diagram within UI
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -6,10 +6,10 @@
 
 Timing diagram now includes events within each task/subworkflow and the rendered width is no longer hard-coded.
 
-### Added better error handling to authentication
-
 ### Increased minimum number of jobs loaded in the UI to better handle page size being set to 100.
 
+### Added better error handling to authentication
```
cjllanwarne: I'd be more specific here and say "in cases where the API is trying to use oauth but the UI is not using CAAS"... or equivalent

PR #572, opened 2019-02-27
Throw 401 errors from Cromwell API, catch them in the UI

### LINKED ISSUE #570: Gracefully handle stale or revoked access tokens
Both @gdlex4015 and I have had situations where signing in to Job Manager for the first time, when already logged in to a Google account, crashes to the "Job Manager couldn't talk to its workflow server" page.

In all cases things worked fine after clearing cookies or using private browsing.

rsasch:
Closes #570

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -458,7 +458,7 @@ def _get_response_massage(response):
 
 def is_jsonable(x):
     try:
-        json.dumps(x)
+        x.json()
         return True
     except:
         return False
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -1,9 +1,12 @@
 import requests
 from flask import current_app
-from werkzeug.exceptions import BadRequest, NotFound, InternalServerError, ServiceUnavailable
+from werkzeug.exceptions import Unauthorized, BadRequest, NotFound, InternalServerError, ServiceUnavailable
 from datetime import datetime
 from dateutil.tz import *
 import dateutil.parser
+import json
+import logging
+import urllib
 
 from jm_utils import page_tokens
 from jobs.controllers.utils.auth import requires_auth
@@ -21,8 +24,6 @@ from jobs.models.health_response import HealthResponse
 from jobs.models.execution_event import ExecutionEvent
 from jobs.controllers.utils import job_statuses
 from jobs.controllers.utils import task_statuses
-import urllib
-import logging
 
 _DEFAULT_PAGE_SIZE = 64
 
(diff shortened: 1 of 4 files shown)


aednichols:
Not sure what happened there, github made it look like I had never approved

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -70,11 +71,11 @@ def update_job_labels(id, body, **kwargs):
...
 
     if response.status_code == InternalServerError.code:
-        raise InternalServerError(response.json().get('message'))
+        raise InternalServerError(_get_response_massage(response))
     elif response.status_code == BadRequest.code:
-        raise BadRequest(response.json().get('message'))
+        raise BadRequest(_get_response_massage(response))
     elif response.status_code == NotFound.code:
-        raise NotFound(response.json().get('message'))
+        raise NotFound(_get_response_massage(response))
```
cjllanwarne: I'm not 100% sure this would be the right time to get a massage...

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -445,3 +448,17 @@ def _get_scattered_task_status(metadata):
     ]:
         if status in statuses:
             return status
+
+
+def _get_response_massage(response):
```
cjllanwarne: this feels like these two could be combined? eg:
```python
def _get_response_massage(response):
    try:
        message = x.json().get('message')
        if message:
                return message
        else:
                return str(response)
    except:
        return str(response)
```

ui/src/app/core/initial-error/initial-error.component.html
```diff
@@ -10,16 +10,7 @@
...
-  <p>
-    Please check the URL and workflow server configuration, or contact your support organization for assistance.
-  </p>
-  <p>
-    <a href="/">Click here to start over.</a>
-  </p>
-  <p>
-    <i>{{errorMessage}}</i>
-  </p>
+  <p>Job Manager is running but encountered a problem getting data from its workflow server.</p>
```
cjllanwarne: I kind of agree that a way to make this more specific for 401s might be nice.

eg I could see this page being sort of contradictory - eg "is it broken...? or should I log back in...? argh!!"

cjllanwarne:
Current 401 error page:

screen shot 2019-02-27 at 3 59 20 pm

ui/src/app/core/initial-error/initial-error.component.html
cjllanwarne: Ignore this - has now been changed.

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: good point – changed!

PR #576, opened 2019-03-01
Modify /query call when pointing at CromIAM

rsasch:


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -314,7 +314,7 @@ def query_jobs(body, **kwargs):
 
     response = requests.post(
         _get_base_url() + '/query',
-        json=cromwell_query_params(query, page, query_page_size),
+        json=cromwell_query_params(query, page, query_page_size, headers),
```
aednichols: It looks like we're sending `headers` both as headers and in the POST JSON - is this intended?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -356,7 +356,7 @@ def page_from_offset(offset, page_size):
     return 1 + (offset / page_size)
 
 
-def cromwell_query_params(query, page, page_size):
+def cromwell_query_params(query, page, page_size, auth_headers):
```
aednichols: Since we don't need them for anything else, it would be nice to evaluate `auth_headers is None` in the calling function and pass in just a boolean here, called `is_cromiam` or something

servers/cromwell/jobs/controllers/jobs_controller.py
cjllanwarne: the JSON uses `headers` as a parameter, not within the result

Issue #577, opened 2019-03-05
Switching users leaks previous user's workflows


aednichols:
Signing out and then back in as another user briefly displays the previous user's workflows on-screen while the `/query` request for the new user does its thing.

This is at the very least confusing and could be construed as a security-ish thing by auditors (though unlikely to impact anyone IRL).

Thanks @aweng98

rsasch:
I believe this was closed by #580.

PR #580, opened 2019-03-07
Force reload of the jobs page on sign out

cjllanwarne:
Addresses #577 by forcing a reload of the `/jobs` page while the session is logged out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Job Manager Change Log
 
+## v0.5.10 Release Notes
+
+### Force reload of /jobs after sign-out to clear out the previous user's list.
+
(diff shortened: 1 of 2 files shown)


aednichols:
Does this potentially make it harder to sign back in, because you're on an empty jobs page upon signout instead of a sign-in page?

cjllanwarne:
@aednichols luckily no, because the refreshed job page then conveniently redirects you to the sign in page.

cjllanwarne:
@rsasch thanks - I think I must have just gotten lucky the first time I tested this. It definitely did what you described when I tried the second time.

Ok, so, attempt 2: force a reload, rather than a redirect

PR #589, opened 2019-03-14
Modify swagger yaml regex for job id

### LINKED ISSUE #584: Filter by numeric-starting workflow ID fails
To reproduce:

 * Open the job list
 * Find a workflow with an ID starting with a number
* In the query bar start typing "workflow" to choose the option "Workflow ID"
* Enter the workflow ID with a number at the start

An error appears, eg:

Screen Shot 2019-03-11 at 7 18 33 PM

rsasch:
Closes #584

### DIFF
--- a/api/jobs.yaml
+++ b/api/jobs.yaml
@@ -398,7 +398,7 @@ definitions:
           matches to each of these labels.
       id:
         type: string
-        pattern: ^[a-zA-Z][a-zA-Z0-9_-]*$
+        pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]+$
         description: >
           Returns only jobs with the specified id.
       name:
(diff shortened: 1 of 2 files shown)


api/jobs.yaml
```diff
@@ -398,7 +398,7 @@ definitions:
           matches to each of these labels.
       id:
         type: string
-        pattern: ^[a-zA-Z][a-zA-Z0-9_-]*$
+        pattern: ^[a-zA-Z0-9][a-zA-Z0-9_-]+$
```
aednichols: I didn't think these IDs could include underscores (I do recognize that this predated your PR)

rexwangcc: Since this regex is used by both dsub and cromwell backend, do we want to have a reviewer from the dsub side to take a look?

calbach: The end result here should be a pattern that matches all possible dsub or cromwell IDs. I think this satisfies it (changing the * --> + should be fine since I'm pretty confident we have a minimum id length).

PR #594, opened 2019-03-15
Changed how scattered tasks' status, duration and attempts are calculated

### LINKED ISSUE #590: Get shard status from latest attempt
Currently, Job Manager shows an inaccurate status, duration and timing diagram for scattered tasks when a shard (or shards) has more than 1 attempt.

rsasch:
Closes #590

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -167,8 +167,6 @@ def health(**kwargs):
 def format_task(task_name, task_metadata):
     # check to see if task is scattered
     if task_metadata[-1].get('shardIndex') != -1:
-        logger.warning('scattered task response {}'.format(
-            format_scattered_task(task_name, task_metadata)))
         return format_scattered_task(task_name, task_metadata)
     latest_attempt = task_metadata[-1]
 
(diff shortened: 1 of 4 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -232,12 +232,13 @@ def format_scattered_task(task_name, task_metadata):
             else:
                 temp_status_collection[
                     status] = temp_status_collection[status] + 1
-        if minStart > _parse_datetime(shard.get('start')):
-            minStart = _parse_datetime(shard.get('start'))
-        if shard.get('executionStatus') not in ['Failed', 'Done']:
-            maxEnd = None
-        if maxEnd is not None and maxEnd < _parse_datetime(shard.get('end')):
-            maxEnd = _parse_datetime(shard.get('end'))
+            if minStart > _parse_datetime(shard.get('start')):
```
rexwangcc: Could you remind me why move this conditional logic a level deeper?

rsasch: it was moved under the `if current_shard != shard.get('shardIndex'):` check to only get the data from the latest attempts of each shard

rexwangcc: Thanks for the explanation, that makes sense!

PR #595, opened 2019-03-18
added includeKey to metadata to reduce size of response from Cromwell

### LINKED ISSUE #574: Cannot view the (sub)workflow details if the metadata is too long
Sometimes when the workflow has too many shards, the content of its metadata will get too long/large, and the link to the workflows in Job Manager cannot work. Example: https://job-manager.caas-prod.broadinstitute.org/jobs/b429ebd9-c2be-4d7a-acef-adb9defa8897 if you click the `analysis` subworkflow, with `inpect` on Chrome, you will see 500 error occurred.

rsasch:
Closes #574

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -88,8 +88,19 @@ def get_job(id, **kwargs):
 
     :rtype: JobMetadataResponse
     """
-    url = '{cromwell_url}/{id}/metadata'.format(
-        cromwell_url=_get_base_url(), id=id)
+
+    include_keys = [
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -88,8 +88,19 @@ def get_job(id, **kwargs):
 
     :rtype: JobMetadataResponse
     """
-    url = '{cromwell_url}/{id}/metadata'.format(
-        cromwell_url=_get_base_url(), id=id)
+
+    include_keys = [
```
rexwangcc: optional: how about making this a `set([])`or a tuple than a list?

rsasch: ok – made it into a tuple

PR #597, opened 2019-03-18
Add in more human-friendly way to narrow down by date to query builder

### LINKED ISSUE #440: Filter by timestamp
Cromwell API allows filtering by time as well as date. Enable this in the UI date/time picker. Probably needs UX? Not sure if the date picker design has this. 

AC:
Time picker allows filtering on both start and end time

rsasch:
The dropdown:
Screen Shot 2019-03-18 at 6 24 40 PM

The options:
Screen Shot 2019-03-18 at 6 24 54 PM

After the user selects an option:
Screen Shot 2019-03-18 at 6 56 19 PM

Closes #440

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Job Manager Change Log
 
+## v0.6.3 Release Notes
+
+### Added drop down menu to query builder for more intuitive filtering by date.
+
 ## v0.6.2 Release Notes
 
 ### Fixed bug where scattered tasks' status, duration, timing diagram and number of attempts were inaccurate.
--- a/ui/src/app/shared/filter-header/filter-header.component.css
+++ b/ui/src/app/shared/filter-header/filter-header.component.css
@@ -193,3 +193,20 @@ clr-tooltip-content {
   font-family: 'Montserrat', sans-serif;
   line-height: 1rem;
 }
+
+mat-form-field.submitted-select {
+  margin-left: 15px;
+  font-size: 0.75rem;
+}
+
+::ng-deep mat-form-field.submitted-select .mat-input-underline.mat-form-field-underline {
+  display: block;
+}
+
+::ng-deep mat-form-field.mat-form-field-type-mat-select.submitted-select .mat-form-field-infix.mat-input-infix {
+  padding: 5px;
+}
+
+::ng-deep mat-form-field.submitted-select .mat-select-arrow-wrapper .mat-select-arrow {
+  color: rgba(0, 0, 0, 0.54);
+}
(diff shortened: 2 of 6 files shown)


calbach:
Some UX feedback at a glance:
- What is the value proposition of this control on the list page? Results are already sorted newest -> oldest so it seems like you're mostly just going to be trimming out older jobs which are not shown on the current page anyways. Having a special UI control for this indicates we're attaching some high priority to this action. 
- Why remove the drop-down after selection?

rsasch:
@ruchim Could you speak to the value-add of the drop down?

@calbach Regarding your second question, the control disappears after the user makes a selection because that's also how the status buttons work; once the user has created a status or submitted chip, they can modify the value by clicking on the chip.

ruchim:
@calbach @rsasch

Why the dropdown?
The drop down bar because if a user is constantly adding a filter to their query, then the drop down is less typing and more intuitive. Also, they can change the other facets of their query and always count on that specific filter to present, but not cluttering the query builder. 

Why submitted date in the drop down?
In Terra, even having things sorted by newest ends up resulting in a users info buried a few pages deep (since their view isn’t subsetted by the workflows a user launched, but by everything that their collaborators have shared across all projects over all time)

calbach:
@ruchim I'm still not following - is there a bug here? Jobs should be sorted by submission time from newest to oldest. This means that limiting to jobs from the past N days should only lop jobs off the end of the list (e.g. reducing the total number of pages) - it would not bring any different jobs to the first page.

ruchim:
Maybe I don’t understand your question. You’re saying that if someone has a relevant workflow that was submitted past what fits on the first page, changing the submitted date range doesn’t help, as it still requires pagination?

calbach:
Correct, unless you filter by an end time range (which is not what this control is doing - nor should it be).

I think having this control + a 7d default + not disappearing would be pretty reasonable - I'm not sure that's the goal here though so I'm trying to understand the rationale for the change.

ruchim:
The main advantages of lopping things off is the difference in user experience around seeing a hundred thousand workflows returned vs a couple hundred. Psychologically, if the goal is to filter down on a delinquent run, it helps if the filter functionality returns less candidates, so that the search space is reduced. And in this case, it’s easily reduced without the expense of losing important info. 

Secondly, Cromwell returns the info for the list view much faster, as it has to scan for less information.

calbach:
Thanks, makes sense. In that case I would suggest keeping the dropdown as a persistent element (relatedly, not having it produce chips in the query builder), and either in this PR or a future PR applying a default age limit, e.g. 7 days.

rsasch:
Going to re-implement this based on the feedback and will create a new PR when ready to review.

PR #598, opened 2019-03-19
Improvements to parsing of error information in metadata

### LINKED ISSUE #588: Missing name and error details when there's an input type error
I submitted a job (https://job-manager.dsde-prod.broadinstitute.org/jobs/62d69037-b6a8-4d45-bdd0-df6aa8dda668?q=submission-id%3D87c83516-5b93-4e9f-9318-5a094f5a5d07) that shows no error details even though the job failed. On the list page, there's nothing in the name column. On the detail page, there's no indication of the cause of the error.

Here's the error I found by drilling down in FireCloud:

> Failed to evaluate input 'do_the_thing.status' (reason 1 of 1): No coercion defined from '0' of type 'spray.json.JsNumber' to 'String'.

rsasch:
- added handling of workflow errors that aren't embedded in the 'calls' section of the metadata
- small CSS tweaks to display of error info in job details

Closes #588

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -109,10 +109,14 @@ def get_job(id, **kwargs):
     job = response.json()
 
     failures = [
-        format_failure(name, m)
+        format_task_failure(name, m)
         for name, metadata in job.get('calls', {}).items() for m in metadata
         if m.get('failures') is not None
     ]
+
+    # if there are no tasks/subworkflows but there are errors, get them
+    if not len(failures) and job.get('failures'):
+        failures = [format_workflow_failure(f) for f in job.get('failures')]
 
     tasks = [
         format_task(task_name, task_metadata)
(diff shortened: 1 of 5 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -109,11 +109,15 @@ def get_job(id, **kwargs):
     job = response.json()
 
     failures = [
-        format_failure(name, m)
+        format_task_failure(name, m)
         for name, metadata in job.get('calls', {}).items() for m in metadata
         if m.get('failures') is not None
     ]
 
+    # if there are no tasks/subworkflows but there are errors, get them
```
rexwangcc: Not sure if this comment is describing what the line below is doing 🤔

rsasch: How so?  `not len(failures)` tests if the previous attempt to gather failures from `calls` worked and `job.get('failures')` tests if there are failures elsewhere in the metadata.  Suggestions for making this clearer?

rexwangcc: Ah, I got it now. I somehow misunderstand the `if there are no tasks/subworkflows` just now. It's pretty clear!

PR #608, opened 2019-03-26
Move all output icons into one component and re-use

### LINKED ISSUE #602: Fix stderr/stdout paths
Today, when using Job Manager through Terra and clicking on the stdout/stderr links -- the redirect brings one to the correct file path -- but the file is treated like a directory, as the error says "There are no live objects in this folder".

Screen Shot 2019-03-22 at 9 59 22 AM

And when I traverse the directory structure one level up, I see the actual execution directory, which includes a link to the stdout/stderr files.
Screen Shot 2019-03-22 at 10 05 03 AM

AC: 
1. Correct the links to the stderr/stdout files as objects, instead of folders.
2. Ensure the redirect links open the new URLs in a new tab, instead of the existing one.
3. Add tests to make sure these behaviors (#1 and #2) don't regress in the future, as these links are crucial to debugging.

### LINKED ISSUE #609: missing auth user from execution dir link
when looking at a workflow, in the log files column, there are links to stdout, stderr and execution directory. The links for stdout and stderr include the `authuser` parameter but execution directory does not. This means that if the user is signed in with more than one user they may see an access denied screen.

rsasch:
Reduce possibility of certain icons behaving/linking in a different way dependent on context.

Closes #602 
Closes #609

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed bug where failure message(s) are not displayed if the job failed before Cromwell was able to run it (most likely due to a validation error).
 
+### Made behavior of message, log and execution icons consistent across all contexts in the job details page.
+
 ## v0.6.2 Release Notes
 
 ### Fixed bug where scattered tasks' status, duration, timing diagram and number of attempts were inaccurate.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Made behavior of message, log and execution icons consistent across all contexts in the job details page.
 
+Also fixed bug where log file links were being treated like directories.
+
 ## v0.6.2 Release Notes
 
 ### Fixed bug where scattered tasks' status, duration, timing diagram and number of attempts were inaccurate.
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Remove 'loading' screen when user has made an invalid query on job list page so the user can make changes.
 
+### Made behavior of message, log and execution icons consistent across all contexts in the job details page.
+
+Also fixed bug where log file links were being treated like directories.
```
aednichols: It may be more helpful to describe the user impact of this fix - "no more 404s" - instead of the underlying cause

ui/src/app/job-details/common/debug-icons/debug-icons.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  constructor(private authService: AuthService,
+              private sanitizer:DomSanitizer) {
+  }
+
+  ngOnInit() {
+    // mark URLs as safe
+  }
+
+  getResourceUrl(url: string): SafeUrl {
```
aednichols: Is the `SafeUrl` return type using the type system to indicate that the URL has gone through the sanitizer? Very cool!

CHANGELOG.md
rsasch: Changed to hopefully be clearer.

PR #610, opened 2019-03-27
add authuser to task dir link

dvoet:
https://github.com/DataBiosphere/job-manager/issues/609

this needs testing


rsasch:
Will review after #608 is merged, which centralizes where `getTaskDirectory` gets defined.

PR #618, opened 2019-04-05
Fix duration bug for scattered tasks

rsasch:
Undo bug introduced in previous feature (added execution events to shards in scattered tasks) that removed needed `end` time from scattered tasks.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -231,12 +231,11 @@ def format_scattered_task(task_name, task_metadata):
     filtered_shards = []
     current_shard = ''
     min_start = _parse_datetime(task_metadata[0].get('start'))
+    max_end = _parse_datetime(task_metadata[-1].get('end'))
 
     # go through calls in reverse to grab the latest attempt if there are multiple
     for shard in task_metadata[::-1]:
         if current_shard != shard.get('shardIndex'):
-            if min_start > _parse_datetime(shard.get('start')):
-                min_start = _parse_datetime(shard.get('start'))
             filtered_shards.append(
                 TaskShard(
                     execution_status=task_statuses.cromwell_execution_to_api(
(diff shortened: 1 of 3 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -231,12 +231,11 @@ def format_scattered_task(task_name, task_metadata):
     filtered_shards = []
     current_shard = ''
     min_start = _parse_datetime(task_metadata[0].get('start'))
+    max_end = _parse_datetime(task_metadata[-1].get('end'))
 
     # go through calls in reverse to grab the latest attempt if there are multiple
     for shard in task_metadata[::-1]:
```
aednichols: [ToL] This is well beyond the scope of this PR but whenever I see a construct like
```
biggest_so_far = 0
for thing in stuff:
  if some_function(thing) > biggest_so_far:
    biggest_so_far = some_function(thing)
```
I wonder if we could eliminate state by doing something like
```
biggest = max(map(some_function, stuff))
```

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -246,6 +245,13 @@ def format_scattered_task(task_name, task_metadata):
                     shard_index=shard.get('shardIndex'),
                     execution_events=_get_execution_events(
                         shard.get('executionEvents'))))
+            if min_start > _parse_datetime(shard.get('start')):
+                min_start = _parse_datetime(shard.get('start'))
+            if shard.get('executionStatus') not in ['Failed', 'Done']:
```
aednichols: [ToL] String literals used as constants in correctness-critical places makes me a bit uneasy... would it make sense to centralize these in an enum?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -246,6 +245,13 @@ def format_scattered_task(task_name, task_metadata):
...
+                max_end = None
+            if max_end is not None and max_end < _parse_datetime(
```
aednichols: [ToL] I am notoriously bad at reasoning about double negatives - is there a `is_defined` construct of some kind we could employ instead?

PR #631, opened 2019-04-25
fixed incorrect map of Cromwell status to UI status

### LINKED ISSUE #629: Cromwell status is incorrectly translated
The Cromwell status `RetryableFailure` is incorrectly mapped to the Job Manager status `Running`; it should be `Failed`, instead.

rsasch:
Closes #629


servers/cromwell/jobs/controllers/utils/task_statuses.py
```diff
@@ -33,7 +33,7 @@
     CromwellExecutionStatus.UNSTARTABLE: ApiStatus.FAILED,
     CromwellExecutionStatus.ABORTED: ApiStatus.ABORTED,
     CromwellExecutionStatus.BYPASSED: ApiStatus.SUBMITTED,
-    CromwellExecutionStatus.RETRYABLEFAILURE: ApiStatus.RUNNING,
```
rexwangcc: If I understand it correctly, `RETRYABLEFAILURE ` is a Cromwell internal TEMP status indicating that the workflow/task failed due to a failure that's retry-able and will be retried soon. If so, marking it as `FAILED` might scare the users.

rsasch: `RetryableFailure` is not temporary in the sense that it will appear in the metadata of a job that is no longer running.  What status do you think would be more appropriate?

rexwangcc: > RetryableFailure is not temporary in the sense that it will appear in the metadata of a job that is no longer running.
If that's the case, i think marking it as `Failed` makes sense.

PR #645, opened 2019-05-09
added Terra-like styling to Job Details page tabs

### LINKED ISSUE #640: Implement Terra-like styles for Job Details page tabs
Screen Shot 2019-05-03 at 5 16 48 PM

In user testing, it seems like the tabs are sometimes missed, so adding more prominent (and familiar, if they're coming from Terra) styles could help.

rsasch:
Before:
Screen Shot 2019-05-09 at 5 09 19 PM

After:
Screen Shot 2019-05-09 at 5 08 57 PM

Closes #640

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## v0.7.3 Release Notes
 
+### Added styling to Job Details page tabs to add clarity.
+
 ### Elevated sub-workflow errors to the parent workflow error card on the Job Details page.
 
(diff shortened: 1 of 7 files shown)


rsasch:
@kshakir Re: your points

- The Terra styles are being tweaked, and I was going based on the styles that are coming (see the originating ticket, #640).
- I added a hover – thanks for noticing!
- That animation is an Angular thing.  I couldn't find a simple way to turn that off.

PR #651, opened 2019-05-20
Specify latest node-sass in package.json

### LINKED ISSUE #650: Local dev apparently broken due to dep issue
Steps to reproduce:
1. Follow the instructions in the document _Job Manager Quick-Start for Cromwell Devs_
2. Job Manager fails to start with `Error: Cannot find module 'node-sass'`

Doing a `docker-compose build` fails with the following error
```
Downloading binary from https://github.com/sass/node-sass/releases/download/v4.7.2/darwin-x64-72_binding.node
Cannot download "https://github.com/sass/node-sass/releases/download/v4.7.2/darwin-x64-72_binding.node": 

HTTP error 404 Not Found
```
Version 4.7.2 does exist and the file `darwin-x64-72_binding.node` exists on more recent releases - but not on the one we're requesting.

Figure out what is going on with this dependency and get local dev going again.

aednichols:
Closes #650 

I totally muddled my way through this, having never used `npm` before.

The A/C of local dev working again appears to be met.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -10,10 +10,10 @@
       "integrity": "sha512-BAYCVZ10ro6mgZQDZiNiVbX8ppygw4q7z/stpwG8WjMswgMRIcxsxYoC1VFuWcUPAf4UyfTIav6e8UZWA5+xnQ==",
       "dev": true,
       "requires": {
-        "loader-utils": "1.1.0",
-        "source-map": "0.5.7",
-        "typescript": "2.6.2",
-        "webpack-sources": "1.1.0"
+        "loader-utils": "^1.1.0",
+        "source-map": "^0.5.6",
+        "typescript": "~2.6.2",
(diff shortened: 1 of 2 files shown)


aednichols:
I'm a bit worried about what's going on in `package-lock.json`.

For example, we went from `"tslib": "1.9.0"` to `"tslib": "^1.7.1"` for no apparent reason.

I get that `^` means, "automatically take the latest major version" but it's confusing we would ever go backwards ( where `1.2.3` is `release.major.minor` )

cjllanwarne:
I assume it means "take the latest you can find, but definitely no earlier than x"?

aednichols:
There's more to it than that though, because previously we did not have the `^` and would not automatically accept new versions.

I feel that "automatically accept new versions" should be a conscious decision, not an inadvertent side effect as it seems to be here.

PR #652, opened 2019-05-21
Add in automatic sign out after an interval of inactivity

### LINKED ISSUE #623: Automatic Sign out
From the perspective of clinically compliant software, Job Manager needs to be able to log users out of the application along with their Google account after 5 minutes of idle activity.

rsasch:
Auto-sign-out will be enabled if the `capabilities_config.json` contains the following in the `authentication` section:

>     "forcedLogoutDomains": [
>       <one or more domain>
>     ],
>     "forcedLogoutTime": <number of milliseconds of inactivity that will result in sign-out>

Closes #623

### DIFF
--- a/servers/cromwell/README.md
+++ b/servers/cromwell/README.md
@@ -119,8 +119,8 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
         - If the field is `editable`, then `fieldType` is required.
         - If the field is `editable`, then `filterable` will be ignored.
 
-- (Required, CaaS only) Configure fields to display
-  - **Note:** If you want to use use Job Manager against Cromwell-as-a-Service, which is using SAM/Google OAuth for authZ/authN, the `capabilities_config.json` must also include some extra fields, as well as proper scopes, which are shown as below:
+- (Required, CromIAM only) Configure fields to display
+  - **Note:** If you want to use use Job Manager against CromIAM, which is using SAM/Google OAuth for authZ/authN, the `capabilities_config.json` must also include some extra fields, as well as proper scopes, which are shown as below:
 ```json
 {
   "displayFields": [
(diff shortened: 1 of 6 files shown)


ui/src/app/app.component.ts
```diff
@@ -14,7 +14,6 @@ export class AppComponent {
     private readonly authService: AuthService
   ) {}
 
-
```
kshakir: Revert? I'm generally not a fan of adding whitespace-only changes to code commits.

rsasch: Thanks for catching that – reverted.

ui/src/app/core/auth.service.ts
```diff
@@ -50,6 +63,12 @@ export class AuthService {
       if (!capabilities.authentication || !capabilities.authentication.isRequired) {
         return;
       }
+
+      if (capabilities.authentication.forcedLogoutDomains && capabilities.authentication.forcedLogoutTime && capabilities.authentication.forcedLogoutTime > (this.WARNING_INTERVAL * 2)) {
```
aednichols: I'm not sure I understand the meaning of `WARNING_INTERVAL * 2`.

Is it saying that if the `forcedLogoutTime` is really short - only twice the warning interval - don't bother showing a warning?

ui/src/app/core/auth.service.ts
```diff
@@ -96,7 +115,47 @@ export class AuthService {
...
+    this.resetTimers();
+  }
+
+  public resetTimers(): void {
+    window.clearTimeout(this.logoutTimer);
+    window.clearTimeout(this.warningTimer);
+    this.snackBar.dismiss();
+
+    this.warningTimer = window.setTimeout(() => {
+      this.snackBar.open('You are about to be logged out due to inactivity');
```
aednichols: I think it would be great to tell the user that they have 10 seconds in this message - it's pretty short compared to the one minute or 30 seconds I've seen elsewhere

ui/src/app/core/auth.service.ts
rsasch: This isn't a test of whether or not to show a warning, it's whether or not to set up the variables that will be used to force a logout after an idle period (see line 47).

aednichols: Why would we skip setting the variables if `capabilities.authentication.forcedLogoutTime ≤ (this.WARNING_INTERVAL * 2)`

rsasch: This was to guard against someone configuring the timeout to be an unreasonably short time and having the UI just keep logging users out.  The only validation of `capabilities.authentication.forcedLogoutTime` that's in place is that it needs to be a valid `int`.

ui/src/app/core/auth.service.ts
```diff
@@ -96,7 +115,47 @@ export class AuthService {
     return auth2.signOut();
   }
 
+  private revokeToken(): Promise<any> {
```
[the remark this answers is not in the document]
rsasch: This was based on how `signOut()` was implemented.  I've now made `signIn()`, `signOut()` and `revokeToken()` use the same pattern.

ui/src/app/core/auth.service.ts
rsasch: I think having the static value of "10 seconds" will be confusing to users.  What if they only spot the message when they actually have only 5 seconds to go?  Unfortunately, I haven't found a way to easily update the message in the `SnackBar` component without closing it and opening it again, so I don't know how to show an actual countdown.

servers/cromwell/README.md
```diff
@@ -188,6 +188,81 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
...
+  "queryExtensions": [
+    "hideArchived"
+  ],
+  "authentication": {
+    "isRequired": true,
+    "scopes": [
+      "openid",
+      "email",
+      "profile"
+    ]
```
aednichols: Missing `,` breaks JSON

rsasch: whoops – thanks for catching that; fixed

PR #656, opened 2019-06-20
Gracefully handle tasks with no start datetime [BA-4978]

rsasch:
Make sure that a missing `start` time in a task will not result in weird or broken behavior by changing the fallback value from `now()` to `end`.

Closes https://broadworkbench.atlassian.net/browse/BA-4978.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -38,8 +38,7 @@ attempt_include_keys = ('attempt', 'callCaching:hit', 'callRoot', 'end',
 
 job_include_keys = attempt_include_keys + (
     'calls', 'description', 'executionEvents', 'labels', 'parentWorkflowId',
-    'returnCode', 'shardIndex', 'start', 'status', 'submission',
-    'subWorkflowId', 'workflowName')
+    'returnCode', 'status', 'submission', 'subWorkflowId', 'workflowName')
 
 
 @requires_auth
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -38,8 +38,7 @@
 
 job_include_keys = attempt_include_keys + (
     'calls', 'description', 'executionEvents', 'labels', 'parentWorkflowId',
-    'returnCode', 'shardIndex', 'start', 'status', 'submission',
-    'subWorkflowId', 'workflowName')
+    'returnCode', 'status', 'submission', 'subWorkflowId', 'workflowName')
```
mcovarr: this is just removing the `shardIndex` and `start` keys that were duplicated?

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -278,7 +277,7 @@ def format_task(task_name, task_metadata):
         execution_status=task_statuses.cromwell_execution_to_api(
             latest_attempt.get('executionStatus')),
         start=_parse_datetime(latest_attempt.get('start'))
-        or datetime.utcnow(),
+        or _parse_datetime(latest_attempt.get('end')) or datetime.utcnow(),
```
mcovarr: I'm not following how this addresses missing start times 😞

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: yup

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: if `start` is not defined, the old behavior would set it at `datetime.utcnow()`, which, if the task failed super quickly, would be after the `end` time, and that would cause issues.

mcovarr: Ah okay that makes sense, thanks. Maybe add that brief explanation as a comment?

rsasch: will do

PR #659, opened 2019-06-28
Add operation details [BA-5762]

rsasch:
Requires a CromIAM that has the `GET /api/workflows/{version}/{id}/backend/metadata/{backendId}`
endpoint and a `outsideAuth` set to `true` in the `authentication` section of `capabilities_config.json` file.

The icon:
Screen Shot 2019-06-28 at 4 36 51 PM

The dialog box that opens when you click it:
Screen Shot 2019-06-28 at 4 37 07 PM

Uses the [ngx-json-viewer](https://github.com/hivivo/ngx-json-viewer) javascript library.

Closes https://broadworkbench.atlassian.net/browse/BA-5762 (although does not include support for Dsub operation details)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -335,6 +348,11 @@ def format_scattered_task(task_name, task_metadata):
                 failure_messages = [
                     f.get('message') for f in shard.get('failures')
                 ]
+            operation_details = None
+            if _should_get_operation_details(shard):
```
cjllanwarne: If I'm understanding this right, you _always_ look up the operations details from Google (via rawls, via orchestration), even if the user never opens that modal?

Given that we're already battling to minimize (a) response times and (b) response sizes, would it be possible to make this operations fetching another endpoint in the shim, which is called if - but only if - the user tries to open the operations details panel?

servers/cromwell/jobs/test/test_auth_utils.py
```diff
@@ -15,7 +15,8 @@ def test_token_auth_returns_200(self, mock_request):
             'cromwell_url': self.base_url,
             'cromwell_user': '',
             'cromwell_password': '',
-            'use_caas': True
+            'use_caas': True,
+            'capabilities': {}
```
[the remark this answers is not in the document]
rsasch: see above – requires `outsideAuth` to be set to `true` in the authentication section of the `capabilities_config.json` file

CHANGELOG.md
```diff
@@ -10,7 +14,7 @@
 
 ## v1.2.0 Release Notes
 
-### Added the ability – with the addition of a specific scope to the capabilities config file – to see the contents of Google Storage log files within the UI.
+### Added the ability -- with the addition of a specific scope to the capabilities config file -- to see the contents of Google Storage log files within the UI.
```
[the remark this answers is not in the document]
rsasch: yup – wanted to avoid the use of non-ASCII characters

servers/cromwell/jobs/controllers/jobs_controller.py
rsasch: Does https://github.com/DataBiosphere/job-manager/pull/659#discussion_r299234778 address this?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Job Manager Change Log
 
+## v1.3.0 Release Notes
+
+### Added icon that, when clicked, will provide Google Pipelines operation details. (So far) only implemented for a Cromwell backend, and it requires `outsideAuth` to be set to `true` in the authentication section of the `capabilities_config.json` file.
+
 ## v1.2.3 Release Notes
```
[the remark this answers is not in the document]
rsasch: fixed

servers/dsub/jobs/controllers/jobs_controller.py
```diff
@@ -207,6 +207,13 @@ def get_task_attempts(id, task, **kwargs):
     raise NotImplementedError('Tasks not supported by dsub.')
 
 
+def get_operation_details():
```
aednichols: I have a mild preference for stub functions to explode rather than do nothing when called

servers/cromwell/jobs/controllers/jobs_controller.py
cjllanwarne: This one is more a general question - do we really want to make the call to rawls every time, even if the user never opens the modal? It seems likely to slow down the responsiveness of JM's workflow details page and increase the size of the response (and eg use more memory in the browser)

rsasch: I was reluctant to add the additional infrastructure to separate out this call without any evidence that it would have an effect on performance.  The only slowness I observed was after the job details page had been loaded and the JSON-formatting library was setting up particularly large contents of the operation details dialog (which would happen no matter what).

cjllanwarne: How many tasks did you test with? Those requests look like they'll happen in series, and once per task.

eg this is the set of network requests that will now be needed to load a workflow:
```
[Once per workflow]  shim -> CromIAM -> Cromwell -> CromIAM -> shim
# then...
[Once per task]      shim -> Rawls -> PAPIv2 -> Rawls -> shim
```

And those "once per task" requests will happen in series, so every task in a scatter is presumably adding another set of network requests and response wait times?

cjllanwarne: eg if that `[Once per task]` chain takes 100ms to complete, a workflow with a 100-wide scatter will have to make 100 of those request chains in series, and the time will presumably add up between them?

rsasch:
Need to re-work; 100 failures (and the associated operation details) causes request to time out.

PR #663, opened 2019-07-11
two small bug fixes [BA-5815]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/BA-5815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed a bug where clicking on operation details icon will show attempt data.
 
-### Fixed a bug in the handling of scattered task shards with no start date.
+### Fixed a bug that caused Job Manager to throw a 500 error when it attempted to process a scattered task shard with no start time.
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Job Manager Change Log
 
+## v1.3.1 Release Notes
+
+### Fixed a bug where clicking on operation details icon will show attempt data.
+
+### Fixed a bug in the handling of scattered task shards with no start date.
```
aednichols: I think it could be valuable to describe exactly what happens when the bug triggers so that affected users can see the notes and say, "YES!! they fixed MY bug."

```suggestion
### Fixed a bug that caused Job Manager to XXXX when it attempted to process a scattered task shard with no start time.
```

rsasch: Changed – is this better?

aednichols: Yep

PR #668, opened 2019-07-30
Added ability to link to/view backend log [BA-5507]

rsasch:
Screen Shot 2019-07-30 at 3 09 07 PM

Closes https://broadworkbench.atlassian.net/browse/BA-5507


ui/src/app/app.component.ts
```diff
@@ -1,19 +1,23 @@
...
 
   constructor(
     private readonly router: Router,
-    private readonly authService: AuthService
-  ) {}
+    private readonly authService: AuthService,
+    private customIconService: CustomIconService
+  ) { }
 
+  ngOnInit() {
```
aednichols: Deliberate no-op function?

If this function is required because it overrides a default behavior to specify a no-op behavior, it would be great to comment that so it doesn't get deleted by a later maintainer

rsasch: good catch – removed

PR #671, opened 2019-08-09
Use Google Storage XML API to get log contents [BA-5646]

rsasch:
- if configured, use SAM service to get pet account token with the right scope
- use pet account token to get the contents of a file in the execution directory
- use contents to show disabled icon if log doesn't exist or is empty
- if there are log contents, display in a dialog box when the icon is clicked on

Closes https://broadworkbench.atlassian.net/browse/BA-5646


servers/cromwell/README.md
```diff
@@ -264,6 +264,78 @@ Thin shim around [`cromwell`](https://github.com/broadinstitute/cromwell).
   - The `forcedLogoutDomains` setting is an array of user domains where this should apply.
   - The `forcedLogoutTime` is the amount of inactive time (in milliseconds) that will trigger an automatic sign-out.
 
+
+- (Optional, CromIAM and SAM) Configure fields to display
+  - **Note:** If you want to use use Job Manager against CromIAM and you also have access to a SAM server to handle authentication, the `capabilities_config.json` can be set up with the `outsideAuth` setting in the `authentication` section, which will allow job manager to get Google Pipelines operation details and the contents of log files stored in the job's execution directory:
```
aednichols:
```suggestion
  - **Note:** If you want to use use Job Manager against CromIAM and you also have access to a SAM server to handle authentication, the `capabilities_config.json` can be set up with the `outsideAuth` setting in the `authentication` section, which will allow Job Manager to get Google Pipelines operation details and the contents of log files stored in the job's execution directory:
```

kshakir: "...If you want to **use use** Job Manager..."

rsasch: it was actually in 3 places, so thanks for the catch!  fixed

PR #673, opened 2019-08-14
added a link to the top-level execution directory [BA-5788]

rsasch:
Screen Shot 2019-08-14 at 5 34 50 PM

Closes https://broadworkbench.atlassian.net/browse/BA-5788.

### DIFF
--- a/ui/src/app/job-details/panels/panels.component.ts
+++ b/ui/src/app/job-details/panels/panels.component.ts
@@ -164,7 +164,7 @@ export class JobPanelsComponent implements OnInit {
     if (this.job.extensions && this.job.extensions.tasks) {
       this.job.extensions.tasks.forEach((task) => {
         if (this.topLevelExecutionDirectory) return;
-        if (task.callRoot) {
+        if (task.callRoot && task.callRoot.includes(this.job.id)) {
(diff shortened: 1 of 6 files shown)


ui/src/app/job-details/panels/panels.component.ts
```diff
@@ -156,6 +160,20 @@ export class JobPanelsComponent implements OnInit {
     }, 1500);
   }
 
+  getTopLevelDirectory(): string {
+    if (this.job.extensions && this.job.extensions.tasks) {
+      this.job.extensions.tasks.forEach((task) => {
+        if (this.topLevelExecutionDirectory) return;
+        if (task.callRoot) {
+          const urlParts = task.callRoot.split(this.job.id);
+          this.topLevelExecutionDirectory = urlParts[0] + this.job.id;
```
kshakir: Is the `task.callRoot` guaranteed to contain the `this.job.id`? I only ask because there's an `if` statement above checking if the `task.callRoot` exists, and it seems to be assuming the job.id will therefore also be present in this non-null call root.

rsasch: added check for `job.id` within `callRoot`

PR #676, opened 2019-08-21
Display only the backend log (no more stdout and stderr) [BA-5907]

rsasch:
- removed stdout and stderr logs, since the content is contained in the backend log
 - increased max size of backend log to 500KB

Before:
Screen Shot 2019-08-21 at 12 21 14 PM

After:
Screen Shot 2019-08-21 at 12 17 27 PM

Closes https://broadworkbench.atlassian.net/browse/BA-5907

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -37,8 +37,7 @@ logger = logging.getLogger('{module_path}'.format(module_path=__name__))
 
 attempt_include_keys = ('attempt', 'backendLogs:log', 'callCaching:hit',
                         'callRoot', 'end', 'executionStatus', 'failures',
-                        'inputs', 'jobId', 'outputs', 'shardIndex', 'start',
-                        'stderr', 'stdout')
+                        'inputs', 'jobId', 'outputs', 'shardIndex', 'start')
 
 job_include_keys = attempt_include_keys + (
     'calls', 'description', 'executionEvents', 'labels', 'parentWorkflowId',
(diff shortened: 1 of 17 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -607,15 +600,15 @@ def tail_file_contents(bucket, object, **kwargs):
     headers = {
         'Content-Type': 'text/xml',
         'Authorization': 'Bearer ' + token,
-        'Range': 'bytes=-20000'
+        'Range': 'bytes=-500000'
```
cjllanwarne: That's quite a bump... I assume we aren't too concerned about security or money since this is a manual and opt-in thing to be doing?

servers/cromwell/jobs/test/test_jobs_controller.py
```diff
@@ -89,8 +89,7 @@ def test_update_job_labels_returns_200(self, mock_request):
             'test.analysis.outputs': 'gs://project-bucket/test/outputs.txt'
         }
         labels = {}
-        std_err = '/cromwell/cromwell-executions/id/call-analysis/stderr'
-        std_out = '/cromwell/cromwell-executions/id/call-analysis/stdout'
+        backend_log = '/cromwell/cromwell-executions/id/call-analysis/call-analysis-log'
```
cjllanwarne: I'm surprised that `backend_log` has to be explicitly added in here since (presumably) it was already being populated before (hence why we no longer need the `stdout` and `stderr` fields)?

PR #677, opened 2019-08-21
Don't treat preempted attempts as failures [BA-5910]

rsasch:
Closes https://broadworkbench.atlassian.net/browse/BA-5910


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -121,7 +121,8 @@ def get_job(id, **kwargs):
     failures = [
         format_task_failure(name, m)
         for name, metadata in job.get('calls', {}).items() for m in metadata
-        if m.get('failures') is not None
+        if m.get('failures') is not None and (
```
geoffjentry: This feels like a weird way to split the line but I'm not a pythonista anymore so 🤷‍♂

rsasch: yapf made me do it

PR #678, opened 2019-08-22
Add shard index to failures card and tab [BA-5860]

rsasch:
Before:
before

After:
after

Closes https://broadworkbench.atlassian.net/browse/BA-5860


rsasch:
> Looks good to me. Have you confirmed that the "no shard" case still works and looks appropriate (eg and we don't leak the awkward `shard -1` implementation detail into the UI?)

Screen Shot 2019-08-22 at 1 33 53 PM

PR #680, opened 2019-08-30
Switch from detox to tox BA-5968

kshakir:


.circleci/config.yml
```diff
@@ -68,10 +68,11 @@ jobs:
...
             /bin/bash ./scripts/rebuild_swagger.sh
       - run:
-          name: Install detox and yapf
-          # detox fails with newer versions of tox: https://github.com/tox-dev/detox/issues/33
+          name: Install tox and yapf
+          # tox (and detox) fail with older versions of more-itertools: https://github.com/pytest-dev/pytest/issues/4770
           command: |
-            pip install detox tox==3.5.0 yapf
+            pip install --upgrade pip
+            pip install tox>=3.13.2 pytest>=4.6.5 yapf
```
aednichols: Any reason not to use the latest tag, or latest minor version tag?

I have no specific knowledge that this is possible in `pip`, if it's impossible or annoying definitely ignore.

kshakir: Unlike `docker` tags, I wasn't able to find the concept of `latest` with `pip`. FYI, I did learn about `~=` versus `>=`:

- https://stackoverflow.com/questions/49854628/how-do-i-pip-install-the-latest-patch-number-of-a-package
- https://pip.pypa.io/en/stable/reference/pip_install/#example-requirements-file
- https://pip.pypa.io/en/stable/reference/pip_install/#satisfying-requirements

That said, I don't care if pip ends up "satisfying requirements" with an even newer version of `tox`/`pytest`. I just know older versions cryptically failed both on my laptop and CircleCI, and these new versions Worked For Me™.

Any preference on `>=` versus using `~=` to pin to a major.minor?

aednichols: Not really, it was just curiosity. Feel free to merge w/o any further changes

PR #688, opened 2019-11-25
fixed duration time for aborted scattered tasks [WA-79]

rsasch:
The Cromwell shim layer was not setting the end time, so duration was incorrectly calculated within the UI.

### DIFF
--- a/servers/cromwell/jobs/controllers/jobs_controller.py
+++ b/servers/cromwell/jobs/controllers/jobs_controller.py
@@ -376,7 +376,9 @@ def format_scattered_task(task_name, task_metadata):
             if shard.get('start') and min_start > _parse_datetime(
                     shard.get('start')):
                 min_start = _parse_datetime(shard.get('start'))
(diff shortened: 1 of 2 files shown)


servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -376,7 +376,9 @@ def format_scattered_task(task_name, task_metadata):
             if shard.get('start') and min_start > _parse_datetime(
                     shard.get('start')):
                 min_start = _parse_datetime(shard.get('start'))
-            if shard.get('executionStatus') not in ['Failed', 'Done']:
+            if shard.get('executionStatus') not in [
+                    'Failed', 'Done', 'Aborted'
```
kshakir: Don't know PEP, Is this two-indents (8-spaces) or a single indent (4-spaces)

rsasch: it's whatever the python linter Gods say it should be 🤷‍♀

PR #697, opened 2020-06-27
Update Dockerfile for nginx vuln

davidbernick:


davidbernick:
Thanks Khalid. Yeah I did some hardcore hacking updating maven repo urls from my phone.

PR #698, opened 2020-09-08
Fix an issue that messes up subworkflows in specific Cromwells

rexwangcc:
## Reason

In general, we are deploying JMUI against a Cromwell instance that uses OAuth (an OIDC based AuthProxy) but does not have CromIAM. This is affected by https://github.com/DataBiosphere/job-manager/pull/576 that subworkflows are showing up in the job details page.

See discussion here: https://broadinstitute.slack.com/archives/CFA6Q2QBU/p1599146973203800?thread_ts=1599145468.201200&cid=CFA6Q2QBU

## Changes

Add a `INCLUDE_SUBWORKFLOWS` flag and set it to `True` by default so subworkflows are filtered out by default as well. non-CromIAM Cromwell backends that are using OAuth should set it to `False`.

### DIFF
--- a/cromwell-caas-compose.yml
+++ b/cromwell-caas-compose.yml
@@ -22,6 +22,7 @@ services:
       - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1}
       - SAM_URL=${SAM_URL}
       - USE_CAAS=True
+      - IS_CROMIAM=True
   jobs-proxy:
     extends:
       file: common-compose.yml
--- a/cromwell-caas-compose.yml
+++ b/cromwell-caas-compose.yml
@@ -22,7 +22,7 @@ services:
       - CROMWELL_URL=${CROMWELL_URL:-https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1}
       - SAM_URL=${SAM_URL}
       - USE_CAAS=True
-      - IS_CROMIAM=True
+      - INCLUDE_SUBWORKFLOWS=True
   jobs-proxy:
     extends:
       file: common-compose.yml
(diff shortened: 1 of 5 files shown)


samanehsan (approved):
This looks great, thanks! :octocat:

rsasch (approved):
Nit: if the purpose of the new config option is **just** to not include sub-workflows in some cases, I think naming the option IS_CROMIAM is confusing (unless you anticipate other logic will need to change based on whether or not CromIAM is involved).

rexwangcc:
> Nit: if the purpose of the new config option is **just** to not include sub-workflows in some cases, I think naming the option IS_CROMIAM is confusing (unless you anticipate other logic will need to change based on whether or not CromIAM is involved).

@rsasch 
Thanks! That is a good point, at least for now the only purpose is to filter out sub-workflows by default. Do you think renaming it to "INCLUDE_SUBWORKFLOWS" and set it to `False` is safe? (safe means this won't break any Terra components/logic)

rsasch:
> Thanks! That is a good point, at least for now the only purpose is to filter out sub-workflows by default. Do you think renaming it to "INCLUDE_SUBWORKFLOWS" and set it to False is safe? (safe means this won't break any Terra components/logic)
That name works for me, but I think it should be `true` by default since Terra shows sub-workflows on the job details page.

rexwangcc:
> > Thanks! That is a good point, at least for now the only purpose is to filter out sub-workflows by default. Do you think renaming it to "INCLUDE_SUBWORKFLOWS" and set it to False is safe? (safe means this won't break any Terra components/logic)
> 
> That name works for me, but I think it should be `true` by default since Terra shows sub-workflows on the job details page.

Great, thanks! TIL that by default since Terra shows sub-workflows on the job details page, since we prefer the opposite in daily operations 😄

PR #703, opened 2021-01-11
[BT-85] [BT-78] Add AppSec GitHub Trivy Action

dinvlad:
This change adds Trivy security scan for the Docker image, as discussed in the Workbench Cross-Team meeting.

The action will run on every PR and give you feedback if the Docker image contains critical OS vulnerabilities.

More info:
https://github.com/broadinstitute/dsp-appsec-trivy-action
https://github.com/broadinstitute/dsp-appsec-blessed-images

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -18,4 +18,6 @@ jobs:
       - uses: actions/checkout@v2
 
       - uses: broadinstitute/dsp-appsec-trivy-action@v1
-        working-directory: ${{ matrix.workdir }}
+        with:
+          dockerfile: ${{ matrix.workdir }}/Dockerfile
+          context: ${{ matrix.workdir }}
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,21 @@
+name: dsp-appsec-trivy
+on: [pull_request]
+
+jobs:
+  appsec-trivy:
+    # Parse Dockerfile and build, scan image if a "blessed" base image is not used
+    name: DSP AppSec Trivy check
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        workdir:
+        - servers/cromwell
+        - servers/dsub
+        - ui
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: broadinstitute/dsp-appsec-trivy-action@v1
+        working-directory: ${{ matrix.workdir }}
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -19,4 +19,4 @@ jobs:
 
       - uses: broadinstitute/dsp-appsec-trivy-action@v1
         with:
-          context: ${{ matrix.workdir }}
+          dockerfile: ./${{ matrix.workdir }}/Dockerfile
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -19,5 +19,4 @@ jobs:
 
       - uses: broadinstitute/dsp-appsec-trivy-action@v1
         with:
-          dockerfile: ${{ matrix.workdir }}/Dockerfile
           context: ${{ matrix.workdir }}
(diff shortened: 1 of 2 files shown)


breilly2 (approved):
👍   Looks good to me. Thanks for adding.

@dinvlad What should we do about the fact that one of the checks is currently failing? It looks like we need to address it soon, but we probably don't want to delay merging the addition of the checks.

dinvlad:
Yup, as mentioned in BT-85, we do have a compliance requirement to fix any Criticals within 14 days of discovery.

Feel free to either fix it in this PR, or merge and create a separate ticket for fixing.

The one that's failing is due to reliance on `openjdk` image, which is generally less-well-maintained (by Oracle) than `adoptopenjdk` images. So if you'd like a quick fix, I'd suggest using one of those!

dinvlad:
As a shameless plug, AppSec also regularly builds and scans "blessed" base images, https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/jre - those live under `us.gcr.io/broad-dsp-gcr-public/base/` and they do use `adoptopenjdk` as base. If you decide to use those, the scans will be skipped completely!

breilly2:
I'm not sure it's openjdk in this case. The other Dockerfiles in this repo use openjdk as well, but they all also have multiple `FROM` lines. In the case of `ui`, the sequence is:
```
FROM openjdk
...
FROM node
...
FROM nginx:1.19.0
```
I think it might be the nginx image. Any suggestions for fixing that?

dinvlad:
Ah you're right, it's `nginx:1.19.0` as it's the last stage (Trivy only scans that). I've verified that the latest `1.19` (as opposed to `1.19.0`) is "clean". Could you use `1.19` perhaps (or `1.19.6`, although it will probably be outdated quickly as well..)?

dinvlad:
Update: it seems they patched `1.19.0` upstream (conveniently!), but I'd still suggest picking a less strict version like `1.19`, so it can stay ahead on the updates, while still being fairly stable..

breilly2:
For deployed production artifacts, I'm more inclined to lock it down to a specific version. Otherwise, we might not have visibility into which specific version is running. For example, if we don't build and deploy for a while, we won't necessarily be on the latest `1.19` image.

dinvlad:
LGTM!

mcovarr:
Is it okay that Trivy is only scanning the last `FROM` image in a multi-stage build? Couldn't the Dockerfile be copying insecure artifacts out of an earlier insecure `FROM` image? This `openjdk` is implicitly `latest` which would currently work out to be [version 15](https://github.com/docker-library/docs/blame/master/openjdk/README.md#L98) which is far removed from the version 8 that we [knew to be insecure](https://github.com/broadinstitute/cromwell/pull/6152/files#diff-1d3730a3eeb0ee66b0710a6adbec9d4651258406092f5239f5b2b2ae20869092), but shouldn't we be checking all the images we're using?

mcovarr (approved):
The concerns in my comments regard how Trivy scans Dockerfiles, no issues with how it's set up for this particular repo. 👍

dinvlad:
@mcovarr good point here - the reason for that is that we only worry about "OS-level" vulnerabilities here, and all the other runtime "artifacts" are scanned using a different service/mechanism.

kshakir (commented):
Letting the other two reviewers handle this PR

kshakir:
NOTE: This PR contains not just the BT-85 scan, but should also fix the issue that we believe prompted BT-78. Thanks @breilly2 for removing the detected vulnerability in this PR! 🥇

mcovarr:
@dinvlad yes "all the other images in the Dockerfile" was what I meant. Since it sounds like this is covered and you've got all the thumbs and passing checks you need, merge at will! 🙂

dinvlad:
Perfect, thanks all!

mcovarr:
Hi @dinvlad, can this be merged now? This closes a couple of tickets for us that we'd like to mark Done. 🙂  Thanks!

dinvlad:
yes, please do!

PR #730, opened 2021-07-30
Fix remaining moderate & low security issues [BW-774]

aednichols:
The more base images we can convert to [blessed](https://github.com/broadinstitute/dsp-appsec-blessed-images) ones, the better.


dinvlad (approved):

dinvlad:
LGTM for now! Btw it's not necessary to use blessed in all stages of a Dockerfile, we only care about the last stage (which is used at runtime).

aednichols:
> it's not necessary to use blessed in all stages of a Dockerfile

@dinvlad TIL, thanks. It can't hurt though, right? If someone compromises an image used in a non-final stage, that could still contaminate the end product, I would think.

dinvlad:
Yep, good point! Trivy is mostly a _runtime_ requirement on our side, since it reveals vulnerabilities that could _potentially_ be exploited once the service is _deployed_ in prod. It does not scan for "malicious" images afaik (although some have reported it detects things like cryptomining bots). In addition, we're only scanning for OS-level dependencies and omit others (e.g. Python/Java packages etc., for which we use SourceClear instead). So you can treat it as just a "compliance checkmark" :-)

aednichols:
Makes sense. To be honest, I didn't know that much about how Docker image building worked before this ticket, so I cast a wide net.

README.md
```diff
@@ -1,12 +1,11 @@
 # Job Manager
 
 [![CircleCI branch](https://img.shields.io/circleci/project/github/DataBiosphere/job-manager/master.svg?label=Tests%20on%20Circle%20CI&style=flat-square)](https://circleci.com/gh/DataBiosphere/job-manager/tree/master)
-![Github](https://img.shields.io/badge/Supported%20Backends-cromwell%20%7C%20dsub-blue.svg?&style=flat-square)
+![Github](https://img.shields.io/badge/Supported%20Backends-cromwell-blue.svg?&style=flat-square)
```
cahrens, line 4: I'm surprised that cromwell-sub is still showing.

image

cahrens (approved):

README.md
aednichols: I think that only updates after merge

aednichols: You can preview it by checking out the branch and viewing in IntelliJ markdown split view

cahrens: Got it, thanks!

PR #736, opened 2021-12-06
Upgrade js-yaml [CROM-6825]

jgainerdewar:
Also made a few changes to get the development image working for me.

### DIFF
--- a/cromwell-instance-compose.yml
+++ b/cromwell-instance-compose.yml
@@ -10,6 +10,7 @@ services:
       service: jmui
     command: ["npm", "run-script", "ng", "--", "serve", "--host", "jmui"]
     volumes:
+      - /ui/src/assets/environments/
       - ${PWD}/ui/assets/environments/environment.cromwell.instance.json:/ui/src/assets/environments/environment.json
   cromwell:
     build:
--- a/ui/Dockerfile.dev
+++ b/ui/Dockerfile.dev
@@ -1,6 +1,8 @@
 # Note: This is the dockerfile for development purposes
 
-FROM node
+FROM node:14
+
+RUN apt-get update && apt-get -y install python
 
 WORKDIR /ui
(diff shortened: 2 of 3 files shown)


cromwell-instance-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       service: jmui
     command: ["npm", "run-script", "ng", "--", "serve", "--host", "jmui"]
     volumes:
+      - /ui/src/assets/environments/
```
jgainerdewar, line 13: Create a volume for the below config file to be mounted into. I only 60% understand why this was needed - it's related to the fact that `ui/src` is already separately mounted in the compose file this one inherits from. Without this, the mounted config file was always empty.

ui/Dockerfile.dev
```diff
@@ -1,6 +1,8 @@
 # Note: This is the dockerfile for development purposes
 
-FROM node
+FROM node:14
+
+RUN apt-get update && apt-get -y install python
```
jgainerdewar, line 5: Explicitly install Python because otherwise it can't be found when needed during npm install.

ui/Dockerfile.dev
```diff
@@ -1,6 +1,8 @@
 # Note: This is the dockerfile for development purposes
 
-FROM node
+FROM node:14
```
jgainerdewar, line 3: Specify (downgrade) node version to fix C compilation errors (see [here](https://github.com/nodejs/node/issues/38367)).

ui/Dockerfile.dev
aednichols: Can you either leave a comment or specify explicitly in the code whether this is `python2` or `python3`?

It is confusing (and not your fault) that package managers/OSes use the same name even if they are installed concurrently!

aednichols (approved):

breilly2 (approved):

PR #726, opened 2021-07-27
Unbreak the build, some image updates [BW-752]

aednichols:
~Draft PR because it seems no builds get kicked off on commits~

This PR does not intend to fully address the [ticket](https://broadworkbench.atlassian.net/browse/BW-752), but I want to check this in for my own sanity.

PR #742, opened 2022-03-16
Upgrade to Ubuntu 20 for Circle [BT-576]

mcovarr:


servers/dsub/jobs/test/base_test_cases.py
```diff
@@ -25,18 +25,21 @@
 from jobs.models.aggregation_response import AggregationResponse
 from jobs.models.status_counts import StatusCounts
 
-DOCKER_IMAGE = 'ubuntu:14.04'
+DOCKER_IMAGE = 'ubuntu:20.04'
```
breilly2, line 28: Curious to know if these tests work. However, according to the README:
> dsub support is deprecated and will be removed in a future release.

.circleci/config.yml
```diff
@@ -36,10 +37,32 @@ jobs:
           key: node-modules-{{ .Branch }}--{{ checksum "ui/package.json" }}
       - run:
           name: nvm install node and UI dependencies
+          # The Circle Ubuntu machine images of yore apparently came with nvm installed, but the current images require
```
breilly2, line 40: Looks like this all worked in the CircleCI build for this PR.

breilly2 (approved):
👍 Thanks for tackling this!

jgainerdewar (approved):

servers/dsub/jobs/test/base_test_cases.py
mcovarr: it looks like the dsub tests were actually [disabled last year](https://github.com/DataBiosphere/job-manager/pull/726/files)

PR #744, opened 2022-04-06
Security updates [BT-606]

jgainerdewar:
Resolves security issues in *non-dev* dependencies. After applying these updates I brought up the app locally and clicked around through everything I could see, seems to work normally.


mspector (approved):
Makes sense to me!

breilly2 (approved):
👍  I didn't test, but looks good to me. The biggest risk is the significant major version bump to node-sass, but most of the breaking changes are just removing support for older versions of node. As long as we're on 12+, we should be fine.

PR #747, opened 2022-04-27
BT-628 Upgrade node to 14

jgainerdewar:
Narrow upgrade of node 12 -> 14 because node 12 is EOL. Note that the dev docker file already specified node 14, so it is not included in these changes.

I ran the app locally and clicked around everything I could find, and it all seemed to work normally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,11 +50,6 @@ jobs:
             export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
             [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
             
-            # The CI code that ran on the previous "latest" Circle machine image used the nvm included on that image
-            # to install what that version of nvm considered to be the LTS version of Node. However the current Node LTS
-            # is now 16 and some of JMUI's dependencies break with Node 14+, so install Node 12 explicitly.
-            #
-            # npm WARN deprecated chokidar@1.7.0: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
             nvm install 14
             
             cd ui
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -55,12 +55,12 @@ jobs:
             # is now 16 and some of JMUI's dependencies break with Node 14+, so install Node 12 explicitly.
             #
             # npm WARN deprecated chokidar@1.7.0: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
```
breilly2, line 57: Nit: Is this comment completely outdated now? I don't see any other instances of `chokidar` in the repo.

breilly2 (approved):
👍  I didn't test, but I trust that it works.

.circleci/config.yml
jgainerdewar: It's in `package-lock.json`. I went back and forth about whether to remove this... it's a legitimate problem, but only for development (chokidar is a dev-only dependency), and it's one of many scary warnings output by npm. I think it's probably better to remove it, though, and will do so.

breilly2: Oh, if it actually _is_ still used and hasn't been updated, then we should probably update it? Or change the comment to say that `chokidar` is now broken until it gets updated.

mspector:
I tested this on my M1 Mac, and ran into a bunch of errors, seemingly related to Python and libssl.

I fixed it by removing the following in `ui/Dockerfile.dev`:
```
RUN apt-get update && apt-get -y install python
```

and replacing with the following:
```
RUN apt-get update && apt-get install libssl-dev openssl && \
     wget https://www.python.org/ftp/python/3.7.13/Python-3.7.13.tgz && \
     tar xzvf Python-3.7.13.tgz && \
     cd Python-3.7.13 && \
     ./configure && \
     make && \
     make install
```

I probably copied this solution from somewhere while I was trying to get Job Manager running on my machine a couple of weeks ago.

jgainerdewar:
Huh! It looks like that installs Python 3, though, and the comment above that line says we're deliberately installing Python 2 because some of the npm packages need it. Does the switch from 12 to 14 affect what you're seeing?

.circleci/config.yml
jgainerdewar: I'll add it to `Dockerfile.dev`, since it's more relevant there.

mspector (approved):

PR #750, opened 2022-08-09
[BT-720] Update `extend`

breilly2:
Update usage of the `extend` library to at least `3.0.2`.

After initially believing that this was actually a prod dependency, further investigation makes me unsure now. The closest links I could find were through `request:2.81.0`, but that traces back to `chokidar` (a file watcher for running development mode, via `fsevents:1.1.3` -> `node-pre-gyp:0.6.39`) and `less:2.7.3` (via `@angular/cli:1.6.5`).

However, since I believe I've already done the work to update this, I'd like to push it through. I think there's still value to be gained in confirming that this will clear the SourceClear warning.

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -3825,9 +3825,9 @@
       }
     },
     "extend": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/extend/-/extend-3.0.1.tgz",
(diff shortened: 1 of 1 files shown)


kraefrei (approved):
LGTM!

mspector (approved):

PR #751, opened 2022-08-18
[BT-734] Migrate from npm to yarn

breilly2:
Migrate from npm to yarn (3.2.2) using the node-modules linker. (I know I kept saying "yarn 2", but 3.2.2 is the most recent version, and I think there were more significant changes between 1.x to 2.x than from 2.x to 3.x, at least for the purposes of this work.)

This changes the dev build (`Dockerfile.dev`) and production build (`Dockerfile`) to use yarn. In addition, the way that the yarn version and configuration is managed, this also guarantees that running `yarn` on a local workstation will use the same version as the docker image builds. This was definitely _not_ true when using npm, making it more difficult to assess local changes to dependencies; I've had to run `npm install` (and `npm-why` or `npm explain`) in a docker container just to make sure I was seeing consistent results.

I also made a couple of additions to the Terra quickstart doc.

I tried several routes to get to a `yarn.lock`. With any version of yarn, `yarn install` will resolve dependencies described in `package.json`. In addition, yarn 1.x has a `yarn import` command (missing from yarn 3.x) that also considers `package-lock.json` in an effort to create a more accurate reproduction of that. I think they removed `yarn import` because they've improved the dependency resolution logic enough to make `import` redundant. However, I didn't find that to be 100% true. The best path I found was to run `yarn import` with yarn 1.x to make use of `package-lock.json`, then `yarn install` with yarn 3.x to bring the `yarn.lock` up-to-date with the latest yarn version.

I did not attempt to adapt PnP. That can be something to consider for the future.


TERRA_QUICKSTART.md
```diff
@@ -30,6 +30,10 @@ $ ifconfig | grep inet | grep broadcast
     inet 192.XXX.Y.ZZZ netmask 0xffffff00 broadcast 192.XXX.Y.255
 $ export CROMWELL_URL="http://192.XXX.Y.ZZZ:8000/api/workflows/v1"
 ```
+or the one-liner:
+```
+$ export CROMWELL_URL="http://$(ifconfig | grep inet | grep broadcast | cut -d ' ' -f 2):8000/api/workflows/v1"
```
jgainerdewar, line 35: NICE

jgainerdewar (approved):
I checked out this branch, set my `CROMWELL_URL`, ran `docker-compose up`, and everything worked great. 👍

TERRA_QUICKSTART.md
breilly2: I've gotten this wrong more than once before by either copying the broadcast address instead of the computer's address or forgetting to include the port number, so having the one-liner helps me avoid those mistakes.

kraefrei (approved):

PR #755, opened 2022-09-12
BT-749 BT-748 Fix tests for updated dependencies

kraefrei:


ui/src/app/job-list/job-list.component.spec.ts
```diff
@@ -193,7 +193,8 @@ describe('JobListComponent', () => {
     expect(de.queryAll(By.css('.spinner-container')).length).toEqual(0);
   }));
 
-  it('paginates forward', fakeAsync(() => {
+  // Ignoring broken test, job list page not currently supported
```
kraefrei, line 196: comment was a good idea here!

kraefrei (commented):
LGTM

ui/src/app/testing/fake-config-loader.service.ts
```diff
@@ -0,0 +1,15 @@
+import { ConfigLoaderService } from "../../environments/config-loader.service";
+
+export class FakeConfigLoaderService extends ConfigLoaderService {
+    private testEnvironmentConfig: object = {
+        apiUrl : "https://go.to.here.com"
```
aednichols, line 5: I don't have that much to add in other areas of this PR so I will nitpick the one detail I do know about... it is considered suboptimal practice to use valid domains because someone could register them and collect traffic sent to them. `example.com` is [defined by RFC](https://www.iana.org/domains/reserved) to be a safe target.

```suggestion
        apiUrl : "https://go.to.here.example.com"
```

ui/src/app/job-details/tabs/tabs.component.spec.ts
```diff
@@ -190,7 +220,7 @@ describe('JobTabsComponent', () => {
     let de: DebugElement = fixture.debugElement;
 
     expect(de.queryAll(By.css('mat-expansion-panel.list-row div.task-name')).length)
-      .toEqual(tasks.length);
+      .toEqual(tasks.length + attempts.length);
```
aednichols, line 223: This seems like an unexpectedly substantive change. Was there a bug here?

jgainerdewar: Yeah, I think Old Angular reliably ran these tests in the order they appear in the file, while New Angular runs them in a random order. A test further down in this file modified a member of `tasks` by adding two attempt objects to it, which results in this DOM query returning two additional rows (iff that test runs first). I "fixed" the broken test by adding the two attempts to the task at creation time, so the test fixture appears the same to every test in the file no matter when its run.

aednichols (approved):

PR #759, opened 2022-10-03
Futurize Job Manager (to Py3)

kraefrei:


servers/cromwell/jobs/util.py
```diff
@@ -1,8 +1,11 @@
-from typing import GenericMeta
-from datetime import datetime, date
+from datetime import date, datetime
+
 from six import integer_types, iteritems
 
 
+class GenericMeta(type):
```
kraefrei, line 6: I found this as a suggested workaround for the removal of GenericMeta in python 3.8 [here:](https://github.com/anchore/anchore-engine/issues/611)

servers/cromwell/Dockerfile
```diff
@@ -7,15 +7,15 @@ ADD ./ /job-manager
 RUN /bin/bash -c scripts/rebuild_swagger.sh
 
 # We cannot yet adopt a blessed Python image because Job Manager uses python2 and blessed only has 3
-FROM gcr.io/google-appengine/python
+FROM python:3
```
aednichols, line 10: No need to implement immediately, but "blessed" suggests a selection from [this list](https://github.com/broadinstitute/dsp-appsec-blessed-images/tree/main/python) which are hosted [on GCR](https://console.cloud.google.com/gcr/images/broad-dsp-gcr-public/us/base/python?project=broad-dsp-gcr-public).

aednichols (commented):
Looking good! Left 1 suggestion but will hold off with full review until marked non-draft.

servers/cromwell/Dockerfile
aednichols: Great question. In general, the strategy is to choose the most minimal one that is still sufficient for everything to work correctly.

The preference hierarchy is distroless > alpine > debian.

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -482,7 +486,7 @@ def get_last_page(total_results, page_size):
 def page_from_offset(offset, page_size):
     if offset == 0:
         return 1
-    return 1 + (offset / page_size)
+    return math.floor(1 + (offset / page_size))
```
kraefrei, line 489: Floor is used here to revert to normal integer division, which I think was being relied on before. The tests assume an integer result.

servers/cromwell/Dockerfile
kraefrei: So I ended up using the debian image, as the other two did not work out of the box. I think this may just be an issue because we need gcc for a python package and that is the only distro with it, so its possible we could address that directly, but I decided not to investigate further.

ui/Dockerfile.dev
```diff
@@ -1,14 +1,14 @@
 # Note: This is the dockerfile for development purposes
 
 # We are aware that some dev dependencies are broken in node 14, for example chokidar
-# npm WARN deprecated chokidar@1.7.0: Chokidar 2 will break on node v14+. 
+# npm WARN deprecated chokidar@1.7.0: Chokidar 2 will break on node v14+.
```
jgainerdewar, line 4: Unrelated to this PR, but this comment can also be removed now. Chokidar was upgraded by the UI work.

jgainerdewar (approved):
Looks reasonable to me!

servers/cromwell/jobs/test/test_workflow/test_workflow.wdl
```diff
@@ -143,7 +143,7 @@ task TestAdapterSmartSeq2SingleCellTask {
     echo "Int: "${retry_seconds}
     echo "Int: "${timeout_seconds}
 
-    python <<CODE
+    python3 <<CODE
```
aednichols, line 146: Impressive comprehensiveness updating Python inside a test WDL's runtime!

FWIW this WDL is unused except as a getting-started example in `servers/cromwell/README.md`, candidate for removal if we want to do more cleanup in a future ticket.

servers/jm_utils/requirements.txt
```diff
@@ -1 +1 @@
-pytz==2017.3
+pytz==2022.4
```
aednichols, line 1: Wow!

ui/Dockerfile.dev
```diff
@@ -1,14 +1,14 @@
...
 FROM node:14
 
-# Install python 2, needed for npm packages
-RUN apt-get update && apt-get -y install python
+# Install python 3, needed for npm packages
+RUN apt-get update && apt-get -y install python3
```
aednichols, line 8: I would hope that `apt` installs Python 3 when you say `install python`, the special `python3` alias seems a few years out of date at this point. (It was created so people could keep Python 2 installed as `python`.)

aednichols (approved):

aednichols (approved):

servers/cromwell/jobs/util.py
```diff
@@ -23,11 +31,10 @@ def _deserialize(data, klass):
...
-    elif type(klass) == GenericMeta:
-        if klass.__extra__ == list:
-            return _deserialize_list(data, klass.__args__[0])
-        if klass.__extra__ == dict:
-            return _deserialize_dict(data, klass.__args__[1])
+    elif str(typing.List) in str(klass):
+        return _deserialize_list(data, klass.__args__[0])
+    elif str(typing.Dict) in str(klass):
+        return _deserialize_dict(data, klass.__args__[1])
     else:
```
jgainerdewar, line 38: This seems like a totally fine solution to me. I would guess there's a better way to do it, but <shrug>. I'm curious, though, is it the `GenericMeta` part that no longer worked? Or the `__extra__`?

jgainerdewar (approved):
🎉 🎉 🎉

servers/cromwell/jobs/util.py
kraefrei: So at some level both, the __extra__ field I think is unique to either GenericMeta, or the List and Dict classes so we can't check for __extra__ and avoid errors for other classes it seemed. The new list and dict classes do not have the common base class GenericMeta anymore, so this check always failed. I tried some other strategies to just check if the class was a List or a Dict, but the template type meant that it would not match, I'd have to check for the possible sub types, which defeats the purpose of it being generic. 

This is a good reminder though I think I can remove the GenericMeta above, since I am no longer using it.

PR #765, opened 2022-11-16
WX-814 Security Updates

jgainerdewar:


jgainerdewar:
Dealing with some Python version issues, more changes incoming...

kraefrei (approved):

servers/cromwell/jobs/controllers/jobs_controller.py
```diff
@@ -32,7 +32,12 @@
 from werkzeug.exceptions import (BadRequest, Forbidden, InternalServerError,
                                  NotFound, ServiceUnavailable, Unauthorized)
 
-collections.Callable = collections.abc.Callable
+# This is needed to support different Python versions - this
+# package structure changed in Python 3.10.
+try:
+    collections.Callable = collections.abc.Callable
+except ImportError:
+    pass
```
aednichols, line 40: Huh, so our code expects `collections.Callable` and so if only `collections.abc.Callable` is available in ≥3.10 we reassign it? And otherwise swallow the exception into a no-op.

aednichols (approved):

servers/cromwell/jobs/controllers/jobs_controller.py
jgainerdewar: Exactly - except that it's not our code, it's library code we're importing. You can see that in a previous commit I removed it because I thought it was unused. TBH I don't understand why this is needed - shouldn't the library handle its own imports? But it is.

PR #757, opened 2022-09-22
BT-758 404 for job list

jgainerdewar:
At least temporarily, disable nonfunctional job list page.

PR #756, opened 2022-09-20
BT-750 Job manager UI modernization

jgainerdewar:
This is the merge of the UI modernization project we've been working on in a feature branch. The goal is to reset ourselves to modern versions of all UI packages, in order to resolve existing security vulnerabilities and make future updates easier. We're also moving from npm to yarn.

This work is focused on enabling Job Manager in Terra. As of this PR, functionality not needed for Terra is not guaranteed to work.

PR #780, opened 2023-03-13
WX-915 Restore Job Lists Page

JVThomas:
Addresses [WX-915](https://broadworkbench.atlassian.net/browse/WX-915)

PR restores the job lists page by restoring the Job Lists Component to the UI routing as well as providing missing variable arguments for sub-components. PR also updates backend encoding/decoding logic to account for changes to how `str` and `bytes` are treated in Python 3.  

In Python 2, `str` was used for both text and `bytes`,  making `str` data ambiguous. So for Python 2, assigning either text or `bytes` to a `str` variable for a JSON response would resolve peacefully.  In Python 3 `bytes` are distinct from `str`, so passing `bytes` as part of a JSON response will now throw an error. 

In terms of updates, `byte` tokens generated for front-end pagination now need to be decoded into strings before being sent off as part of a JSON response. For pagination queries,  incoming `str` tokens need to be encoded as `bytes` before they can be processed in later steps.

To test it out, you'll should follow the steps outlined in https://github.com/DataBiosphere/job-manager/pull/779 for a mock Terra setup since there's enough dev data to test the pagination flow, but feel free to use a local Cromwell instance if you have enough data saved locally.

**NOTE**: You'll also need to run `docker compose build` due to updates to `jm_util` methods.

[WX-915]: https://broadworkbench.atlassian.net/browse/WX-915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JVThomas:
> Oh man, I'm embarrassed if this single issue was the thing preventing it from working. 😂
> 
> We commented some tests out as part of disabling this page, so we should un-comment them and ensure they pass. We'll also want to:
> 
> * Fully revert [BT-758 404 for job list #757](https://github.com/DataBiosphere/job-manager/pull/757)
> * Restore the button on the job detail page that links to the job list. It was removed [here](https://github.com/DataBiosphere/job-manager/pull/756/files#diff-3ed3bcf48dc5db8088133dcf894280c1f11c50c7ae6dd32e1b96d6c6f90f9b7c).  (See also the change in that PR to the `job-details.component` template)

Got it, but I don't think I need to do a full reversion of [BT-758](https://github.com/DataBiosphere/job-manager/pull/757). The `PageNotFound` component is nice to have for invalid, non-root paths.

jgainerdewar:
I see a few weird styling things, and a few functional things that I'm fairly certain worked before. 

Misaligned arrows:
Screen Shot 2023-03-20 at 3 19 43 PM

This badge showing the status we're filtering on has a weird doubled effect. Also, clicking on the X does not remove it (though clicking the small X on the far right does).
Screen Shot 2023-03-20 at 3 18 48 PM

This gear is supposed to display a menu that enables users to configure the columns displayed in the table. I think the contents of this popup should be a list of columns? There's a chance this is a config issue on my side.
Screen Shot 2023-03-20 at 3 23 30 PM

Still poking around but wanted to get this stuff out.

JVThomas:
> I see a few weird styling things, and a few functional things that I'm fairly certain worked before.
> 
> Misaligned arrows: Screen Shot 2023-03-20 at 3 19 43 PM
> 
> This badge showing the status we're filtering on has a weird doubled effect. Also, clicking on the X does not remove it (though clicking the small X on the far right does). Screen Shot 2023-03-20 at 3 18 48 PM
> 
> This gear is supposed to display a menu that enables users to configure the columns displayed in the table. I think the contents of this popup should be a list of columns? There's a chance this is a config issue on my side. Screen Shot 2023-03-20 at 3 23 30 PM
> 
> Still poking around but wanted to get this stuff out.

The tag not closing and the missing column list are both bugs. The `x` on the far right of the tags seems intentional. There is a coded button that, on click, clears all tags.

I've update the PR to handle the bug fixes and associated test.

JVThomas:
> Looking good! I tested the filtering and pagination and they seem to be working correctly. Few more small UI things I noticed:
> 
> * The badges for status filtering no longer look doubled but still have a gray outline on the right and left
> * The dropdown for selecting page size isn't styled as a dropdown, you have to click on the number to tell it's interactable
> 
> Couple things that I don't know are regressions but want to point out just in case:
> 
> * When I add a comment to a run, I need to refresh the page before my update is reflected in the UI.
> * The Set Metadata button appears when I check off workflows but is grayed out. This corresponds to the `bulk-edit` component. Have you found a way to test that bulk editing works?

The first two might be CSS issues or it could be a third party library. Either way I'll try to fix them with a CSS update

The comment issue seems like a reques/ response issue, looking into it as well

The metadata being greyed out is interesting because it's visible for me
Screen Shot 2023-03-21 at 12 09 51 PM

In your `capabilities_config.json` file, each of the `displayField` objects can have an optional key `bulkEditable`. If any one of them has it set to true the button should be enabled

jgainerdewar:
> In your capabilities_config.json file, each of the displayField objects can have an optional key bulkEditable. If any one of them has it set to true the button should be enabled

Nice, this worked. It does work like the individual field editing, I need to reload the page in order to see the update.

jgainerdewar (approved):
I agree that the box-shadow on the filter tags isn't a blocker. 👍 

Thank you so much for all your work on this, feels so good to get this page working again!

PR #781, opened 2023-03-16
WX-954 WX-937 Security updates

jgainerdewar:
Based on https://github.com/DataBiosphere/job-manager/pull/779 to ensure these updates work with our new auth logic.


JVThomas (approved):
Everything is running as expected locally.

PR #790, opened 2023-07-27
WX-1192 Updated semver versions for flagged dependencies

JVThomas:
Addresses [WX-1192](https://broadworkbench.atlassian.net/browse/WX-1192)

PR updates the `semver` dependency in `sass` to `7.5.2` in order to patch out a CVE present in older versions.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -63,5 +63,9 @@
     "ts-node": "^10.9.1",
     "typescript": "~4.8.2"
   },
+  "resolutions": {
+    "node-gyp/semver@^7.3.5": "7.5.2",
+    "@npmcli/fs/semver@^7.3.5": "7.5.2"
+  },
   "packageManager": "yarn@3.2.2"
 }
--- a/ui/yarn.lock
+++ b/ui/yarn.lock
@@ -9250,6 +9250,17 @@ __metadata:
   languageName: node
   linkType: hard
 
+"semver@npm:7.5.2":
+  version: 7.5.2
+  resolution: "semver@npm:7.5.2"
+  dependencies:
+    lru-cache: ^6.0.0
+  bin:
+    semver: bin/semver.js
+  checksum: 3fdf5d1e6f170fe8bcc41669e31787649af91af7f54f05c71d0865bb7aa27e8b92f68b3e6b582483e2c1c648008bc84249d2cd86301771fe5cbf7621d1fe5375
+  languageName: node
+  linkType: hard
+
 "semver@npm:^5.3.0, semver@npm:^5.6.0":
   version: 5.7.1
   resolution: "semver@npm:5.7.1"


servers/cromwell/constraints.txt
```diff
@@ -0,0 +1 @@
+cython<3.0.0
```
THWiseman, line 1: Will this constraint apply to every python package? I'm worried that it will end up causing more problems down the road as other packages begin to rely on new cython features. Is it possible to constrain just the problematic package?

servers/cromwell/tox.ini
```diff
@@ -2,6 +2,7 @@
 envlist = py310
 
 [testenv]
+setenv=PIP_CONSTRAINT={toxinidir}/constraints.txt
```
THWiseman, line 5: I believe it's good practice to include the `constraints.txt` inside a particular `requirements.txt`. I think it could make it a little easier down the road to figure out what is being constrained and why. [This article](https://luminousmen.com/post/pip-constraints-files) seems helpful. Not a huge deal if there's some reason we can't do that.

THWiseman (commented):
I think it would be good if we left some sort of comment, somewhere, explaining how we're messing with python and why (and maybe a link to that github issue). I can see something like this coming up again...

servers/cromwell/tox.ini
JVThomas: I went with the above per `tox` documentation (here)[https://tox.wiki/en/3.26.0/example/basic.html], which outlines setting constraints as seen above.

Didn't know you could do the above (most Github posts have people adding the `-c <file name>` flag to `pip install`), but seeing how it conflicts with the documented usage in `tox` I've opted for this instead.

That said I do need to update any `pip install` references to include the constraints flag. I'll add a new commit that does it.

servers/cromwell/constraints.txt
JVThomas: Yeah it will, however we pin dependencies to their exact values. If a fixed dependency updates it's `Cython` sub-dependency from `v2` to `v3` and still uses the deprecated features of `Cython v2` then it'll break as well. 

Constraining the version won't change the functionality of the library, and new features on a library would be tied to a version update which JM won't see due to pinned versions.

THWiseman: That makes sense.

THWiseman (approved):
LGTM!

servers/cromwell/tox.ini
THWiseman: I'm in favor of doing what `tox` documentation does. 

Good catch on the pip/docker stuff!

README.md
```diff
@@ -105,7 +105,7 @@ Monitors jobs launched by the [Cromwell workflow engine](https://github.com/broa
 #### Notes
 1. Websocket reload on code change does not work in docker-compose (see
 https://github.com/angular/angular-cli/issues/6349).
-2. Changes to `package.json` or `requirements.txt` or [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
+2. Changes to `package.json`, `requirements.txt` or `constraints.txt` [regenerating the API](#updating-the-api-using-swagger-codegen) require a rebuild with:
```
jgainerdewar, line 108: Happy to go with your preference here but can't miss this chance to stand up for the oxford comma.

jgainerdewar (approved):

JVThomas:
Sorry, I'm going to ask for a re-review on my PR. While `tox` was able to handle the backend installation correctly, building the Docker image locally did not. Further investigation led me to this comment:

https://github.com/yaml/pyyaml/issues/736#issuecomment-1653209769

Turns out that using the `-c` flag on pip install does not work properly if the package is installed through a wheel. In order to ensure that the constraints file is respected, the file itself must be assigned to the`PIP_CONSTRAINT` env variable alongside the individual install of PyYAML via `pip install` (PyYAML is also commented out of the `requirements.txt` file.

Once the above completes, the Dockerfile will install the rest of the packages via `requirements.txt`

THWiseman (approved):

PR #795, opened 2024-12-09
AN-301 Update certifi

jgainerdewar:


.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ run_secret_setup: &run_secret_setup
     command: |
       git clone https://github.com/awslabs/git-secrets.git
       pushd git-secrets >/dev/null
-      git checkout ad82d68ee924906a0401dfd48de5057731a9bc84
+      git checkout 5357e18bc27b42a827b6780564ea873a72ca1f01
```
LizBaldo, line 22: Could you add a comment explaining what pointing to this specific commit does?

LizBaldo (approved):
Thanks for doing this, looks like some python setup fun. I have one minor request for adding a comment but should not be a blocker for merging. :shipit:

.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     <<: *job_defaults
     machine:
       # latest pinned 20.04 version https://circleci.com/developer/machine/image/ubuntu-2004
-      image: ubuntu-2004:202201-02
+      image: ubuntu-2004:2024.11.1
```
aednichols, line 31: Could be worth updating to the [latest LTS](https://circleci.com/developer/machine/image/ubuntu-2404).

20.04 maintenance ends in April 2025.

```suggestion
      image: ubuntu-2404:2024.11.1
```

aednichols (approved):

.circleci/config.yml
jgainerdewar: Good call!

PR #796, opened 2024-12-12
AN-296 Upgrade Werkzeug to 3.0.6

jgainerdewar:
This upgrade ended up being, uh, nontrivial. 
 * The Werkzeug 2 -> 3 upgrade also resulted in Flask and Connexion needing the same major version upgrade, and Connexion's update required code changes. 
 * The Flask update from 2 -> 3 broke `flask_testing`, which seems to be abandonware, so updated tests to use plain `unittest` instead. 
 * Needed to re-order endpoints in the swagger yaml to put narrower definitions at the beginning (ex. `/jobs/operationDetails` before `/jobs/{id}`), otherwise parameter parsing failed.
 * I removed the special handling for PyYAML because I see no evidence that we still need it.


aednichols (approved):
Nice work!

High time to replace Job Manager so we don't do this again in two years.

api/jobs.yaml
```diff
@@ -33,60 +33,56 @@ paths:
       tags:
         - Capabilities
 
-  '/jobs/{id}/abort':
-    post:
-      operationId: AbortJob
-      summary: Abort a job by ID
+  '/jobs/operationDetails':
```
LizBaldo, line 36: Nit, but what do you think about adding a comment in the Readme, or at the beginning of this file to let future dev know about how the endpoints need to be ordered?

servers/cromwell/jobs/test/test_auth_utils.py
```diff
@@ -5,20 +5,24 @@
 from asyncio.log import logger
 
 import requests_mock
+import unittest
 
-from ..__main__ import loadCapabilities
-from ..models.capabilities_response import CapabilitiesResponse
-from . import BaseTestCase
+from . import create_app
```
LizBaldo, line 10: Nit, and not necessarily something to tackle in this PR as it seemed to have n=been the prior behavior, but using relative paths to import modules are not very robust, and setting up each __init__.py file so that we can import like `from test import create_app` would be a lot nicer

LizBaldo (approved):
Amazing, thanks for figuring this out 😱 I only have a couple of minor comments, lmk what you think.

api/jobs.yaml
jgainerdewar: Yes, good idea!

servers/cromwell/jobs/test/test_auth_utils.py
jgainerdewar: Agreed... I'm not going to do that in this PR, I think there's enough churn already, but good to keep in mind next time we're in here.

PR #797, opened 2025-01-07
Move abort button

lucymcnatt:
Abort and "Go back" buttons are in the same place in the job manager UI
A user can easily accidentally hit the abort button when trying to go back

(Also updated the readme because docker-compose is deprecated)

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Monitors jobs launched by the [Cromwell workflow engine](https://github.com/broa
 
 
 ### Run Locally
-- Run `docker-compose up` from the root of the repository:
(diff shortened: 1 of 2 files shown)


jgainerdewar:
I hate to destroy the purity of this beautiful cut/paste PR, but I think the styling in the new location is a little awkward:
Screenshot 2025-01-07 at 2 48 50 PM

Maybe we can left-align it in the gray box and center-align it vertically? Can also move the button between the workflow id and the gray box if that's easier.

jgainerdewar (approved):
Thanks for making those README changes!

salonishah11 (approved):

PR #798, opened 2025-03-13
[AN-429] Make wf input text wrap

lucymcnatt:
Old:
image

New:
image


lucymcnatt:
I couldn't easily figure out how to make the border for the name column to be at the same level at the value column...if we think that's important I can keep trying things

jgainerdewar (approved):

aednichols (approved):

